PR #5, opened 2021-01-23
add an example of multi-task (NOT multi-thread) implemented by s_task.

xhawk18:
Respberry Pi Pico has 264 kb SRAM, which is large enough to run a simple multi-task system,
so I make a porting of s_task for Pico in this pull request.


kilograham:
Thanks, but people can certainly use your s_task library if you like, but the pico-examples is just for examples showcasing the SDK library.

Issue #6, opened 2021-01-23
Get an unique device ID


thibaultmeyer:
Hello,

is there a way to obtain a unique identifier for each device? Something like MAC address on other device Who has a network card ?

Wren6991:
All RP2040s are identical, and there's no persistent state of any kind on-device, so there's no UID for RP2040s. Flash chips tend to have a unique identifier (usually the 4Bh command), so that would get you a per-board ID on Pico.

There's some careful footwork involved in sending serial commands to the flash chip, owing to the fact you are also executing code from it, so we can look at adding a general interface to the `hardware_flash` library for posting commands and getting a response back.

xhawk18:
unique id is the basic but necessary feature for all MCU ...

Wren6991:
> unique id is the basic but necessary feature for all MCU ...

Yes, it's also a standard feature on the flash chips we boot from, which have a 1:1 association with the MCU.

Small OTP stores take up a _lot_ of space on 40nm (it's the programming hardware, not the storage) so why make people pay for a duplicate feature on their silicon? We do need to add an example though, since as you say it needs to be something that's trivial to get at in user software. I'll get on that.

kripton:
Is the serial number the device presents to the USB host when running the bootloader also coming from the flash?
```
usb 1-4: Product: RP2 Boot
usb 1-4: Manufacturer: Raspberry Pi
usb 1-4: SerialNumber: E0C912D24340
```

XanaduNWH:
> Is the serial number the device presents to the USB host when running the bootloader also coming from the flash?
> 
> ```
> usb 1-4: Product: RP2 Boot
> usb 1-4: Manufacturer: Raspberry Pi
> usb 1-4: SerialNumber: E0C912D24340
> ```

I have a same SerialNumber in my dmesg :(

aallan:
> usb 1-4: SerialNumber: E0C912D24340

This is the firmware revision number. It will be the same on all chips.

kilograham:
merged into develop

PR #14, opened 2021-01-24
Add system/unique_board_id

Wren6991:
Example of accessing the unique 64 bit ID of external flash, which can be used as a unique board identifier


Wren6991:
See #6

system/unique_board_id/unique_board_id.c
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
```
lurch, line 2: 2021? :wink:

Wren6991: Fixed via force push, thanks

Wren6991:
Note this depends on https://github.com/raspberrypi/pico-sdk/pull/12

Wren6991:
Now updated to use the safe `pico/unique_id.h` accessor

PR #20, opened 2021-01-28
Standardise start/end comment tags (to our internal format)

lurch:


aallan:
@lurch You're intending to check the books for problems? Or have done already. Those tags are **really** fragile.

lurch:
It doesn't change the tag-names, only the tag-markup, to stop it appearing in the doxygen output (e.g. the `// tag::hello_divider[]` will disappear from https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__divider.html#divider_example )

kilograham (approved):

PR #28, opened 2021-02-02
Add Serial Output to LED Blink code

DhruvaG2000:
Generally, LED blink code also comes with serial output of LED ON and LED OFF, hence I feel this example be provided instead of bare blink.

### DIFF
--- a/blink/CMakeLists.txt
+++ b/blink/CMakeLists.txt
@@ -2,6 +2,10 @@ add_executable(blink
         blink.c
         )
 
+# enable usb output, and uart output
+pico_enable_stdio_usb(hello_usb 1)
+pico_enable_stdio_uart(hello_usb 1)
+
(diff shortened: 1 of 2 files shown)


lurch:
I _think_ the existing blink example was intended to be as-simple-as-possible, so I'm not sure if adding printf to this is desirable? If you try stepping through the code with a debugger, you'll see that printf and serial-output (especially over USB) add quite a lot of low-level complexity.

I wonder if a separate blink_with_serial_output (or serial_output_with_blink) example might be useful though? :shrug:

aallan:
@lurch Blink with serial output is the "Create your own" project in Chapter 7 the "[Getting Started](https://datasheets.raspberrypi.org/pico/getting-started-with-pico.pdf)" book. I don't think adding this to pico-examples is going to be useful.

DhruvaG2000:
okayy @aallan

PR #39, opened 2021-02-04
Fix WS2812 Timing

nickfox-taterli:
Original Timing: 1 => 840ns H + 360 ns L 0 => 240 ns H + 960 ns L (!There is a 10ns overrun of the upper limit.)
Correct Timing: 1 => 720ns H + 600 ns L 0 => 360 ns H + 720 ns L
Datasheet Timing: 1 => 350ns H + 800ns L 0 => 700ns H + 600ns L

### DIFF
--- a/pio/ws2812/ws2812.pio
+++ b/pio/ws2812/ws2812.pio
@@ -17,12 +17,13 @@
 
 .wrap_target
 bitloop:
-    out x, 1       side 0 [T3 - 1] ; Side-set still takes place when instruction stalls
-    jmp !x do_zero side 1 [T1 - 1] ; Branch on the bit we shifted out. Positive pulse
+    out x, 1       side 0 [3] ; Side-set still takes place when instruction stalls
+    jmp !x do_zero side 1 [2] ; Branch on the bit we shifted out. Positive pulse
 do_one:
-    jmp  bitloop   side 1 [T2 - 1] ; Continue driving high, for a long pulse
+    nop            side 1 [2] ; Continue driving high, for a long pulse
+    jmp  bitloop   side 0 ; Padding zero
 do_zero:
-    nop            side 0 [T2 - 1] ; Or drive low, for a short pulse
+    nop            side 0 [1] ; Or drive low, for a short pulse
 .wrap
 
 % c-sdk {


henrygab (commented):
First of three questions / comments...

<details><summary>T1, T2, T3 match FastLED WS2812 timings?</summary><P/>


[FastLED](https://github.com/FastLED/FastLED) is a well-respected library, and gives a fairly good summary of the [definition of T1, T2, T3](https://github.com/FastLED/FastLED/blob/1fb0d7b2afd87695eb5f3c4af27c286b5e9f5fe0/src/chipsets.h#L432-L436):
```C++
// Base template for clockless controllers.  These controllers have 3 control points in their cycle for each bit.
// At T=0        : the line is raised hi to start a bit
// At T=T1       : the line is dropped low to transmit a zero bit
// At T=T1+T2    : the line is dropped low to transmit a one bit
// At T=T1+T2+T3 : the cycle is concluded (next bit can be sent)
```

the PIO file lists T1/T2/T3 as 2/5/3:

https://github.com/raspberrypi/pico-examples/blob/33854562cd08398c4b48bb1ed7fa022c2177076d/pio/ws2812/ws2812.pio#L10-L12

Presuming 125ns clock configured, this corresponds to 250ns/675ns/375ns, which exactly matches the [FastLED WS2812 timings](https://github.com/FastLED/FastLED/blob/1fb0d7b2afd87695eb5f3c4af27c286b5e9f5fe0/src/chipsets.h#L576-L578).  

Expanding this out, using 1 bit being defined as `T1+T2 H, T3 L`, and 0 bit defined as `T1 H, T2+T3 L`, gives the following timings:

* 1 bit : 875ns H + 375ns L
* 0 bit : 250ns H + 1000ns L

The values you recorded seem very, very close...

</details><hr/>

Can you help me understand where you obtained your "original timing" values?  Were you using an oscilliscope?

henrygab (commented):
If you accept that the 2/5/3 are acceptable values for T1/T2/T3, then it seems the program should work as written.  One gotcha, if reviewing timing manually, is the timing of when the delay occurs.   This could be more clear in the SDK, but it does state:

> Each PIO instruction is 16 bits in size. Generally, 5 of those bits in each instruction are used for the “delay” which is usually 0 to 31 cycles (after the instruction completes and before moving to the next instruction). 

Importantly, the delay occurs **_AFTER_** the instruction completes.

The program is otherwise well documented in the [PICO SDK](https://datasheets.raspberrypi.org/pico/raspberry-pi-pico-c-sdk.pdf) ... see section 3.2.2.1.

henrygab (changes requested):
Part 3:

Your recommended changes move away from the T1/T2/T3 concept, because it uses different values for T3, depending on whether the bit is a zero or a one.  This is risky at best ... recommended to stick within the T1/T2/T3 paradigm, as it's much more well known for this protocol.

At the same time, perhaps you can provide the link to the specific datasheet that you want to match?  If so, I might be able to help you find reasonable values for T1/T2/T3.  Just note that the maximum value for any of T1/T2/T3 will be 16 (such that subtracting one will still fit in 4 bits remaining in the opcode).

henrygab:
@nickfox-taterli -- For your own use, I have some alternate values you could try for T1/T2/T3.

<details><summary>WS2812 Example Datasheet Timing</summary></P>

Here's an example datasheet table, that is fairly typical
for WS2812 LEDs.  Note that there's a range of legal times,
defined in the datasheet:

| T          | Symbol | Base | Min |  Max |
|------------|--------|------|-----|------|
| `T1      ` | T0H    | 350  | 200 |  500 |
| `T1+T2   ` | T1H    | 700  | 550 |  850 |
| `   T2+T3` | T0L    | 800  | 650 |  950 |
| `      T3` | T1L    | 600  | 450 |  750 |
| `T1+T2+T3` | TH+TL  | 1250 | 625 | 1875 |

</details><hr/>

<details><summary>Current values vs. datasheet</summary></P>

The existing T1/T2/T3 values gave the following timings:
```
Bit 0 => 250ns H + 1000ns L (1250ns total)
Bit 1 => 875ns H +  375ns L (1250ns total)
```

* Total time per bit is **_spot-on_**
* T0H is **valid**, but on the low end
* T1H is **valid**, somewhat towards low end
* T0L exceeds specified maximum length
* T1L is below specified minimum length

NOTE: These work well in practice, across broad range of WS2812, across years of manufacturing... this is just to help you find T values that work for your particular chips.  See FastLED for a larger set of examples.

</details><hr/>
<details><summary>Where these values came from....</summary></P>

I used Excel's solver add-in, with the following constraints:
0. Total time per bit had to exactly match specification "target"
1. Each of T1, T2, T3 had to be an integer
2. Each of T1, T2, T3 had range of [1..16] (so delay would fit into opcodes)
3. Each of T0H, T1H, T0L, and T1L had to be valid (within spec)
4. Target goal was to minimize the [[ maximum absolute distance from "target" time ]]
</details><hr/>

You could try the values (8/5/12):
```
Bit 0 => 400ns H + 850ns L (1250ns total)
Bit 1 => 650ns H + 600ns L (1250ns total)
```

Again, if you have a different datasheet in mind, send a link....

lurch:
This is also analysed more in-depth at https://github.com/raspberrypi/pico-feedback/issues/121

PR #43, opened 2021-02-05
Add pio_clocked_input example

Wren6991:
See https://github.com/raspberrypi/pico-feedback/issues/38

This adds fairly minimal example of using a PIO state machine to sample data on an external clock. It uses one of the hard SPIs to drive data into the SM with a couple of jumper wires. Once this is in, we can document it in the SDK book, as there's currently a big leap between the WS2812 example and the logic analyser example.

### DIFF
--- a/pio/clocked_input/clocked_input.c
+++ b/pio/clocked_input/clocked_input.c
@@ -1,0 +1,71 @@
+/**
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "pico/stdlib.h"
+#include "hardware/clocks.h"
+#include "hardware/pio.h"
+#include "hardware/spi.h"
+#include "clocked_input.pio.h"
+
+// Set up a PIO state machine to shift in serial data, sampling with an
+// external clock, and push the data to the RX FIFO, 8 bits at a time.
+//
+// Use one of the hard SPI peripherals to drive data into the SM through a
+// pair of external jumper wires, then read back and print out the data from
+// the SM to confirm everything worked ok.
+//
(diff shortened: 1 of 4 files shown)


pio/clocked_input/clocked_input.c
```diff
@@ -0,0 +1,71 @@
+/**
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
```
lurch, line 2: 2021 :wink:

pio/clocked_input/clocked_input.c
```diff
@@ -0,0 +1,71 @@
...
+    for (int i = 0; i < BUF_SIZE; ++i) {
+        txbuf[i] = rand() >> 16;
+        printf("%02x\n", txbuf[i]);
+    }
+
+    // The "blocking" write function will send all the data in one go, and not
+    // return until the full transmission is finished.
+    spi_write_blocking(spi0, (const uint8_t*)txbuf, BUF_SIZE);
+
+    // The data we just sent should now be present in the state machine's FIFO.
```
lurch, line 64: I guess this works because there's enough room in the FIFO for all 8 (`BUF_SIZE`) bytes? Would it make this example much more complicated if it used an IRQ handler instead so that it was capable of dealing with lots more data? :shrug: 
(There's obviously trade-offs to be made between making a simple example that illustrates basic concepts, and making a robust example that doesn't explode when a user tries to extend it. Entirely up to you where that line gets drawn!)

Wren6991: > Would it make this example much more complicated if it used an IRQ handler instead so that it was capable of dealing with lots more data?

Yes I want to avoid that, this is for a section in between WS2812 and logic analyser in the SDK book.

lurch: Thanks for the clarification comment :slightly_smiling_face:

lurch (approved):

kilograham (approved):

PR #44, opened 2021-02-08
Fix logic analyser for non-power-of-2 sample widths, and clean and comment (closes #29)

### LINKED ISSUE #29: pico-examples/pio/logic_analyser not working for some values of CAPTURE_PIN_COUNT
The logic analyser example does not work when CAPTURE_PIN_COUNT does not divide exactly into 32. This is at odds with the comment in the file logic_analyser.c: "`// 1 to 32 pins can be captured.`"

For example: CAPTURE_PIN_COUNT=4 produces the following *correct* output:

`16: ----____________----____________----____________----____________----____________----____________`
`17: ------------____------------____------------____------------____------------____------------____`
`18: ________________________________________________________________________________________________`
`19: ________________________________________________________________________________________________`

However, with CAPTURE_PIN_COUNT=3 the output is:

`16: -----------____________________------------___________________------------___________________---`
`17: _______________----__------____-______________----___-----____--_____________----____----____---`
`18: ---_______-____------___________--____________-------___________-____________--------___________`

This is incorrect - the trace for pins 16,17 and 18 should be identical to the previous example.

The comment could be updated to state that "`// 1,2,4,8,16 or 32 pins can be captured.`" Alternatively the attached file contains a fix that works for all values from 1 to 32. 

Change summary: 
- The underlying issue is that the PIO discards bits when a sample overruns the available space in the ISR. To prevent this, `push_threshold` is set to a multiple of `pin_count`.
- The `capture_buf` size calculation has been updated.
- Calculation of the word and bit offsets in `print_capture_buf()` updated.

[logic_analyser.c.txt](https://github.com/raspberrypi/pico-examples/files/5910657/logic_analyser.c.txt)

Wren6991:
This fixes the PIO + DMA logic analyser example to handle non-power-of-2 sample widths correctly, and also generally cleans it up, adds more commentary, and sets the DMA to high bus priority so you should be able to crank the width all the way up to 32 bits. See #29

### DIFF
--- a/pio/logic_analyser/logic_analyser.c
+++ b/pio/logic_analyser/logic_analyser.c
@@ -14,10 +14,12 @@
 // frequency.
 
 #include <stdio.h>
+#include <stdlib.h>
 
 #include "pico/stdlib.h"
 #include "hardware/pio.h"
 #include "hardware/dma.h"
+#include "hardware/structs/bus_ctrl.h"
 
 // Some logic to analyse:
 #include "hardware/structs/pwm.h"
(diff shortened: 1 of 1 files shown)


pio/logic_analyser/logic_analyser.c
```diff
@@ -65,15 +65,18 @@ void logic_analyser_init(PIO pio, uint sm, uint pin_base, uint pin_count, float
 void logic_analyser_arm(PIO pio, uint sm, uint dma_chan, uint32_t *capture_buf, size_t capture_size_words,
                         uint trigger_pin, bool trigger_level) {
     pio_sm_set_enabled(pio, sm, false);
+    // Need to clear _input shift counter_, as well as FIFO, because there may be
+    // partial ISR contents left over from a prevoius run. sm_restart does this.
```
lurch: previous

kilograham (approved):

PR #45, opened 2021-02-08
Added button debounce example in gpio_irqs

giripriyadarshan:
Yes, debounce is very basic, but that's the purpose of examples anyways.
The onboard LED and usb/uart is used to see the debounce working

### DIFF
--- a/gpio/debounce_gpio_irq/CMakeLists.txt
+++ b/gpio/debounce_gpio_irq/CMakeLists.txt
@@ -1,0 +1,12 @@
+add_executable(debounce_gpio_irq
+        debounce_gpio_irq.c
+        )
+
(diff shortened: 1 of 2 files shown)


giripriyadarshan:
The second commit is for the unique_board_id example.
Just like hello_world examples, maybe even this needs a loop to print the board id via USB/UART.
Without the loop, the id will be printed too fast for us to open minicom (when using USB to see the id)

lurch:
Your two commits are entirely unrelated to each other, so could you split them out into separate PRs please?
(also, you'll want to target your PRs against the `develop` branch rather than the `master` branch)

giripriyadarshan:
Yep ..... I'll do it

PR #46, opened 2021-02-09
Remove debug pin usage from ws2812 parallel output sample

### LINKED ISSUE #41: PIO / WS2812 ... Bug on preprocessor debug disable?
https://github.com/raspberrypi/pico-examples/blob/33854562cd08398c4b48bb1ed7fa022c2177076d/pio/ws2812/ws2812_parallel.c#L285-L291

First, I'm at a loss trying to understand why this is ever disabled, given that the PICO has dedicated SWD pins.  This also isn't checking against `PICO_DEFAULT_UART_TX_PIN`, `PICO_DEFAULT_UART_RX_PIN`, or `PICO_DEFAULT_LED_PIN` ... was it leftover from the FPGA days?

Even if it's OK for the non-parallel, however, it's not correct for this file, since it outputs to 2 consecutive GPIO.  `PIN_TX` is only the *first* GPIO that the state machines will push output through.  Therefore, if `PIN_TX` is defined as `2`, then it would still call `gpio_debug_pins_init()`, even though GPIO2 and GPIO3 would be getting DMA...

Thoughts appreciated?

henrygab:
Fix #41

This branch was based on 0d3b9d7 ... and I only updated a single file.  Not sure why this PR is showing multiple commits?


lurch:
You PRed against the wrong branch :slightly_smiling_face:

PR #53, opened 2021-02-16
hello_double_tap: use pico_bootsel_via_double_reset from SDK

Wren6991:
Use the SDK library added by https://github.com/raspberrypi/pico-sdk/pull/137, and delete the copy in the examples tree. Also use LED 25 as the activity LED as some people do have reset buttons hacked onto Picos.


Wren6991:
SDK code that this relies on is now merged into `develop`

system/hello_double_tap/CMakeLists.txt
```diff
@@ -2,12 +2,19 @@ add_executable(hello_double_tap
...
 target_link_libraries(hello_double_tap
         pico_stdlib
-        double_tap_usb_boot
+        pico_bootsel_via_double_reset
         )
 
+# Entering the bootloader in this way also lets us specify a GPIO to be used
+# as a bootloader activity LED:
+target_compile_definitions(hello_double_tap PRIVATE
+	PICO_BOOTSEL_VIA_DOUBLE_RESET_ACTIVITY_LED=25
```
lurch, line 15: Would there be any way to make this also use `PICO_DEFAULT_LED_PIN` rather than hard-coding `25`? Or is that a "setting" that's only available in C-land and not in cmake-land?

Wren6991: It's set in the board header, so only available in C.

kilograham (approved):

PR #65, opened 2021-02-27
Onboard temperature

fkuhne:
After reading the docs and taking a look at other examples and repo coding style, I just put to work a simple application that prints out the onboard temperature sensor value.

Hope it to be useful. ;)


adc/onboard_temperature/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+add_executable(onboard_temperature onboard_temperature.c)
+
+target_link_libraries(onboard_temperature pico_stdlib hardware_adc)
+
+# enable usb output, disable uart output
```
lurch, line 5: I think most of the pico-examples have uart-output-enabled, usb-output-disabled? (because it's easier to debug with uart-output). Might be best to stick to that "convention"? It's easy enough to change back to usb-output in your local copy :wink:

adc/onboard_temperature/onboard_temperature.c
```diff
@@ -0,0 +1,48 @@
+#include <stdio.h>
```
lurch, line 1: Could you add a copyright header please?

adc/onboard_temperature/onboard_temperature.c
```diff
@@ -0,0 +1,48 @@
...
+#include "pico/stdlib.h"
+#include "hardware/adc.h"
+
+/* References for this implementation:
+ * raspberry-pi-pico-c-sdk.pdf, page 68.
```
lurch, line 6: That PDF is regularly updated, which means the page-numbers shift around. Much better to reference a section-number instead :slightly_smiling_face:

adc/onboard_temperature/onboard_temperature.c
```diff
@@ -0,0 +1,48 @@
...
+
+/* References for this implementation:
+ * raspberry-pi-pico-c-sdk.pdf, page 68. 
+ * pico-examples/adc/adc_console/adc_console.c */
+const float read_onboard_temperature(const char unit) {
+    
+    /* 12-bit conversion, assume max value == ADC_VREF == 3.3 V */
+    const float conversionFactor = 3.3f / (1 << 12);
+
+    adc_select_input(4);
```
lurch, line 13: This only needs to be done once (you're not changing ADC inputs) so could be moved to `main`?

adc/onboard_temperature/onboard_temperature.c
```diff
@@ -0,0 +1,48 @@
...
+        return tempC;
+    } else if (unit == 'F') {
+        return tempC * 9 / 5 + 32;
+    }
+
+    return -1.0;
+}
+
+int main() {
+    const uint LED_PIN = 25;
```
lurch, line 28: This would be better as `const uint LED_PIN = PICO_DEFAULT_LED_PIN;`

adc/onboard_temperature/onboard_temperature.c
```diff
@@ -0,0 +1,48 @@
...
+
+    stdio_init_all();
+    gpio_init(LED_PIN);
+    gpio_set_dir(LED_PIN, GPIO_OUT);
+
+    adc_init();
+    adc_set_temp_sensor_enabled(true);
+
+    while (true) {
+        printf("Onboard temperature = %.02f C\n", read_onboard_temperature('C'));
```
lurch, line 38: Perhaps it makes sense to add a `#define TEMPERATURE_UNITS 'C'` near the top of this file, and then this line could become:
```
printf("Onboard temperature = %.02f " TEMPERATURE_UNITS "\n", read_onboard_temperature(TEMPERATURE_UNITS));
```

Although code like this is sometimes easier to debug if split over separate lines like:
```
float temperature = read_onboard_temperature(TEMPERATURE_UNITS);
printf("Onboard temperature = %.02f " TEMPERATURE_UNITS "\n", temperature);
```

README.md
```diff
@@ -20,6 +20,7 @@ App|Description
 [hello_adc](adc/hello_adc)|Display the voltage from an ADC input.
 [joystick_display](adc/joystick_display)|Display a Joystick X/Y input based on two ADC inputs.
 [adc_console](adc/adc_console)|An interactive shell for playing with the ADC. Includes example of free-running capture mode.
+[onboard_temperature](adc/onboard_temperature)|Print out to the USB serial port the value of the onboard temperature sensor.
```
lurch, line 23: Would be more consistent with the others if this was just "Display the value of the onboard temperature sensor."

lurch (commented):
A few small comments, but looking pretty good. Could you also change it to PR against the `develop` branch please?

adc/onboard_temperature/onboard_temperature.c
fkuhne: Like the other examples? Or should I put my name, email... ?

adc/onboard_temperature/onboard_temperature.c
fkuhne: I thought a little bit about this. Efficiency-wise, yes, it would be better move it to `main`. But I have at least two points in favor of letting it inside `read_onboard_temperature`:
* It looks more concise in terms of scope to put it in the function;
* `adc_select_input()` is a global operation, so something else along a code (not this one) may be changing it.

But that's OK I can follow yout suggestion. For the sake of this example, is OK. ;)

fkuhne:
Implemented all suggestions and will make another PR to develop branch. Thanks @lurch!

adc/onboard_temperature/onboard_temperature.c
lurch: I think as long as you license it under the same 3-Clause-BSD it's up to you whether you assign copyright over to Raspberry Pi or retain copyright yourself? But I'm not a legal expert :wink:

PR #66, opened 2021-02-27
Fix PIN number of onboard LED.

sbehnke:
I found the example for hello_double_tap.c had the incorrect GPIO pin for the onboard LED.


lurch:
Already fixed by #53

PR #71, opened 2021-02-28
Update blink.c

ffdaggae:


kilograham:
i assume this was a mistake

PR #81, opened 2021-03-04
Fix examples to build cleanly if PICO_DEFAULT_LED_PIN isn't defined

lurch:


kilograham:
hmm i could swear i have a PR for this too!

kilograham:
note i will combine this with my changes - the thing I did differently for sure was to make the double_tap work without a LED if there isn't one, as the LED is not very important to the functionality of the example

lurch:
Feel free to chop'n'change or pick'n'mix my PRs however you see best :+1: 

> make the double_tap work without a LED if there isn't one, as the LED is not very important to the functionality of the example

That's a good point, in my naive copy'n'pasting I hadn't considered that.

system/hello_double_tap/hello_double_tap.c
```diff
@@ -13,6 +13,9 @@
 // `pico_bootsel_via_double_reset` library!
 
 int main() {
+#ifndef PICO_DEFAULT_LED_PIN
```
kilograham, line 16: yup, i seem to have lost my changes for this... can you fix this one up to just not use the LED instead

kilograham (changes requested):

system/hello_double_tap/hello_double_tap.c
lurch: Done.

kilograham (approved):

PR #84, opened 2021-03-07
This should make using the WS2812 examples a bit easier.

josefwegner:
Moved parameters to the top as defines:
- Number of LEDs in your NeoPixel/WS2812 strip/ring
- Flag if the LEDs are WRGB or not
Changed PIN to 2 as pin 0 is used for serial UART.
Added a sleep of 10ms to the ws2812 parallel example to match the other ws2812 example.


pio/ws2812/ws2812.c
```diff
@@ -12,6 +12,10 @@
 #include "hardware/clocks.h"
 #include "ws2812.pio.h"
 
+#define PIN_TX 2
```
lurch, line 15: @kilograham Would it be worth using the new `PICO_DEFAULT_WS2812_PIN` here? Or is that specifically only for a single WS2812 rather than a string of WS2812s?

pio/ws2812/ws2812_parallel.c
```diff
@@ -16,7 +16,9 @@
 #include "ws2812.pio.h"
 
 #define FRAC_BITS 4
-#define PIN_TX 0
+#define PIN_TX 2
+// make sure that this number is less or equal to the MAX_LENGTH defined below
+#define NUM_PIXELS 64
```
lurch, line 21: Perhaps it's worth moving the MAX_LENGTH define up to here, and then you could add a compile-time `#error` if `NUM_PIXELS > MAX_LENGTH` ?

pio/ws2812/ws2812_parallel.c
```diff
@@ -300,7 +302,7 @@ int main() {
         int brightness = 0;
         uint current = 0;
         for (int i = 0; i < 1000; ++i) {
-            int n = 64;
+            int n = NUM_PIXELS;
 
             current_string_out = string0.data;
             current_string_4color = false;
```
lurch, line 308: Might be worth making this `current_string_4color = IS_WRGB;` (and adding the define) ? :shrug: 

And I'm sure no-one would mind if you fixed the spelling to `current_string_4colour` :laughing:

pio/ws2812/ws2812.c
```diff
@@ -12,6 +12,10 @@
...
+#define NUM_PIXELS 150
+#define IS_WRGB true
```
lurch, line 17: Hmmm, a quick search suggests that `RGBW` is more common nomenclature than `WRGB`?

pio/ws2812/ws2812_parallel.c
josefwegner: We could also get rid of MAX_LENGTH altogether and just use NUM_PIXELS. But I do not understand the code good enough if there are some deeper reasons for defining a MAX_LENGTH and use a different number of LEDs.

pio/ws2812/ws2812_parallel.c
josefwegner: It seems it is possible to control a WRGB and RGB strip at the same time, so a single define will not do it. Regarding color vs colour: I can do this.

pio/ws2812/ws2812.c
josefwegner: Yes, you are right. Will change to RGBW.

pio/ws2812/ws2812_parallel.c
lurch: You're right, I hadn't looked at the code properly :wink:

pio/ws2812/ws2812_parallel.c
lurch: _Maybe_ the intention was that `n` could be different for different strings (just like `current_string_4colour` is different for the different strings), with `MAX_LENGTH` being the longest string allowed? :shrug:  There's so much going on in this code that I also can't quite tell what it is doing (without spending more time looking at it than I have free right now).

josefwegner: Even if it was the intention, it does not work that way. I removed n and MAX_LENGTH and use NUM_PIXELS instead. You can still use two strips with different length, just define the number of the longer strip.

pio/ws2812/ws2812.c
josefwegner: I updated the code to use PICO_DEFAULT_WS2812_PIN. If it is not defined for board (e.g. the Pi Pico), a warning is generated and the pin 2 is used.

pio/ws2812/ws2812_parallel.c
```diff
@@ -16,9 +16,11 @@
 #include "ws2812.pio.h"
 
 #define FRAC_BITS 4
-#define PIN_TX 2
-// make sure that this number is less or equal to the MAX_LENGTH defined below
 #define NUM_PIXELS 64
+#ifndef PICO_DEFAULT_WS2812_PIN
+#warning "no WS2812 default PIN defined for board, please check if pin 2 is okay"
+#define PICO_DEFAULT_WS2812_PIN 2
```
lurch, line 22: While the change to `ws2812.c` seems okay, I'm not sure if the change here is appropriate?
`ws2812_parallel_program_init` in `ws2812.pio` says:
```
static inline void ws2812_parallel_program_init(PIO pio, uint sm, uint offset, uint pin_base, uint pin_count, float freq) {
    for(uint i=pin_base; i<pin_base+pin_count; i++) {
        pio_gpio_init(pio, i);
    }
```
i.e. this `ws2812_parallel.c` example actually uses both pins 2 and 3. Probably makes sense to put it back to using `PIN_TX`, but also rename it to `PIN_TX_BASE` to make things more obvious?

josefwegner: That makes sense. I renamed it to `WS2812_PIN_BASE`.

pio/ws2812/ws2812_parallel.c
```diff
@@ -300,19 +299,18 @@ int main() {
...
+            pattern_table[pat].pat(NUM_PIXELS, t);
 
-            transform_strings(strings, count_of(strings), colors, n * 4, brightness);
-            dither_values(colors, states[current], states[current ^ 1], n * 4);
+            transform_strings(strings, count_of(strings), colours, NUM_PIXELS * 4, brightness);
+            dither_values(colours, states[current], states[current ^ 1], NUM_PIXELS * 4);
             sem_acquire_blocking(&reset_delay_complete_sem);
-            output_strings_dma(states[current], n * 4);
+            output_strings_dma(states[current], NUM_PIXELS * 4);
+            sleep_ms(10);
```
kilograham, line 313: why have you added a sleep here?

kilograham (commented):
note, also we do actually use color (i.e. US spelling) in all code

lurch:
> note, also we do actually use color (i.e. US spelling) in all code

Oh whoops. Sorry!
I think we consistently use UK spelling in the datasheets, right @aallan ?

pio/ws2812/ws2812_parallel.c
josefwegner: Because ws2812.c also had a sleep_ms(10) in the loop. I think the animations are too fast without it.

josefwegner:
> note, also we do actually use color (i.e. US spelling) in all code

Sorry, reverted all instances back to color.

aallan:
> > note, also we do actually use color (i.e. US spelling) in all code
> 
> Oh whoops. Sorry!
> 
> I think we consistently use UK spelling in the databooks, right @aallan ?

Yes. The documentation uses UK spelling in accordance with our general company style guide.

pio/ws2812/ws2812_parallel.c
kilograham: it destroys the point of dithering, tho; you should slow down the animation itself rather than the rendering loop

josefwegner: I do not see how I could slow down the animation. The pattern_* animation methods just fill the string, and then the string is rendered. And the animation method is called inside the rendering loop, so slowing down filling of the string has the same effect as adding a delay after the strings are rendered at the end of the loop.

kilograham: ok, so only update the string pattern 1 in N times round the loop

josefwegner: Removing the sleep_ms(). I did not find a way to update the string pattern only 1 in N times without messing up the animation or ratio between changes and dithering steps.

PR #85, opened 2021-03-10
added comments for clarification

SEEK-THE-VOID-LIGHT:
I added comments to the blink.c script for beginners to get a better overview of the example script.
When I began with embedded system I would have liked that in my showcase examples, so I wanted to add that to the Raspberry Pi Pico example script.
The code without comments can be a bit intimidating for new users.

### DIFF
--- a/blink/README.md
+++ b/blink/README.md
@@ -1,0 +1,19 @@
+### Code with comments
+
+#include "pico/stdlib.h"                                //include header files needed for the code
+
+int main() {                                            //begin main function - here will be the code that will be run on startup
+#ifndef PICO_DEFAULT_LED_PIN                            //check if PICO_DEFAULT_LED_PIN is defined (onboard led)
+#warning blink example requires a board with a regular LED
+#else                                                   //if there is an onboard led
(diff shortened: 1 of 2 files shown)


aallan:
The example code gets pulled in automagically into the PDF documentation. We can't really have heavy inline commenting for that reason.

SEEK-THE-VOID-LIGHT:
Ok, should I shorten the comments?

aallan:
I'd would tend to agree, I think the code looks awfully cluttered with that amount of additional commentary. A README.md file might be a more appropriate place.

SEEK-THE-VOID-LIGHT:
Should I wait for further rsponse, make a readme, or close the request?

lurch:
If I was you I'd revert the changes you've made to `blink.c`, create a `README.md` instead, and then wait to see what @kilograham says...

SEEK-THE-VOID-LIGHT:
👍

SEEK-THE-VOID-LIGHT:
The README now holds the commented version of the code, I can change the content if you want.

kilograham (approved):

PR #95, opened 2021-03-22
usb/device/dev_low_level: respond to other OUT control requests with a ZLP

JulienBernard3383279:
This makes it so a ZLP is sent as a response to OUT control requests other than SET_ADDRESS and SET_CONFIGURATION. Without it, other OUT control requests may crash communications.
I've experienced it on Windows 10 using the WinUSB driver / libusb library - after a custom control request, all transfer attempts would return LIBUSB_ERROR_ACCESS. This fixed it.


lurch:
ping @liamfraser

kilograham (approved):

PR #97, opened 2021-04-03
Added .gitignore for build/

michaelstoops:
Ignoring `build/`.

I realize that I and other seasoned devs will know to put this in their global `.gitignore`, but in the spirit of lowering the barriers of access and it-works-out-of-the-box, I think we should actually set it somewhere, even if it isn't here. If you disagree with putting this here, please share the repo for the default home dir skeleton or another appropriate location.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,1 @@
-.idea
-.vscode
-_deps
-cmake-*
-build
-.DS_Store
-*.pdf
+build/


lurch:
According to the "Files changed" tab, this is _changing_ `.gitignore` rather than _adding_ `.gitignore`? :confused:

michaelstoops:
I don't know where my process went wrong, but I totally see the `.gitignore`, containing `build`. 🤷 sorry.

PR #99, opened 2021-04-08
move to tinyusb 0.9.0 - switch to using their examples directly rather than copied versions

kilograham:
replace copied tinyusb examples with use of examples directly from tinyusb repo.

this avoids us having to fix up examples ourselves if the upstream API changes, though for now we still have a hard-coded list of examples, and we are using a GLOB (within an SDK provide add_tinyusb_device/host_example method) to generate build targets for them. 

this works currently, though is not ideal. to fix it we need the cmake structure within TinyUSB example to be more friendly towards the use case of including their exmaples' `CMakeLists.txt`s in this way


usb/device/CMakeLists.txt
```diff
@@ -1,4 +1,23 @@
...
-add_subdirectory(dev_hid_composite)
-add_subdirectory(dev_hid_generic_inout)
+# not 100% sure what this does
+add_tinyusb_device_example(audio_test)
+add_tinyusb_device_example(board_test)
+add_tinyusb_device_example(cdc_dual_ports)
+add_tinyusb_device_example(cdc_msc)
+# todo gate with freertos lib
+#add_tinyusb_device_example(cdc_msc_freertos)
+add_tinyusb_device_example(dfu_runtime)
```
lurch, line 8: Is this relevant, given that we recommend flashing via UF2? :shrug:

lurch:
I guess the CI tests here won't pass until after https://github.com/raspberrypi/pico-sdk/pull/321 has been merged.

usb/device/CMakeLists.txt
kilograham: what do you mean?

kilograham:
> I guess the CI tests here won't pass until after [raspberrypi/pico-sdk#321](https://github.com/raspberrypi/pico-sdk/pull/321) has been merged.

correct; would be nice if there is a way to handle this sort of dependency case without too much effort

lurch:
> would be nice if there is a way to handle this sort of dependency case without too much effort

I _think_ you might be able to fix that by editing the `pico-sdk` branch that https://github.com/raspberrypi/pico-examples/blob/master/.github/workflows/cmake.yml points at in this PR, but you'd need to remember to revert that change again before merging this PR. Maybe? :shrug:

usb/device/CMakeLists.txt
lurch: DFU = Device Firmware Upgrade, typically used to write the flash on a microcontroller (which is what we use UF2 for). But I've just checked https://github.com/hathach/tinyusb/blob/master/examples/device/dfu_runtime/src/main.c and see that it says:
> Since this example is minimal, it doesn't actually go into DFU mode but rather change the LED blinking pattern to fast rate as indicator.

So apologies for the irrelevant comment! :laughing:

lurch:
P.S. Should this PR be targeting the `develop` branch?

PR #103, opened 2021-04-21
Prevent extra bit transition at end of data

tommessick:
The initial bit transition would be set even if the queue stalled.
The state machine still takes 16 cycles per bit
The next bit starts on state 1 and 9 instead of 0 and 8 so the timing is the same as before


Wren6991 (approved):
Good fix, thank you. I rebased against `develop`

PR #109, opened 2021-05-01
Update hello_rtc.c

Reneg973:
when RTC gets new datetime values, it needs a few clocks to apply them. If get_datetime() is called before this is done, the first output on serial is wrong

### DIFF
--- a/rtc/hello_rtc/hello_rtc.c
+++ b/rtc/hello_rtc/hello_rtc.c
@@ -32,8 +32,9 @@ int main() {
     rtc_init();
     rtc_set_datetime(&t);
 
-    // clk_sys is >2000x faster than clk_rtc, so datetime is not transferred completely when rtc_get_datetime() is called
-    sleep_us(20);
+    // clk_sys is >2000x faster than clk_rtc, so datetime is not updated immediately when rtc_get_datetime() is called.
+    // tbe dealys is up to 3 RTC cycles (which is 64us with the default clock settings)
+    sleep_us(64);
 
     // Print the time
     while (true) {
--- a/rtc/hello_rtc/hello_rtc.c
+++ b/rtc/hello_rtc/hello_rtc.c
@@ -33,7 +33,7 @@ int main() {
     rtc_set_datetime(&t);
 
     // clk_sys is >2000x faster than clk_rtc, so datetime is not updated immediately when rtc_get_datetime() is called.
-    // tbe dealys is up to 3 RTC cycles (which is 64us with the default clock settings)
+    // tbe delay is up to 3 RTC clock cycles (which is 64us with the default clock settings)
     sleep_us(64);
 
     // Print the time
--- a/rtc/hello_rtc/hello_rtc.c
+++ b/rtc/hello_rtc/hello_rtc.c
@@ -32,6 +32,9 @@ int main() {
     rtc_init();
     rtc_set_datetime(&t);
 
+    // clk_sys is >2000x faster than clk_rtc, so datetime is not transferred completely when rtc_get_datetime() is called
+    sleep_us(20);
+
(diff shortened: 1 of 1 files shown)


lurch:
( This links to https://github.com/raspberrypi/pico-sdk/issues/369 )

rtc/hello_rtc/hello_rtc.c
```diff
@@ -32,11 +32,14 @@ int main() {
...
     rtc_set_datetime(&t);
 
+    // clk_sys is >2000x faster than clk_rtc, so datetime is not transferred completely when rtc_get_datetime() is called
+    sleep_us(20);
+
     // Print the time
     while (true) {
         rtc_get_datetime(&t);
         datetime_to_str(datetime_str, sizeof(datetime_buf), &t);
-        printf("\r%s      ", datetime_str);
```
kilograham, line 39: why did you remove this? it is presumably there so that garbage isn't left when the datetime str gets shorger

kilograham (changes requested):

rtc/hello_rtc/hello_rtc.c
Reneg973: There may be garbage, but behind the \0 char. If there is no \0, there is a bug in datetime_to_str. Anyway a few spaces are not the right solution

kilograham: No, if it said 
```
mon 31 january
```
then
```
thu 1 march
```

you don't want it to leave the result (with `\r`)

```
thu 1 marchary
```

Adding spaces is a pertfectly acceptable way of dealing with it in something like this.

Reneg973: understood, for me it was strange to overwrite garbage which exists just in a few cases and otherwise adds space garbage.
For this example it may not be of interest => removed this change

rtc/hello_rtc/hello_rtc.c
```diff
@@ -32,6 +32,9 @@ int main() {
     rtc_init();
     rtc_set_datetime(&t);
 
+    // clk_sys is >2000x faster than clk_rtc, so datetime is not transferred completely when rtc_get_datetime() is called
```
lurch, line 35: Was the comment here supposed to say `rtc_set_datetime()` ?
And perhaps "not updated immediately" makes more sense than "not transferred completely" ?

rtc/hello_rtc/hello_rtc.c
lurch: @Reneg973 this is "overwriting garbage" in the output (i.e. serial terminal) that `printf` is writing to (because `\r` is being used), it's not "overwriting garbage" in the `datetime_str` itself :wink: 
It's kind of similar to doing:
```
printf("\r                                                                          "); // go to start of line, and overwrite it with spaces
printf("\r%s", datetime_str); // go to start of line, and print out datetime_str
```

Reneg973: I know, my issue was just that it appends spaces which are not required on most cases

rtc/hello_rtc/hello_rtc.c
Reneg973: Issue is that the set function is asynchronous, which is not expected. So the patch here is more or less a workaround. 

I am not a native english speaker, if you think your sentence makes more sense, I can take it. From my pov there are 2 controllers, the fast MCU sends multiple bytes, but the slow RTC didn't get all of it before the getter is called. 
I think a comment for the setter which describes this behavior would be good

lurch: Decision for @kilograham :wink:

kilograham (approved):

Issue #88, opened 2021-03-14
Incorrect blink frequency in pio_blink


sparkyhall:
The formula used in the recent update to use the sys clock is incorrect.

The offending line is the last line in blink.c:

`pio->txf[sm] = clock_get_hz(clk_sys) / 2 * freq;`

This divides sys clock by 2 and multiplies the result by the frequency desired. 

The result is that a frequency of 1 Hz correctly blinks at 1Hz, however, a frequency of 4Hz blinks at 0.25Hz and a frequency of 3Hz at 0.33Hz.

Replacing the multiply with divide corrects this.

rschaer:
Just noticed that too.` pio->txf[sm] = (clock_get_hz(clk_sys)/2) / freq; ` works fine.

Reneg973:
Or - if you include rounding:
`pio->txf[sm] = (clock_get_hz(clk_sys) + freq) / (2 * freq);`

lurch:
Fixed by #121

kilograham:
merged into develop (a long time ago)

Issue #108, opened 2021-04-25
pio blink example incorrect frequency


smist09:
The line in the C program is:

    pio->txf[sm] = clock_get_hz(clk_sys) / 2 * freq;

Should be:

    pio->txf[sm] = clock_get_hz(clk_sys) / freq;

lurch:
Fixed by #121

lurch:
This should probably be closed too @kilograham ?

kilograham:
closes as duplicate (which has also been merged)

PR #121, opened 2021-06-09
Fix pio_blink frequency calculation

urish:
Added missing parenthesis around the divisor. Verified using a logic analyzer and also in simulation:

https://wokwi.com/arduino/projects/300911723685085709


lurch:
Thanks. Does this fix #88 and #108 ?

lurch:
Also, can you please target this at `develop` instead of `master` ?

urish:
> Also, can you please target this at develop instead of master ?

My bad, fixed

> Thanks. Does this fix #88 and #108 ?

So it seems. #108 also removes the division by 2, but from my testing that generates the wrong frequency (e.g. the LED on pin 6 would toggle 4 times a second and thus would blink (on+off) only twice a second, so you'll get frequency of 2Hz instead of 4)

kilograham (approved):

PR #127, opened 2021-06-25
Additional BMP280 I2C example

matiasilva:
This PR adds a new example for the BMP280 temperature and pressure sensor via I2C. It takes care of reading sensor values and also the necessary conversion for the values to make sense. I'm not sure how much "cleaning up" is needed, for now all code is in one file for simplicity but a few definitions want to be in their own header file. Fritzing diagram and file included, accompanying documentation too.


i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
+/**
+ * Copyright (c) 2021 Raspberry Pi (Trading) Ltd.
```
andrum99, line 2: I'm assuming you're employed by Trading and have done this example as part of that work: otherwise this looks a bit odd.

i2c/CMakeLists.txt
```diff
@@ -2,4 +2,5 @@ if (NOT PICO_NO_HARDWARE)
     add_subdirectory(bus_scan)
     add_subdirectory(lcd_1602_i2c)
     add_subdirectory(mpu6050_i2c)
+    add_subdirectory(bmp280_i2c)
```
kilograham, line 5: we tend to keep subdirectories in alphabetical order (perhaps hard to infer from these 3 data points!)

i2c/bmp280_i2c/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(bmp280_i2c
+        bmp280_i2c.c
+        )
+
+# Pull in our (to be renamed) simple get you started dependencies
```
kilograham, line 5: oops; did "(to be rename) simple" come from another CMakeLists.txt? Either way, let's fix \

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+  var2 = ((((float)*temp) / 131072.0 - ((float)params->dig_t1) / 8192.0) *
+          (((float)*temp) / 131072.0 - ((float)params->dig_t1) / 8192.0)) *
+         ((float)params->dig_t3);
+  params->t_fine = (int32_t)(var1 + var2);
+  return (var1 + var2) / 5120.0;
+}
+
+float bmp280_convert_pressure(int32_t *pressure,
+                              struct bmp280_calib_param *params) {
+  float var1, var2;
```
kilograham, line 156: OK, so i'm not sure why you took their integer arithmetic and converted it to floating point... not sure this is clearer (which might be a reason in an example). Unnecessary floating point code should be avoided on device.

The only place you need a float is in the final return value (and even then if this was a library, i'd make a version which returned a fixed point integer, and a float wrapper that did the divide, so the caller could choose whether to touch floating point code at all - the (`-gc-sections`) linker would take care of not pulling in floating point code at all if the function is not called)

kilograham: Note also (orthogonal to above) that `2.0` etc are double constants not floating point constants, so you area also doing a bunch of float<>double conversions at compile/runtime (subject to specific compile options)

If you are doing floating point then use `2.0f`

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+                             GPIO_FUNC_I2C));
+  bi_decl(
+      bi_program_description("BMP280 I2C example for the Raspberry Pi Pico"));
+
+#if !defined(i2c_default) || !defined(PICO_DEFAULT_I2C_SDA_PIN) || \
+    !defined(PICO_DEFAULT_I2C_SCL_PIN)
+#warning i2c/bmp280_i2c example requires a board with I2C pins
+  puts("Default I2C pins were not defined");
+#else
+  printf(
```
kilograham, line 228: not a fan of splitting a single parameter to a newline

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+
+  uint8_t buf[NUM_CALIB_PARAMS] = {0};
+  i2c_write_blocking(i2c_default, (addr & BMP280_WRITE_MODE), &REG_DIG_T1_LSB,
+                     1, true);  // true to keep master control of bus
+  // read in one go as register addresses auto-increment
+  i2c_read_blocking(i2c_default, (addr & BMP280_READ_MODE), buf,
+                    NUM_CALIB_PARAMS, false);  // false, we're done reading
+
+  // store these in a struct for later use
+  params->dig_t1 = (uint16_t)(buf[1] << 8) | buf[0];
```
kilograham, line 197: I'm not sure there is any particular benefit to using individual fields rather than an array, and then you could have a loop. (the unsigned/signed could be handled in the conversion functions which are _magic_ anyway

kilograham: Note this would let you use a loop which is going to be much smaller code wise... of course feel free to plug the code into [compiler explorer](https://godbolt.org/) to see if gcc can/does optimize it back into a loop

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+   board GPIO PICO_DEFAULT_I2C_SCK_PIN (on Pico this is 5 (pin 7)) -> SCL on
+   BMP280 board 3.3v (pin 36) -> VCC on BMP280 board GND (pin 38)  -> GND on
+   BMP280 board
+*/
+
+// number of calibration registers to be read
+#define NUM_CALIB_PARAMS 24
+
+// device has default bus address of 0x76
+const uint8_t addr = 0x76;
```
kilograham, line 33: I don't feel strongly, but curious why you chose `const uin8_t` over `#defines` for these

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+  int16_t dig_p3;
+  int16_t dig_p4;
+  int16_t dig_p5;
+  int16_t dig_p6;
+  int16_t dig_p7;
+  int16_t dig_p8;
+  int16_t dig_p9;
+
+  // intermediate result for pressure calc
+  int32_t t_fine;
```
kilograham, line 95: I'm not a huge fan of this, as it isn't a calib param, and it introduces an undocumented dependency between the two functions... i'd just calculate it twice, or otherwise re-arrange the functions

kilograham (commented):
Overall I like this a *lot* even though i did make a bunch of comments

i2c/bmp280_i2c/CMakeLists.txt
lurch: https://github.com/raspberrypi/pico-examples/search?q=renamed

i2c/bmp280_i2c/README.adoc
```diff
@@ -0,0 +1,41 @@
+= Attaching a BMP280 temp/pressure sensor via I2C
+
+This example code shows how to interface the Raspberry Pi Pico with the popular BMP280 temperature and air pressure sensor manufactured by Bosch. A similar variant, the BME280, exists that can also measure humidity. There is another example that uses this device but talks to it via SPI as opposed to I2c.
```
lurch, line 3: I2c - > I2C

In the last sentence, is "this device" referring to BMP280 or BME280?

i2c/bmp280_i2c/README.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+The code reads data from the sensor's registers every second and prints it via the onboard UART. This example operates the BMP280 in _normal_ mode, meaning that the device continuously cycles between a measurement period and a standby period at a regular interval we can set. This has the advantage that subsequent reads do not require configuration register writes and is the recommended mode of operation to filter out short-term disturbances.
+
+[TIP]
+======
+The BMP280 is highly configurable with 3 modes of operation, various oversampling levels, and 5 filter settings. Find the datasheet online to explore all of its capabilities beyond the simple example given here.
```
lurch, line 9: "Find the datasheet online" - is there a canonical URL for the datasheet that we can link to?

i2c/bmp280_i2c/README.adoc
```diff
@@ -0,0 +1,41 @@
...
+WARNING: The BMP280 has a maximum supply voltage rating of 3.6V. Most breakout boards have voltage regulators that will allow a range of input voltages of 2-6V, but make sure to check beforehand.
+
+[[bmp280_i2c_wiring]]
+[pdfwidth=75%]
+.Wiring Diagram for BMP280 sensor via I2C.
+image::bmp280_i2c_bb.png[]
+
+== List of Files
+
+CMakeLists.txt:: CMake file to incorporate the example in to the examples build tree.
```
lurch, line 25: "in to" -> "into"?

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+
+   You will need to use a level shifter on the I2C lines if you want to run the
+   board at 5v.
+
+   Connections on Raspberry Pi Pico board, other boards may vary.
+
+   GPIO PICO_DEFAULT_I2C_SDA_PIN (on Pico this is 4 (pin 6)) -> SDA on BMP280
+   board GPIO PICO_DEFAULT_I2C_SCK_PIN (on Pico this is 5 (pin 7)) -> SCL on
+   BMP280 board 3.3v (pin 36) -> VCC on BMP280 board GND (pin 38)  -> GND on
+   BMP280 board
```
lurch, line 26: Looks like the line-breaks here are in the wrong places? :confused:

i2c/bmp280_i2c/README.adoc
aallan: https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bmp280-ds001.pdf

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+  int16_t dig_p6;
+  int16_t dig_p7;
+  int16_t dig_p8;
+  int16_t dig_p9;
+
+  // intermediate result for pressure calc
+  int32_t t_fine;
+};
+
+#ifdef i2c_default
```
lurch, line 98: Do we need this `#ifdef` ?

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+void bmp280_init() {
+  // use the "handheld device dynamic" optimal setting (see datasheet)
+  uint8_t buf[2];
+
+  // 500ms sampling time, x16 filter, not set
+  uint8_t reg_config_val = ((0x04 << 5) | (0x05 << 2)) & 0xFC;
```
lurch, line 104: Looks like this (and `reg_ctrl_meas_val`) are effectively consts?

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+  // osrs_t x1, osrs_p x4, normal mode operation
+  uint8_t reg_ctrl_meas_val = (0x01 << 5) | (0x03 << 2) | (0x03);
+  buf[0] = REG_CTRL_MEAS;
+  buf[1] = reg_ctrl_meas_val;
+  i2c_write_blocking(i2c_default, (addr & BMP280_WRITE_MODE), buf, 2, false);
+}
+
+void bmp280_read_raw(int32_t *temp, int32_t *pressure) {
+  // BMP280 data registers are auto-incrementing and we have 3 temperature and
+  // pressure registers, each so we start at 0xF7 and read 6 bytes to 0xFC note:
```
lurch, line 121: "registers, each so" -> "registers each, so" ?

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+}
+
+float bmp280_convert_pressure(int32_t *pressure,
+                              struct bmp280_calib_param *params) {
+  float var1, var2;
+  float converted = 0.0;
+
+  var1 = ((float)params->t_fine / 2.0) - 64000.0;
+  var2 = var1 * var1 * ((float)params->dig_p6) / 32768.0;
+  var2 = var2 + var1 * ((float)params->dig_p5) * 2.0;
```
lurch, line 161: Only a minor nit-pick, but `var2 = var2 + ...` could be `var2 += ...`

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+
+float bmp280_convert_temp(int32_t *temp, struct bmp280_calib_param *params) {
+  // use the 32-bit floating point compensation implementation given in the
+  // datasheet
+  float var1, var2;
+  var1 = (((float)*temp) / 16384.0 - ((float)params->dig_t1) / 1024.0) *
+         ((float)params->dig_t2);
+  var2 = ((((float)*temp) / 131072.0 - ((float)params->dig_t1) / 8192.0) *
+          (((float)*temp) / 131072.0 - ((float)params->dig_t1) / 8192.0)) *
+         ((float)params->dig_t3);
```
lurch, line 149: You seem to be calculating `(((float)*temp) / 131072.0 - ((float)params->dig_t1) / 8192.0)` twice here - is that correct? :shrug:

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Ah, fixed.

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: This was bugging me as well, but was waiting for more things to touch up xD

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Changed

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: I got this from the other I2C examples. It did take me a bit to understand what it was doing, but I assumed it was there for a good reason so let it be.

i2c/bmp280_i2c/README.adoc
matiasilva: Also fixed!

i2c/bmp280_i2c/README.adoc
matiasilva: Done!

i2c/bmp280_i2c/README.adoc
matiasilva: Clarified this

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: I've now split the pressure and temperature calculation into two steps. How does that look?

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: I wasn't sure how to use `#define`s and hex numbers. I think there's a macro but I wasn't sure..

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,260 @@
...
+  int32_t pressure;
+
+  while (1) {
+    bmp280_read_raw(&temp, &pressure);
+    float converted_temp = bmp280_convert_temp(&temp, &params);
+    float converted_pressure = bmp280_convert_pressure(&pressure, &params);
+    printf("Temperature: %.2f °C  Pressure: %.3f kPa\n", converted_temp,
+           converted_pressure);
+    // poll every 750ms as data refreshes every 500ms
+    sleep_ms(750);
```
lurch, line 268: In the README.adoc you wrote "The code reads data from the sensor's registers every second and prints it via the onboard UART." - should that instead say every 750ms ?

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: I chose to go for individual fields for clarity so the person reading the code could see exactly where each parameter was being stored. I think hiding all that in an array would be good code-wise but would seem a bit cryptic with magic indices (unless we used `#define`s for the indices), not sure.

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: I hadn't set up formatting then, I've disabled wrapping now so hopefully this shouldn't happen again

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Ah, all good points! They had fixed and floating point implementations. No particular reason why I went for floating point but I agree it looks a bit clunky. I've swapped in the fixed point implementation.

i2c/bmp280_i2c/CMakeLists.txt
matiasilva: Got this from the other example's CMakeLists.txt and fixed here.

i2c/CMakeLists.txt
matiasilva: Ah, yes. Fixed.

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: I noticed this while fixing up a few of the earlier suggestions, but either way done!

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Yes, it seem so. The floating point implementation from Bosch had this exact line. I've swapped it for the fixed point so this doesn't exist anymore anyway.

i2c/bmp280_i2c/bmp280_i2c.c
lurch: Ahhh, it might just be to stop the compiler warning about unused functions? :)

i2c/bmp280_i2c/bmp280_i2c.c
lurch: Perhaps it's worth calculating `t_fine` just once (in the main loop) and passing it as a a parameter to the `bmp280_convert_temp` and `bmp280_convert_pressure` functions? :shrug:

matiasilva:
> Perhaps it's worth changing your wiring diagram to use a half-size breadboard (as you've only got a very limited number of wires), like the other examples in Appendix A of https://datasheets.raspberrypi.org/pico/raspberry-pi-pico-python-sdk.pdf ?

I've done so anyway and I think it looks quite nice!

lurch:
I know this is intended to be a standalone example, but I wonder if there's any value in trying to keep the "I2C BMP280" and "SPI BME280" examples vaguely similar-ish? Just as one example, the SPI BME280 example does:
```c
printf("Humidity = %.2f%%\n", humidity / 1024.0);
printf("Pressure = %dPa\n", pressure);
printf("Temp. = %.2fC\n", temperature / 100.0);
```
whereas the I2C BMP280 example does:
```c
    float converted_temp = bmp280_convert_temp(&temp, &params) * 0.01f;
    float converted_pressure = bmp280_convert_pressure(&pressure, &temp, &params) * 0.001f;
    printf("Temperature: %.2f °C  Pressure: %.3f kPa\n", converted_temp,
      converted_pressure);
```

lurch:
On an additional side-note @aallan , I also just noticed that the Raspberry Pi Pico PCB images in Appendix A of https://datasheets.raspberrypi.org/pico/raspberry-pi-pico-c-sdk.pdf look less detailed than their counterparts in Appendix A of https://datasheets.raspberrypi.org/pico/raspberry-pi-pico-python-sdk.pdf ?

aallan:
Yes. They need replacing with the new Fritzing part. Most of them were done pre-launch.

i2c/bmp280_i2c/bmp280_i2c.c
kilograham: yeah i don't feel strongly; my usual self is very at odds with the way we want to write examples. I just wasn't sure how much more helpful these numbers were, and yes I was thinking #defines for array indexes perhaps, but then again I haven't read the data sheet and therefore perhaps these are all specifically heralded.

i2c/bmp280_i2c/bmp280_i2c.c
kilograham: Yeah, we use `_u(3)` or whatever in the SDK in case the constant is included by assembly, which doesn't understand `3u` - i.e. we define `_u` to either suffix `u` or not based on where we are.

Note also, we tend to use `u` suffixes directly for shifts etc because higher warning levels complain otherwise, but i shut up about that because again `3u` may confuse some examplees, and so we don't require quietness on such high warning levels for the examples

i2c/bmp280_i2c/bmp280_i2c.c
kilograham: well that function _also_ uses i2c_default as a parameter so wouldn't compile.

but for context, depending on your PICO_BOARD setting you may be using a board with no I2C pins defined (or indeed exposed at all) in which case the example makes little sense.

matiasilva: Shall we remove it then?

lurch: Nope, you need to leave it in. I hadn't spotted that the functions inside this `#ifdef` were using `i2c_default`. (sorry for causing confusion)

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Fair, of note that the BME280 SPI example does something similar, which I only found out after I wrote this example but interesting coincidence.

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: I thought about that, but as Graham mentioned that creates an undocumented dependency on this intermediate parameter t_fine. I think it's more user friendly for each pressure and temperature function to just do its job without relying on t_fine as a parameter. Perhaps those kinds of optimizations are better left to the user when they take this example code and use it in their applications?

kilograham: nah, i'd vote against that

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+  *temp = (buf[3] << 12) | (buf[4] << 4) | (buf[5] >> 4);
+}
+
+void bmp280_reset() {
+  // reset the device with the power-on-reset procedure
+  uint8_t buf[2] = { REG_RESET, 0xB6 };
+  i2c_write_blocking(i2c_default, (addr & BMP280_WRITE_MODE), buf, 2, false);
+}
+
+int32_t bmp280_convert(int32_t* temp, struct bmp280_calib_param* params) {
```
kilograham, line 138: For this and the next 3 functions, why are we passing pointers to int32_t ... these are not in/out vars (I don't think)
If they were to remain pointers, they should be `const` pointers, but please change them to not be const pointers at all.

Note you can try doing
`target_compile_options(bmp280_i2c -Wall)` in your CMakeLists.txt just to see what sort of warnings you may get from the compiler

kilograham (changes requested):

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Done

matiasilva:
I think this should be ready now. I also changed the last few lines in the loop to match the BME280 example as I think the way it's presented there is clearer to the user.

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+  // used for both pressure and temperature conversions
+  int32_t var1, var2;
+  var1 = ((((temp >> 3) - ((int32_t)params->dig_t1 << 1))) * ((int32_t)params->dig_t2)) >> 11;
+  var2 = (((((temp >> 4) - ((int32_t)params->dig_t1)) * ((temp >> 4) - ((int32_t)params->dig_t1))) >> 12) * ((int32_t)params->dig_t3)) >> 14;
+  return var1 + var2;
+}
+
+int32_t bmp280_convert_temp(int32_t temp, struct bmp280_calib_param* params) {
+  // use the 32-bit fixed point compensation implementation given in the
+  // datasheet
```
lurch, line 149: Is this comment now in the wrong place?

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+  var2 = (((var1 >> 2) * (var1 >> 2)) >> 11) * ((int32_t)params->dig_p6);
+  var2 += ((var1 * ((int32_t)params->dig_p5)) << 1);
+  var2 = (var2 >> 2) + (((int32_t)params->dig_p4) << 16);
+  var1 = (((params->dig_p3 * (((var1 >> 2) * (var1 >> 2)) >> 13)) >> 3) + ((((int32_t)params->dig_p2) * var1) >> 1)) >> 18;
+  var1 = ((((32768 + var1)) * ((int32_t)params->dig_p1)) >> 15);
+  if (var1 == 0) {
+    return 0;  // avoid exception caused by division by zero
+  }
+  converted =
+    (((uint32_t)(((int32_t)1048576) - pressure) - (var2 >> 12))) * 3125;
```
lurch, line 169: Presumably there's no reason to split this over two lines?

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+  if (converted < 0x80000000) {
+    converted = (converted << 1) / ((uint32_t)var1);
+  }
+  else {
```
lurch, line 173: ping @kilograham What bracket-formatting-style should be used here?

kilograham: `} else {`

kilograham:
note also it looks like you are using 2 spaces for tabs... we use 4

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+  int32_t raw_pressure;
+
+  while (1) {
+    bmp280_read_raw(&raw_temperature, &raw_pressure);
+    float temperature = bmp280_convert_temp(raw_temperature, &params);
+    float pressure = bmp280_convert_pressure(raw_pressure, raw_temperature, &params);
+    printf("Pressure = %.3f kPa\n", pressure / 1000.0);
+    printf("Temp. = %.2f C\n", temperature / 100.0);
+    // poll every 750ms as data refreshes every 500ms
+    sleep_ms(750);
```
lurch, line 249: I don't have a BMP280 myself, and I've not read the datasheet, but is there a good reason for having the sleep-interval in the program different to the data-refresh-interval of the sensor? :shrug:

aallan: > I don't have a BMP280 myself, and I've not read the datasheet, but is there a good reason for having the sleep-interval in the program different to the data-refresh-interval of the sensor? 🤷

My assumption was so that it didn't hit the refresh boundary, although it might be better not to thave the update time a divisor of the refresh time at that point.

i2c/bmp280_i2c/README.adoc
```diff
@@ -0,0 +1,41 @@
...
+== Bill of Materials
+
+.A list of materials required for the example
+[[bmp280_i2c-bom-table]]
+[cols=3]
+|===
+| *Item* | *Quantity* | Details
+| Breadboard | 1 | generic part
+| Raspberry Pi Pico | 1 | http://raspberrypi.org/
+| BMP280-based breakout board | 1 | generic part
```
aallan, line 37: We used the [Pimoroni part](https://shop.pimoroni.com/products/bmp280-breakout-temperature-pressure-altitude-sensor), we should acknowledge that and link to it here. Although what Fritzing part did you use for the wiring diagram?

i2c/bmp280_i2c/bmp280_i2c.c
lurch: Yeah, that's why I asked, since 2 * 0.75 == 3 * 0.5 :wink: 
I guess if you wanted to keep them deliberately non-overlapping you could do something like:
```
start sensor measuring every 500ms
sleep for 250ms
start loop
    read sensor values
    sleep for 500ms
end loop
```
?

matiasilva: I just thought that if the sensor updates every 500ms and we poll every 500ms there might be a mismatch in timing. Polling halfway through the update cycle means we always get new data...but I'm not sure

i2c/bmp280_i2c/README.adoc
matiasilva: I tried but couldn't find a Fritzing part for the wiring diagram, so I used the closest one I could find, which was just a generic BME/P280 breakout.

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: I've changed to #defines so that it's consistent with other examples I'll be uploading from now.

i2c/bmp280_i2c/bmp280_i2c.c
aallan: But,

500+500+500 = 1,500 ms
750+750 = 1,500 ms

so you still get a sensor mismatch.

matiasilva: Ah, that makes..a lot of sense 😅 Changed now.

README.md
```diff
@@ -76,6 +76,7 @@ App|Description
 App|Description
 ---|---
 [bus_scan](i2c/bus_scan) | Scan the I2C bus for devices and display results.
+[bmp280_i2c](i2c/bmp280_i2c) | Read and convert temperature and pressure data from a BMP280 sensor, attached to an I2C bus.
```
lurch, line 79: `bm` is alphabetically before `bu` ?

matiasilva: Indeed it is 😅

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+    struct bmp280_calib_param params;
+    bmp280_get_calib_params(&params);
+
+    int32_t raw_temperature;
+    int32_t raw_pressure;
+
+    sleep_ms(250); // sleep so that data polling and register update don't collide
+    while (1) {
+        bmp280_read_raw(&raw_temperature, &raw_pressure);
+        float temperature = bmp280_convert_temp(raw_temperature, &params);
```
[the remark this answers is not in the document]
matiasilva: Fixed!

matiasilva:
I've only realized now after some head scratching on another board, that the SDK's `i2c_write_blocking` assumes 7-bit addresses and adds the read/write bits for us. Last commit reflects this.

lurch:
> I've only realized now after some head scratching on another board, that the SDK's i2c_write_blocking assumes 7-bit addresses and adds the read/write bits for us. Last commit reflects this.

I know very little about I2C, but does the same thing affect https://github.com/raspberrypi/pico-examples/pull/143 ? (please leave an appropriate comment on that PR if it does)

README.md
lurch: Actually, while it makes sense for the files listed in `CMakeLists.txt` to be sorted alphabetically, it probably makes sense for the files listed in `README.md` to be sorted in order of increasing complexity? :thinking: 
In which case, `bus_scan` should indeed be listed before `bmp280_i2c` - sorry for changing my mind! :laughing:

i2c/bmp280_i2c/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(bmp280_i2c
+        bmp280_i2c.c
+        )
+
+# Pull in our get you started dependencies
```
lurch, line 5: Needs tweaking :wink:

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,253 @@
...
+    int16_t dig_p8;
+    int16_t dig_p9;
+};
+
+#ifdef i2c_default
+void bmp280_init() {
+    // use the "handheld device dynamic" optimal setting (see datasheet)
+    uint8_t buf[2];
+
+    // 500ms sampling time, x16 filter, not set
```
lurch, line 99: What does the "not set" here mean? :confused:

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,253 @@
...
+    const uint8_t reg_config_val = ((0x04 << 5) | (0x05 << 2)) & 0xFC;
+
+    // LSB of slave address sets read or write mode
```
lurch, line 102: Is this comment redundant, based on your last commit?

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,253 @@
...
+    const uint8_t reg_ctrl_meas_val = (0x01 << 5) | (0x03 << 2) | (0x03);
+    buf[0] = REG_CTRL_MEAS;
+    buf[1] = reg_ctrl_meas_val;
+    i2c_write_blocking(i2c_default, ADDR, buf, 2, false);
+}
+
+void bmp280_read_raw(int32_t* temp, int32_t* pressure) {
+    // BMP280 data registers are auto-incrementing and we have 3 temperature and
+    // pressure registers each, so we start at 0xF7 and read 6 bytes to 0xFC note:
+    // normal mode does not require further ctrl_meas and config register writes
```
lurch, line 118: Perhaps the `note:` should be moved from the end of the line to the start of the next line?

i2c/bmp280_i2c/bmp280_i2c.c
```diff
@@ -0,0 +1,253 @@
...
+
+void bmp280_reset() {
+    // reset the device with the power-on-reset procedure
+    uint8_t buf[2] = { REG_RESET, 0xB6 };
+    i2c_write_blocking(i2c_default, ADDR, buf, 2, false);
+}
+
+int32_t bmp280_convert(int32_t temp, struct bmp280_calib_param* params) {
+    // intermediate function that calculates the fine resolution temperature
+    // used for both pressure and temperature conversions
```
lurch, line 137: I think stylistically it makes sense for comments describing what a function does, to appear above the function signature, with (any) comments inside the function-body explaining what the code is doing? :shrug:

i2c/bmp280_i2c/CMakeLists.txt
matiasilva: lol...

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Ah, perhaps not very clear. Basically the 8 bits configure 3 different settings, and the last one is not set (ie. left to its reset value). I'll remove it, it just confuses people not familiar with this.

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Indeed!

i2c/bmp280_i2c/bmp280_i2c.c
matiasilva: Agreed!

matiasilva:
> I know very little about I2C, but does the same thing affect #143 ? (please leave an appropriate comment on that PR if it does)

Done!

i2c/bmp280_i2c/CMakeLists.txt
lurch: It's like a virus that keeps popping back up every time you think you've killed it off :rofl:

i2c/bmp280_i2c/bmp280_i2c.c
lurch: This comment also applies to e.g. `bmp280_convert_temp` and `bmp280_convert_pressure` ? :wink:

lurch:
I added a comment about `README.md` but as it's in a "resolved" section, GitHub is helpfully auto-hiding it! :confused: So here it is again:

Actually, while it makes sense for the files listed in `CMakeLists.txt` to be sorted alphabetically, it probably makes sense for the files listed in `README.md` to be sorted in order of increasing complexity? :thinking: 
In which case, `bus_scan` should indeed be listed before `bmp280_i2c` - sorry for changing my mind! :laughing:

matiasilva:
> I added a comment about `README.md` but as it's in a "resolved" section, GitHub is helpfully auto-hiding it! confused So here it is again:
> 
> Actually, while it makes sense for the files listed in `CMakeLists.txt` to be sorted alphabetically, it probably makes sense for the files listed in `README.md` to be sorted in order of increasing complexity? thinking
> In which case, `bus_scan` should indeed be listed before `bmp280_i2c` - sorry for changing my mind! laughing

Ah, yeah I didn't see that, thanks for bringing it up again. I agree, and perhaps we should indicate this in a little comment too?

aallan:

aallan:
> There is an outstanding change request flagged, but I believe it has been fixed but I cannot see a way of flagging it has been fixed.

I think we'd just have to "re-request review" from @kilograham. There [seems to be no way](https://github.community/t/github-changes-requested-in-an-pr-completely-misleads/799) to get rid of the flag otherwise.

PR #49, opened 2021-02-09
Add a SSD1306 I2C example

JamesH65:
This is a 128x64 monochrome display, 0.96" diagonal so very small.

This example initialises the display, and implements basic pixel, line drawing and
text output.

PR #130, opened 2021-06-30
Add example for OLED 128x32 display via I2C

matiasilva:
This PR adds an example for a 128x32 pixel OLED display with an SSD1306 driver. Demo below:

gif

## Checklist

- [x]  documentation
- [x] working example
- [x] Fritzing schematic and image
- [x] added to CMakeLists.txt


i2c/oled_i2c/README.adoc
```diff
@@ -0,0 +1,68 @@
+= Attaching an OLED display via I2C
+
+This example code shows how to interface the Raspberry Pi Pico with an 128x32 OLED display board based on the SSD1306 display driver.
+
+The code displays a series of tiny raspberries that scroll horizontally, in the process showing you how to initialize the display, write to the entire display, write to only a portion of the display, and configure scrolling.
+
+In contrast to the sensors in the examples library, the SSD1306 is not operated via direct reads/writes to registers. Instead, a list of versatile commands is defined (see datasheet) that allows the user to access all the capabilities of the driver. Ultimately, these commands end up writing to internal registers but the key point is that the I2C write process is now different. After sending a slave address, the data that follows can be either a command, flags to follow up a command or data to be written directly into the display's RAM. A control byte is required for each write after the slave address so that the driver knows what type of data is being sent.
```
aallan, line 7: Can you link to the datasheet? It's an Adafruit part we used, so it's probably fine to use the Adafruit hosted version at https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf.

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,301 @@
...
+
+    GPIO PICO_DEFAULT_I2C_SDA_PIN (on Pico this is 4 (pin 6)) -> SDA on display
+    board
+    GPIO PICO_DEFAULT_I2C_SCK_PIN (on Pico this is 5 (pin 7)) -> SCL on
+    display board
+    3.3v (pin 36) -> VCC on display board
+    GND (pin 38)  -> GND on display board
+ */
+
+#define OLED_RESET_PIN 28
```
aallan, line 31: Unless I'm misunderstanding what's going on here, the wiring diagram has the RST pin on the display wired to GP15? We tend to avoid GP15 due to RP2040-E5, but in any case you're setting it to GP28 here?

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,301 @@
...
+
+    NOTE: Ensure the device is capable of being driven at 3.3v NOT 5v. The Pico
+    GPIO (and therefore I2C) cannot be used at 5v.
+
+    You will need to use a level shifter on the I2C lines if you want to run the
+    board at 5v.
+
+    Connections on Raspberry Pi Pico board, other boards may vary.
+
+    GPIO PICO_DEFAULT_I2C_SDA_PIN (on Pico this is 4 (pin 6)) -> SDA on display
```
aallan, line 23: We tend to refer to "GP4" rather than just use "4", and then "physical pin 6" if we referring to the physical numbering of pins. So that'd be "on Raspberry Pi Pico this is GP4 (physical pin 6)".

i2c/oled_i2c/README.adoc
```diff
@@ -0,0 +1,68 @@
...
+== Bill of Materials
+
+.A list of materials required for the example
+[[oled_i2c-bom-table]]
+[cols=3]
+|===
+| *Item* | *Quantity* | Details
+| Breadboard | 1 | generic part
+| Raspberry Pi Pico | 1 | http://raspberrypi.org/
+| OLED display board | 1 | generic part
```
aallan, line 66: We're using the Adafruit part, and their Fritzing part in the wiring diagram so we should link to the Adafruit part in the BoM. Which I think is this one? https://www.adafruit.com/product/4440

i2c/oled_i2c/oled_i2c.c
lurch: I think this comment also applies to your BMP280 PR?

lurch:
Is this similar to #49 ? Or does this PR now make #49 redundant? :man_shrugging:

i2c/oled_i2c/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(oled_i2c
+        oled_i2c.c
+        )
+
+# Pull in our (to be renamed) simple get you started dependencies
```
lurch, line 5: In the BMP280 PR didn't we decide to change this line? :wink:

i2c/oled_i2c/README.adoc
lurch, line 7: I'm not sure if the "In contrast to the sensors in the examples library" verbiage is necessary - it seems to assume that the user has looked at the sensors example first, whereas they're more likely to just be jumping straight into this example?

i2c/oled_i2c/README.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+The code displays a series of tiny raspberries that scroll horizontally, in the process showing you how to initialize the display, write to the entire display, write to only a portion of the display, and configure scrolling.
+
+In contrast to the sensors in the examples library, the SSD1306 is not operated via direct reads/writes to registers. Instead, a list of versatile commands is defined (see datasheet) that allows the user to access all the capabilities of the driver. Ultimately, these commands end up writing to internal registers but the key point is that the I2C write process is now different. After sending a slave address, the data that follows can be either a command, flags to follow up a command or data to be written directly into the display's RAM. A control byte is required for each write after the slave address so that the driver knows what type of data is being sent.
+
+This display is 32 pixels high by 128 pixels wide. These 32 pixels are partitioned into 4 pages, each 8 bits in height. In RAM, This looks roughly like:
+
+----
+           | COL0 | COL1 | COL2 | COL3 | ...
```
lurch, line 12: Does this actually extend out to 128 columns? If so, that should perhaps be clarified like:
```
           | COL0 | COL1 | COL2 | COL3 | ... | COL126 | COL127 |
```
?
And perhaps "These 32 pixels" should be "These 32 vertical pixels" ? And "8 bits in height" -> "8 pixels in height" ? :shrug:

i2c/oled_i2c/README.adoc
```diff
@@ -0,0 +1,68 @@
...
+----
+          | COL0 | COL1 | COL2 | COL3 | ...
+    COM 0 |      |      |      |      |
+    COM 1 |      |      |      |      |
+       :  |      |      |      |      |
+    COM 7 |      |      |      |      |
+    ------------------------------------
+----
+
+Note the difference between columns in RAM and the actual segment pins on the display. The distinction between these two is important as we can remap, or add offsets between the two in the driver itself instead of moving pixels in our frame buffer code.
```
lurch, line 31: "actual segment pins"? Seems unclear.

i2c/oled_i2c/README.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+The driver has 3 modes of transferring the pixels in RAM to the display on clock cycle (provided that the driver is set to use its RAM content to drive the display, ie. command 0xA4 is sent). We choose horizontal addressing mode which, after setting the column address and page address registers to our desired start positions, will increment the column address register until the OLED display width is reached (127 in our case) after which the column address register will reset to its starting value and the page address is incremented. Once the page register reaches the end, it will wrap around as well. Effectively, this scans across the display from top to bottom, left to right in blocks that are 8 pixels high. When a byte is sent to be written into RAM, it sets all the rows for the current position of the column address register. So, if we send 10101010, and we are on PAGE 0 and COL1, COM0 is set to 1, COM1 is set to 0, COM2 is set to 1, and so on. Effectively, the byte is "transposed" to fill a single page's column. The datasheet has further information on this and the two other modes.
```
lurch, line 33: Is "on clock cycle" necessary in the first sentence?

i2c/oled_i2c/README.adoc
lurch: Also, I suspect "OLED display board" is a **much** too generic description, as there are probably many many different OLED display boards? :laughing:

aallan:
> Is this similar to #49 ? Or does this PR now make #49 redundant? 🤷‍♂️

Oops. I'd forgotten about #49, I guess so did @JamesH65. 🤦‍♂️ 

We should probably look through the rest of the open PRs and make sure we haven't forgotten anything else that's now on the list of hardware we're intending to get @matiasilva and Aayushi to support.

i2c/oled_i2c/oled_i2c.c
lurch: This should also be mentioned in the "Connections on Raspberry Pi Pico board" comment above.

i2c/oled_i2c/README.adoc
aallan: Many! That it's a SSD1306-driven board is probably more of a differentiator. Hence I think we should link to the specific part in the BoM, with discussion that the example code will probably work for other SSD1306-based displays in the main text.

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,301 @@
...
+#define OLED_SET_VCOM_DESEL _u(0xDB)
+#define OLED_SET_CHARGE_PUMP _u(0x8D)
+#define OLED_SET_HORIZ_SCROLL _u(0x26)
+#define OLED_SET_SCROLL _u(0x2E)
+
+#define OLED_ADDR _u(0x3C)
+#define OLED_HEIGHT _u(32)
+#define OLED_WIDTH _u(128)
+#define OLED_PAGE_HEIGHT _u(8)
+#define OLED_PAGES_NUM OLED_HEIGHT / OLED_PAGE_HEIGHT
```
lurch, line 58: Doesn't really matter, but I think `OLED_NUM_PAGES` might be clearer than `OLED_PAGES_NUM` ?

matiasilva:
Oh, wow I completely missed that other PR. Is there anything we should bring from that one to this one?

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,301 @@
...
+void fill(uint8_t buf[], uint8_t fill) {
+    // fill entire buffer with one byte
+    for (int i = 0; i < OLED_BUF_LEN; i++) {
+        buf[i] = fill;
+    }
+};
+
+void fill_page(uint8_t buf[], uint8_t fill, uint8_t page) {
+    // fill entire page with one byte
+    for (int i = page * OLED_WIDTH; i < OLED_WIDTH; i++) {
```
lurch, line 82: This doesn't look right? If `page` is 2, then we'll have
```
for (int i = 256; i < 128; i++) {
```
and the loop will never iterate?

I suspect you might have wanted
```
for (int i = 0; i < OLED_WIDTH; i++) {
   buf[(page * OLED_WIDTH) + i] = fill;
}
```
(assuming I've correctly understood the addressing-modes, of course!)
Or you could probably do something with pointer-arithmetic to (effectively) do `buf[page * OLED_WIDTH]` just once before the loop, but my low-level C is a bit rusty.

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,301 @@
...
+    uint8_t start_col;
+    uint8_t end_col;
+    uint8_t start_page;
+    uint8_t end_page;
+
+    int buflen;
+};
+
+void fill(uint8_t buf[], uint8_t fill) {
+    // fill entire buffer with one byte
```
lurch, line 74: Perhaps change "with one byte" to "with same byte"? (and also for next function)

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,301 @@
...
+    // prints all pages of a full length buffer
+    for (int i = 0; i < OLED_PAGES_NUM; i++) {
+        printf("--page %d--\n", i);
+        print_buf_page(buf, i);
+    }
+}
+
+void print_buf_area(uint8_t buf[], struct render_area* area) {
+    // print a render area of generic size
+    uint8_t area_width = area->end_col - area->start_page + 1;
```
lurch, line 110: `area->end_col - area->start_col + 1;` ?

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,301 @@
...
+    uint8_t area_height = area->end_page - area->start_page + 1; // in pages, not pixels
+    for (int i = 0; i < area_height; i++) {
+        for (int j = 0; j < OLED_PAGE_HEIGHT; j++) {
+            for (int k = 0; k < area_width; k++) {
+                printf("%u", (buf[i * area_width + k] >> j) & 0x01);
+            }
+            printf("\n");
+        }
+    }
```
lurch, line 119: I not sure, but my spidey-sense tells me this might not work properly if `start_page` or `start_col` are non-zero?

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,301 @@
...
+void oled_send_cmd(uint8_t cmd) {
+    // I2C write process expects a control byte followed by data
+    // this "data" can be a command or data to follow up a command
+
+    // Co = 1, D/C = 0 => the driver expects a command
+    uint8_t buf[2] = { 0x80, cmd };
+    i2c_write_blocking(i2c_default, (OLED_ADDR & OLED_WRITE_MODE), buf, 2, false);
+}
+
+void oled_send_buf(uint8_t buf[], int BUF_LEN) {
```
lurch, line 138: Conventionally, ALL_CAPS is used for constants and mixed_case is used for variables and parameters, so `BUF_LEN` here should be `buf_len`.

i2c/oled_i2c/raspberry26x32.h
```diff
@@ -0,0 +1,4 @@
+#define IMG_WIDTH 26
+#define IMG_HEIGHT 32
+
+static uint8_t raspberry26x32[] = { 0x0, 0x0, 0xe, 0x7e, 0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xfe, 0xfe, 0xfc, 0xf8, 0xfc, 0xfe, 0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xfe, 0x7e, 0x1e, 0x0, 0x0, 0x0, 0x80, 0xe0, 0xf8, 0xfd, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xfd, 0xf8, 0xe0, 0x80, 0x0, 0x0, 0x1e, 0x7f, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f, 0x1e, 0x0, 0x0, 0x0, 0x3, 0x7, 0xf, 0x1f, 0x1f, 0x3f, 0x3f, 0x7f, 0xff, 0xff, 0xff, 0xff, 0x7f, 0x7f, 0x3f, 0x3f, 0x1f, 0x1f, 0xf, 0x7, 0x3, 0x0, 0x0};
```
lurch, line 4: Was this generated by some external tool? Might be nice to add details to the README so that users can use their own logo if they want?

matiasilva: I wrote a quick Python script and it did the job but it's not very pretty xD

aallan: It's an SVG(?) image converted to Base 64, right? In which case there's almost certainly an online tool we can link to, some of them might even be hosted somewhere we'd want to (be willing to) link to...

lurch: AFAICT it's an SVG image rendered out to a 26*32 monochrome bitmap, bit-swizzled to the page/column formatting suitable for the SSD1306 chip, and then encoded as a C-array of hex-bytes?

aallan: That might be harder to find an online tool for… 🤔

i2c/oled_i2c/oled_i2c.c
matiasilva: Oh whoops! I think this lingered from the brief moment when I considered making this a #define.

i2c/oled_i2c/README.adoc
matiasilva: Done!

i2c/oled_i2c/oled_i2c.c
matiasilva: The reset pin never ends up being used in the code, as our board  auto-resets on startup. The reset pin is mostly useful for shortcutting to sensible defaults for various configs and as we set all of them explicitly, I'll just mention it in passing in the README but remove it from the code and Fritzing diagram.

i2c/oled_i2c/oled_i2c.c
matiasilva: Agreed..fixed both.

i2c/oled_i2c/README.adoc
matiasilva: Makes sense. I wasn't sure what the policy was on mentioning third parties in the docs so I left it deliberately ambiguous. Fixed now.

i2c/oled_i2c/CMakeLists.txt
matiasilva: Grrr...

i2c/oled_i2c/README.adoc
matiasilva: True, I suppose that might even end up confusing some people as well.

i2c/oled_i2c/README.adoc
matiasilva: Much clearer!

i2c/oled_i2c/README.adoc
matiasilva: Hopefully this is better?

i2c/oled_i2c/README.adoc
matiasilva: Probably not

i2c/oled_i2c/oled_i2c.c
matiasilva: Agreed

i2c/oled_i2c/oled_i2c.c
matiasilva: Whoops, that's exactly it! I just tested this with one page, and that was page 0 xD.

i2c/oled_i2c/oled_i2c.c
matiasilva: Yes...The only reason this made its way here is that `start_col` and `start_page` are both equal to 0!

i2c/oled_i2c/oled_i2c.c
matiasilva: Good point. I've switched `i` to be equal to `area->start_page` (no checks to make sure start < end), but since this is an area and not the full 128x32 frame it makes sense to ignore the `start_col` and always start at the leftmost end. The motivation behind this function was so that I could print out the raspberry to the serial to debug it.

matiasilva: image

i2c/oled_i2c/oled_i2c.c
matiasilva: Switched to `buflen` to also make this consistent with the `render_area` struct.

i2c/oled_i2c/raspberry26x32.h
matiasilva: Would it be worth touching it up and including it in the folder?

matiasilva: > AFAICT it's an SVG image rendered out to a 26*32 monochrome bitmap, bit-swizzled to the page/column formatting suitable for the SSD1306 chip, and then encoded as a C-array of hex-bytes?

Yes, it's exactly that!

i2c/oled_i2c/README.adoc
```diff
@@ -1,36 +1,36 @@
...
+The SSD1306 is not operated via direct reads/writes to registers. Instead, a list of versatile commands is defined (see datasheet) that allows the user to access all the capabilities of the driver. Ultimately, these commands end up writing to internal registers but the key point is that the I2C write process is now different. After sending a slave address, the data that follows can be either a command, flags to follow up a command or data to be written directly into the display's RAM. A control byte is required for each write after the slave address so that the driver knows what type of data is being sent.
```
lurch, line 7: Perhaps the whole "Ultimately, these commands end up writing to internal registers but the key point is that the I2C write process is now different." sentence is superfluous? People probably only care about the commands they send to the SSD1306 chip, and are less interested about what the chip does behind the scenes? :man_shrugging:

i2c/oled_i2c/README.adoc
```diff
@@ -1,36 +1,36 @@
...
 
-This display is 32 pixels high by 128 pixels wide. These 32 pixels are partitioned into 4 pages, each 8 bits in height. In RAM, This looks roughly like:
+This display is 32 pixels high by 128 pixels wide. These 32 vertical pixels are partitioned into 4 pages, each 8 pixels in height. In RAM, This looks roughly like:
```
lurch, line 9: Maybe it's worth adding a note that the SSD1306 can also drive OLED screens with other pixel-dimensions?

i2c/oled_i2c/raspberry26x32.h
aallan: > Yes, it's exactly that!

In which case I think it's worth beautify'ing the Python script and including it along with the example, and add it in the 'list of files' so it gets pulled into the docs. I'm crossing my fingers it doesn't use any complicated tomfoolery and it's all just default installed Python modules so folks don't have to mess with their Python installation to run it (i.e. you have Python, the script runs)..?

i2c/oled_i2c/README.adoc
aallan: There's another OLED in the box of things, it's a https://thepihut.com/collections/adafruit-lcds-displays/products/adafruit-monochrome-1-3-128x64-oled-graphic-display-stemma-qt-qwiic, I think?

i2c/oled_i2c/README.adoc
```diff
@@ -1,36 +1,36 @@
...
+          | COL0 | COL1 | COL2 | COL3 |  ...  | COL126 | COL127 |
+    COM 0 |      |      |      |      |       |        |        |   
+    COM 1 |      |      |      |      |       |        |        |   
+       :  |      |      |      |      |       |        |        |  
+    COM 7 |      |      |      |      |       |        |        |
+    -------------------------------------------------------------
 ----
 
-Note the difference between columns in RAM and the actual segment pins on the display. The distinction between these two is important as we can remap, or add offsets between the two in the driver itself instead of moving pixels in our frame buffer code.
+Note the difference between columns in RAM and the actual segment pads that connect the driver to the display. The distinction between these two is important as we can remap, or add offsets between the two in the driver itself instead of moving pixels in our frame buffer code.
```
lurch, line 31: I'm still not sure what this paragraph is trying to say? :confused:   (maybe that's because I've not read the SSD1306 datasheet?)

matiasilva: The display's pins are physically connected to the driver via segment and common pins. The RAM addresses COL0 - COL127 are then mapped to these segment pins SEG0 - SEG127. But we can change the mapping, to easily mirror the image, etc without having to write a flipped buffer to the RAM again.

i2c/oled_i2c/README.adoc
matiasilva: Yeah, fair. This was probably me rationalizing the dense language I had read in the datasheet 😅

i2c/oled_i2c/raspberry26x32.h
matiasilva: Almost..It uses Pillow which is not in the standard library but is almost as universal as the likes of `matplotlib` and `numpy`, I think?

i2c/oled_i2c/oled_i2c.c
```diff
@@ -99,17 +98,17 @@ void print_buf_page(uint8_t buf[], uint8_t page) {
...
     }
 }
 
 void print_buf_area(uint8_t buf[], struct render_area* area) {
     // print a render area of generic size
-    uint8_t area_width = area->end_col - area->start_page + 1;
+    uint8_t area_width = area->end_col - area->start_col + 1;
     uint8_t area_height = area->end_page - area->start_page + 1; // in pages, not pixels
-    for (int i = 0; i < area_height; i++) {
+    for (int i = area->start_page; i < area_height; i++) {
```
lurch, line 111: Should this be `i <= area->end_page;` ?

Similarly, perhaps the
```c
for (int k = 0; k < area_width; k++) {
```
should be 
```c
for (int k = area->start_col; k <= area->end_col; k++) {
```
?
(although off the top of my head I'm not sure what the bit inside `buf[  ]` would need to become!)

i2c/oled_i2c/raspberry26x32.h
aallan: Yeah, Pillow is fairly standard. Brief pip install installation instructions in the comments at the top should be enough for that one.

i2c/oled_i2c/README.adoc
lurch: Perhaps this is another "behind the scenes" detail that the user doesn't need to know? WDYT @aallan ?

aallan: Sounds like good detail to have in there, but it's definitely an aside. Should be in a `[NOTE]` block.

aallan:
@matiasilva As with #127, can you change the breadboard used a mini one rather than a full sized. Should fit I think?

i2c/oled_i2c/README.adoc
```diff
@@ -4,9 +4,9 @@ This example code shows how to interface the Raspberry Pi Pico with an 128x32 OL
...
+The SSD1306 is operated via a list of versatile commands (see datasheet) that allows the user to access all the capabilities of the driver. After sending a slave address, the data that follows can be either a command, flags to follow up a command or data to be written directly into the display's RAM. A control byte is required for each write after the slave address so that the driver knows what type of data is being sent.
 
-This display is 32 pixels high by 128 pixels wide. These 32 vertical pixels are partitioned into 4 pages, each 8 pixels in height. In RAM, This looks roughly like:
+This display is 32 pixels high by 128 pixels wide and the SSD1306 can drive displays that are up to 64 pixels high and 128 pixels wide. These 32 vertical pixels are partitioned into 4 pages, each 8 pixels in height. In RAM, This looks roughly like:
```
lurch, line 9: Feels like the new-addition to this line would be less confusing if it was in a separate `NOTE` block?

"In RAM, This looks" - lowercase 't' on 'this' :slightly_smiling_face:

i2c/oled_i2c/oled_i2c.c
matiasilva: Hmmm, actually I think we should switch it back to 0. The reason is that there's a difference between printing just a little area (like the raspberry) and seeing what the entire frame (128x32) buffer has. If we print a little area, we just want to see the area and so place it in the left hand corner, so effectively start_col = 0 and start_page = 0. This also just makes things easier for the bit inside buf[] as we don't have to subtract any offsets, if, for example start_col is not 0. We really only the need the height and width of the area, but I wanted to make the "print" and "render" functions consistent and use the `render_area` struct.

i2c/oled_i2c/raspberry26x32.h
matiasilva: Done

i2c/oled_i2c/oled_i2c.c
lurch: IMHO if you're going to change the function to always print from the top-left corner, then you ought to make that explicit by passing in separate `width` and `height` parameters, rather than passing in a `render_area` struct and just ignoring its `start_page` and `start_col` values.
But it's probably worth waiting to see what @kilograham says, as he knows the "feel" of the SDK and Examples much better than I do :slightly_smiling_face:

i2c/oled_i2c/img_to_array.py
```diff
@@ -0,0 +1,79 @@
...
+img_path = sys.argv[1]
+
+try:
+    im = Image.open(img_path)
+except OSError:
+    print("Oops! The image could not be opened.")
+    sys.exit()
+
+if im.size[0] > OLED_WIDTH or im.size[1] > OLED_HEIGHT:
+    print("OLED display only 128 pixels wide and 32 pixels high!")
```
lurch, line 33:
```python
   print("OLED display only {} pixels wide and {} pixels high!".format(OLED_WIDTH, OLED_HEIGHT))
```
:wink:

i2c/oled_i2c/img_to_array.py
```diff
@@ -0,0 +1,79 @@
...
+    sys.exit()
+
+if not (im.mode == "1" or im.mode == "L"):
+    print("Images must be grayscale only")
+    sys.exit()
```
lurch, line 38: Instead of doing:
```python
print(message)
sys.exit()
```
you could just do
```python
raise Exception(message)
```
?

i2c/oled_i2c/img_to_array.py
```diff
@@ -0,0 +1,79 @@
...
+    for j in range(IMG_WIDTH):
+        out_byte = 0
+        for k in range(8):
+            out_byte |= pixels[k*IMG_WIDTH + start_index + j] << k
+        buffer.append(f'{out_byte:#04x}')
+
+buffer = ", ".join(buffer)
+buffer_hex = f"static uint8_t[] = {{{buffer}}}"
+
+print(buffer_hex)
```
lurch, line 79: Might make more sense to have this write out a file (based on the name of the input image), rather than just printing the hex array to screen?

i2c/oled_i2c/img_to_array.py
```diff
@@ -0,0 +1,79 @@
...
+    sys.exit()
+
+if not (im.mode == "1" or im.mode == "L"):
+    print("Images must be grayscale only")
+    sys.exit()
+
+# black or white
+out = im.convert("1")
+
+IMG_WIDTH = out.size[0]
```
lurch, line 43: I guess there's no reason for these variables to be CAPITALISED?

i2c/oled_i2c/img_to_array.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+
+# Converts a greyscale image into a format able to be
+# displayed by the SSD1306 driver in horizontal addressing mode
+
+# usage: ./img_to_array.py <logo.bmp> on a Raspberry Pi
+# or python3 img_to_array.py <logo.bmp>
```
lurch, line 7: "on a Raspberry Pi" is "on Linux, MacOS, etc." :wink: (i.e. any unix-like OS that reads the `#!` line)

Probably easiest to just say:
"# usage: python3 img_to_array.py <logo.bmp>"
as that covers all bases.

i2c/oled_i2c/img_to_array.py
```diff
@@ -0,0 +1,79 @@
...
+# depends on the Pillow library
+# `python3 -m pip install --upgrade Pillow`
+
+from PIL import Image
+import sys
+
+OLED_HEIGHT = 32
+OLED_WIDTH = 128
+OLED_PAGE_HEIGHT = 8
+OLED_NUM_PAGES = OLED_HEIGHT / OLED_PAGE_HEIGHT
```
lurch, line 18: This isn't used anywhere?

matiasilva: We use `OLED_NUM_PAGES` and I didn't want that to be a magic number so I added the rest of the variables in. Could also be useful if someone decides to use the 128x64 version?

i2c/oled_i2c/img_to_array.py
matiasilva: Ah, yes. I wanted the convenience of just being able to `./` it but I suppose anyone clued up on that will see the shebang and just do it.

i2c/oled_i2c/img_to_array.py
```diff
@@ -26,22 +26,20 @@
...
     im = Image.open(img_path)
 except OSError:
-    print("Oops! The image could not be opened.")
-    sys.exit()
+    raise Exception("Oops! The image could not be opened.")
 
 if im.size[0] > OLED_WIDTH or im.size[1] > OLED_HEIGHT:
-    print("OLED display only 128 pixels wide and 32 pixels high!")
-    sys.exit()
+    raise Exception(f"OLED display only {OLED_WIDTH} pixels wide and {OLED_HEIGHT} pixels high!")
```
lurch, line 32: Perhaps this should also print out the dimensions of the image, to make it clearer what the problem actually is? :wink:

i2c/oled_i2c/img_to_array.py
```diff
@@ -26,22 +26,20 @@
...
 
 if not (im.mode == "1" or im.mode == "L"):
-    print("Images must be grayscale only")
-    sys.exit()
+    raise Exception("Images must be grayscale only")
```
lurch, line 35: Images -> Image

i2c/oled_i2c/img_to_array.py
lurch: I agree that the rest of them ought to be kept, but it seems like `OLED_NUM_PAGES` isn't actually used in the Python script? (which is actually what I meant with my previous comment, apologies for the ambiguity)

i2c/oled_i2c/img_to_array.py
```diff
@@ -0,0 +1,80 @@
...
+        for k in range(OLED_PAGE_HEIGHT):
+            out_byte |= pixels[k*img_width + start_index + j] << k
+        buffer.append(f'{out_byte:#04x}')
+
+buffer = ", ".join(buffer)
+buffer_hex = f'static uint8_t {img_name}[] = {{{buffer}}}\n'
+
+with open(f'{img_name}.h', 'wt') as file:
+    file.write(f'#define IMG_WIDTH {img_width}\n')
+    file.write(f'#define IMG_HEIGHT {img_height}\n')
```
lurch, line 79: Probably want an extra `\n` here so that it matches your `raspberry26x32.h` ?

i2c/oled_i2c/img_to_array.py
matiasilva: Oh true!

i2c/oled_i2c/img_to_array.py
```diff
@@ -53,27 +51,30 @@
...
         out_byte = 0
-        for k in range(8):
-            out_byte |= pixels[k*IMG_WIDTH + start_index + j] << k
+        for k in range(OLED_PAGE_HEIGHT):
+            out_byte |= pixels[k*img_width + start_index + j] << k
         buffer.append(f'{out_byte:#04x}')
 
 buffer = ", ".join(buffer)
-buffer_hex = f"static uint8_t[] = {{{buffer}}}"
+buffer_hex = f'static uint8_t {img_name}[] = {{{buffer}}}\n'
```
lurch, line 76: If you were feeling fancy, you could insert a regex-check to check that there were no characters in `img_name` that aren't valid in a C variable-name :slightly_smiling_face:  (spaces, hyphens, etc.)
For extra fanciness you could automatically convert invalid chars to underscores, so that "my awesome-image!.bmp" creates `my awesome-image!.h` containing `static uint8_t my_awesome_image_[]`.

An alternative approach would be to always create the array with the same name, e.g. `static uint8_t image_data[]`.

matiasilva: Yeah, lol, I was hoping to outsource as much of that as possible to `pathlib` but I'm happy to give it a go? The only reason I've used the filename is that other examples that use images in header files all have matching names (like the pio LCD one)

i2c/oled_i2c/img_to_array.py
```diff
@@ -28,17 +27,19 @@
 except OSError:
     raise Exception("Oops! The image could not be opened.")
 
-if im.size[0] > OLED_WIDTH or im.size[1] > OLED_HEIGHT:
+img_width = im.size[0]
+img_height = im.size[1]
+
+if img_width > OLED_WIDTH or img_height > OLED_HEIGHT:
+    print(f'Your image is f{img_width} pixels wide and {img_height} pixels high, but...')
     raise Exception(f"OLED display only {OLED_WIDTH} pixels wide and {OLED_HEIGHT} pixels high!")
```
lurch: I think it'd be better to put a `\n` in the middle of the long Exception message, rather than splitting the message between `print` and `Exception` :man_shrugging:

i2c/oled_i2c/img_to_array.py
lurch: I dunno how familiar you are with [regexes](https://docs.python.org/3/library/re.html), but `array_name = re.sub(r'[^\w]', '_', img_name)` is probably what you want? (Replace anything NOT in `a-zA-Z0-9_` with an underscore)

aallan:

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,295 @@
...
+    // prints all pages of a full length buffer
+    for (int i = 0; i < OLED_NUM_PAGES; i++) {
+        printf("--page %d--\n", i);
+        print_buf_page(buf, i);
+    }
+}
+
+void print_buf_area(uint8_t buf[], struct render_area* area) {
+    // print a render area of generic size
+    uint8_t area_width = area->end_col - area->start_col + 1;
```
kilograham, line 109: again stylistic... would not make these small types (they aren't being stored)

i2c/oled_i2c/oled_i2c.c
```diff
@@ -0,0 +1,295 @@
...
+    // in horizontal addressing mode, the column address pointer auto-increments
+    // and then wraps around to the next page, so we can send the entire frame
+    // buffer in one gooooooo!
+
+    // copy our frame buffer into a new buffer because we need to add the control byte
+    // to the beginning
+
+    // TODO find a more memory-efficient way to do this..
+    // maybe break the data transfer into pages?
+    uint8_t temp_buf[buflen + 1];
```
kilograham, line 147: ugh. stack based array of indeterminate size... presumably this is limited to 4K because of 128x32 here...

JamesH65:
@matiasilva Hi Matias, can you click the right button to give me push access to your branches so I can make the requried changes and push them back to your PR?

PR #132, opened 2021-07-02
Improve and cleanup cmake config file comments

matiasilva:
Two improvements from the examples written so far:

* remove the "to be renamed" line 
* Add "GP" to clarify the pins in question are GPIO numbered

### DIFF
--- a/clocks/hello_48MHz/CMakeLists.txt
+++ b/clocks/hello_48MHz/CMakeLists.txt
@@ -2,7 +2,7 @@ add_executable(hello_48MHz
         hello_48MHz.c
         )
 
-# Pull in our pico_stdlib which pulls in commonly used features
+# pull in common dependencies
 target_link_libraries(hello_48MHz pico_stdlib hardware_clocks)
 
 # create map/bin/hex file etc.
--- a/clocks/hello_48MHz/CMakeLists.txt
+++ b/clocks/hello_48MHz/CMakeLists.txt
@@ -2,7 +2,7 @@ add_executable(hello_48MHz
         hello_48MHz.c
         )
 
-# pull in common dependencies
+# pull in common dependencies and additional clocks hardware support
 target_link_libraries(hello_48MHz pico_stdlib hardware_clocks)
 
 # create map/bin/hex file etc.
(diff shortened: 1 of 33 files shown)


lurch:
LGTM :+1:  But this is one for @kilograham to decide.

kilograham:
yeah, i think if we're going to fix up the comments, lets go the whole way

"pico_stdlib" is for the "common" dependencies you could call them (maybe even "standard")

In the cases we're including other libraries, we should probably say "and additional xxxx hardware support"

matiasilva:
Will add that soon to this PR

matiasilva:
Done!

clocks/hello_48MHz/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ add_executable(hello_48MHz
         hello_48MHz.c
         )
 
-# Pull in our pico_stdlib which pulls in commonly used features
+# pull in common dependencies
```
lurch, line 5: ...and additional clocks hardware support?

uart/hello_uart/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ add_executable(hello_uart
         hello_uart.c
         )
 
-# Pull in our pico_stdlib which pulls in commonly used features
+# pull in common dependencies and additional uart hardware support
```
lurch, line 5: `hardware_uart` doesn't appear on the line below?

clocks/hello_48MHz/CMakeLists.txt
matiasilva: Whoops!

kilograham (approved):

interp/hello_interp/CMakeLists.txt
```diff
@@ -3,12 +3,12 @@ if (TARGET hardware_interp)
             hello_interp.c
             )
 
-    # Pull in dependencies
+    # pull in common dependencies and additional interpolation hardware support
```
lurch, line 6: I dunno if this ought to be "interpolator" rather than "interpolation" ?  (minor nit-pick so feel free to ignore)

matiasilva: I agree!

lurch (approved):

kilograham (approved):
lgtm

PR #133, opened 2021-07-02
Add monochrome e-ink display (SSD1681) example

matiasilva:
This PR adds an example for the 200x200 monochrome e-ink display based on the SSD1681. Demonstrates use of the onboard SRAM and the SD card (possibly).

## Checklist

- []  documentation
- [] working example
- [] Fritzing schematic and image
- [] added to CMakeLists.txt
- [] added to repo README


spi/CMakeLists.txt
```diff
@@ -3,4 +3,5 @@ if (NOT PICO_NO_HARDWARE)
     add_subdirectory(mpu9250_spi)
     add_subdirectory(bme280_spi)
     add_subdirectory(spi_dma)
+    add_subdirectory(eink_mono_spi)
```
lurch, line 6: Alphabetical sorting :wink:

spi/eink_mono_spi/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(eink_mono_spi
+        eink_mono_spi.c
+        )
+
+# Pull in our (to be renamed) simple get you started dependencies
```
lurch, line 5: Uh-oh :laughing:

lurch:
On the wiring diagram, it's a shame that the wires overlap the labels on the display-board :slightly_frowning_face: Maybe the best way of fixing that is to connect the jumper-wires to the breadboard instead of directly to the display-board? But well done for at least using a different colour for each wire :+1:

aallan:
> On the wiring diagram, it's a shame that the wires overlap the labels on the display-board 🙁 Maybe the best way of fixing that is to connect the jumper-wires to the breadboard instead of directly to the display-board? But well done for at least using a different colour for each wire 👍

Agreed. You should probably move the screen down to the other side of the breadboard divider, and route the wires into the relevant rows below the screen, rather than directly to the screen. The wires *definitely* should not be connected directly to the Pico, but into the rows either side.

Also, I'm not a huge fan of using loopy wires in these diagrams I think they're easier to understand with 90-degree corners. So I tend to do diagrams as below…

image

image

image

spi/CMakeLists.txt
matiasilva: Left this one at the end until the other PR gets merged to prevent conflicts :D

spi/eink_mono_spi/CMakeLists.txt
matiasilva: OMG. I think something similar happened and that's when I decided to submit the other two PRs.

matiasilva:
I agree as well. I wasn't feeling 100% about the curvy wires either, but decided to try something new. Thanks for the feedback early on!

lurch:
Your Fritzing diagram looks better now, but IMHO you ought to tweak it so that the yellow wire doesn't "disappear" behind the green wire, and the red and black wires should just connect to the breadboard, rather than connecting directly to the Pico and display PCBs.

matiasilva:
> Your Fritzing diagram looks better now, but IMHO you ought to tweak it so that the yellow wire doesn't "disappear" behind the green wire, and the red and black wires should just connect to the breadboard, rather than connecting directly to the Pico and display PCBs.

Ah the red/black wires were an oversight. How does this look for the yellow and green?

image

lurch:
I realise that's probably the "easiest" way to fix the current diagram, but I think it looks a bit odd with the yellow wire running over the edge of the blue PCB like that? But lets see what @aallan says...

aallan:
> I realise that's probably the "easiest" way to fix the current diagram, but I think it looks a bit odd with the yellow wire running over the edge of the blue PCB like that? But lets see what @aallan says...

Looking at this,

image

I'd route all of the wires from the Pico straight down off the breadboard and run them parallel to the edge outside of the board. The cross overs can then happen on white space (see the diagrams I did above for examples of where I did exactly this) instead of on the confusing background of the breadboard.

If you can't get the yellow and green writes to sit right I might be tempted to pull the display off the breadboard and have it in white space like I did in the example above with the RFM95x breakout?

matiasilva:
Done, check now?

aallan:
> Done, check now?

Much better. I'd probably move the parallel lines down a bit so the white and blue lines are off the board as well.

image

lurch:
> pull the display off the breadboard and have it in white space like I did in the example above with the RFM95x breakout?

The "problem" with that @aallan is that you've then covered up the PCB-labels with wires (which is what my original comment was about!) which would make it harder to use the wiring diagram with a breakout-board from another manufacturer, which might have the pins in a different order? :man_shrugging:  However I've not played with Fritzing enough myself to be able to suggest a better alternative.

> Done, check now?

Yup, that looks much nicer @matiasilva :+1:

matiasilva:
Between this diagram and the explicit wiring instructions, hopefully the person tinkering with the example should be able to figure it out.

lurch:
> hopefully the person tinkering with the example should be able to figure it out.

"WARNING: It is _very_ important that you use a purple wire for the SCK connection. Using any other colour of wire may cause the example to fail." :rofl: :wink: :rofl:

matiasilva:
After much head scratching, I'm not really sure why this isn't working. All I've tried to do is just zero the display (all black) but nothing happens. After I send all the buffer data, and tell the display to activate, it just turns on the BUSY pin indefinitely and hangs.

README.md
```diff
@@ -148,6 +148,7 @@ App|Description
 App|Description
 ---|---
 [bme280_spi](spi/bme280_spi) | Attach a BME280 temperature/humidity/pressure sensor via SPI.
+[eink_mono_spi](spi/eink_mono_spi) | Display text and images on a 200x200 monochrome e-ink display driven by the SSD1618 with onboard SRAM.
```
lurch, line 151: Is "onboard SRAM" relevant at this high-level overview? :shrug:

spi/bme280_spi/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ add_executable(bme280_spi
         bme280_spi.c
         )
 
-# Pull in our (to be renamed) simple get you started dependencies
+# Pull in our get you started dependencies
```
lurch, line 5: I think the change to this file shouldn't be in this PR?

spi/eink_mono_spi/eink_mono_spi.c
```diff
@@ -0,0 +1,296 @@
...
+    gpio_init(BUSY_PIN);
+    gpio_init(ENA_PIN);
+    gpio_set_dir(DC_PIN, GPIO_OUT);
+    gpio_set_dir(RST_PIN, GPIO_OUT);
+    gpio_set_dir(BUSY_PIN, GPIO_IN);
+    gpio_set_dir(ENA_PIN, GPIO_OUT);
+    gpio_put(DC_PIN, 0);
+    // Make the CS pin available to picotool
+    bi_decl(bi_1pin_with_name(RST_PIN, "Display D/C"));
+    bi_decl(bi_1pin_with_name(RST_PIN, "Display RST"));
```
lurch, line 284: `RST_PIN` used twice?

spi/eink_mono_spi/eink_mono_spi.c
```diff
@@ -0,0 +1,296 @@
...
+    bi_decl(bi_1pin_with_name(BUSY_PIN, "Display ENA"));
+    bi_decl(bi_1pin_with_name(ENA_PIN, "Display BUSY"));
```
lurch, line 286: These appear to be the wrong way around?

spi/eink_mono_spi/eink_mono_spi.c
```diff
@@ -0,0 +1,296 @@
...
+
+    // initialize DC, RST, BUSY, ENA pins
+    gpio_init(DC_PIN);
+    gpio_init(RST_PIN);
+    gpio_init(BUSY_PIN);
+    gpio_init(ENA_PIN);
+    gpio_set_dir(DC_PIN, GPIO_OUT);
+    gpio_set_dir(RST_PIN, GPIO_OUT);
+    gpio_set_dir(BUSY_PIN, GPIO_IN);
+    gpio_set_dir(ENA_PIN, GPIO_OUT);
```
lurch, line 280: Do you need to drive this pin high? :man_shrugging:

aallan:
Where are we on this @matiasilva, your initial checklist is still missing "working example"?

matiasilva:
> Where are we on this @matiasilva, your initial checklist is still missing "working example"?

I'm going to revisit this now!

JamesH65:
@matiasilva How far did you get with this one?

PR #134, opened 2021-07-05
Consider filter drop for Vref and use max ADC as 2^12 - 1

polarbearing:
Pico datasheet, 4.3. Using the ADC: "The ADC draws current (about 150μA if the temperature sense diode is disabled, but it varies from chip to chip) and therefore there will be an inherent offset of about 150μA*200 = ~30mV. There is a small difference in current draw when the ADC is sampling (about +20μA)"
Thus when sampling Vref is nominal (150+20)μA*200 = ~34mV

Splitting the variable in two lines makes the change for the clamped situation more intuitive. 

The value range of a 12 bit resolution is 0-4095, i.e. 2^12 minus one.

### DIFF
--- a/adc/hello_adc/hello_adc.c
+++ b/adc/hello_adc/hello_adc.c
@@ -21,8 +21,10 @@ int main() {
     adc_select_input(0);
 
     while (1) {
-        // 12-bit conversion, assume max value == ADC_VREF == 3.3 V
-        const float conversion_factor = 3.3f / (1 << 12);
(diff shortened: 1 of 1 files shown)


matiasilva:
I suspect the best branch for this to be merged into should probably be `develop`?

aallan:
> I suspect the best branch for this to be merged into should probably be `develop`?

Yes. It'll need to be rebased onto `develop`. We don't accept PRs to `master`.

PR #135, opened 2021-07-09
Add example for analog ICS-40180 microphone

matiasilva:
This PR adds an example for an analog ICS-40180 based microphone from SparkFun (but really any analog microphone should do). It includes a simple Python script that takes information directly from the serial and plots it for easy visualization.

## Checklist

- [x]  documentation
- [ ] working example
- [x] Fritzing schematic and image
- [x] added to CMakeLists.txt
- [x] added to repo README


adc/microphone_adc/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(microphone_adc
+        microphone_adc.c
+        )
+
+# Pull in our get you started dependencies
```
lurch, line 5: I guess this should be "the new format" ?

adc/microphone_adc/README.adoc
```diff
@@ -0,0 +1,49 @@
+= Attaching a microphone using the ADC
+
+This example code shows how to interface the Raspberry Pi Pico with a standard analog microphone via the onboard analog to digital converter (ADC). In this example, we use an ICS-40180 breakout board by SparkFun but any analog microphone should be compatible with this tutorial. SparkFun have https://learn.sparkfun.com/tutorials/mems-microphone-hookup-guide[written a guide] for this board that goes into more detail about the board and how it works.
+
+[TIP]
+======
+An analog to digital converter (ADC) is responsible for reading continuous input signals that may range from 0 to a specified reference voltage (in the Pico's case this reference voltage is set by the supply voltage and can be measured on pin 35, ADC_VREF) and converting them into 1s and 0s, ie. a number that can be digitally stored.
```
lurch, line 7: Perhaps "binary" might be better than "1s and 0s", as the latter might imply the ADC only returns 1 or 0, not a 12-bit binary value? :shrug:

adc/microphone_adc/README.adoc
```diff
@@ -0,0 +1,49 @@
...
+======
+
+The Pico has a 12-bit ADC, meaning that a read operation will return a number ranging from 0 to 4095 (2^12 - 1) for a total of 4096 possible values. Therefore, the resolution of the ADC is 3.3/4096, so roughly steps of 0.8 millivolts. The SparkFun breakout uses an OPA344 operational amplifier to boost the signal coming from the microphone to voltage levels that can be easily read by the ADC. An important side effect is that a bias of 0.5*Vcc is added to the signal, even when the microphone is not picking up any sound.
```
lurch, line 10: Is `Vcc` the same thing as `ADC_VREF`?

@aallan Do we need to mention ENOB here, or is it fine to pretend that the ADC is perfect?

adc/microphone_adc/README.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+The ADC provides us with a raw voltage value but when dealing with sound, we're more interested in the amplitude of the audio signal. This is defined as one half the peak-to-peak amplitude. Included with this example is a very simple Python script that will plot the values it receives via the serial port. By tweaking the sampling rates, and various other parameters, the data from the microphone can analyzed in various ways, such as in a Fast Fourier Transform to see what frequencies make up the signal.
```
lurch, line 12: "can analyzed" -> "can be analysed"

adc/microphone_adc/README.adoc
```diff
@@ -0,0 +1,49 @@
...
+[[microphone_adc_plotter_image]]
+[pdfwidth=75%]
+.Example output from included Python script
+image::microphone_adc_plotter.png[]
+
+== Wiring information
+
+Wiring up the device requires 3 jumpers, to connect VCC (3.3v), GND, and AOUT. The example here uses ADC0, which is GP26. Power is supplied from the 3.3V pin.
+
+WARNING: Most boards will take a range of voltages from the Pico's default 3.3v to the 5 volts commonly seen on other microcontrollers. Ensure your board works with either the 3.3v or 5v supplied by the Pico.
```
lurch, line 23: Does this also need to warn about not connecting a voltage greater than 3.3V to Pico's ADC?
(also, I think we prefer `5V` to `5v`)

adc/microphone_adc/microphone_adc.c
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#include <stdio.h>
+#include "pico/stdlib.h"
+#include "hardware/gpio.h"
+#include "hardware/adc.h"
+#include "hardware/uart.h"
+#include "pico/binary_info.h"
+
+ /* Example code to plot analog values from a microphone
```
lurch, line 14: This `C` file doesn't do any plotting, so perhaps this comment needs tweaking?

I guess it depends whether the comment here is referring to the whole project, or just this individual file within the project? :shrug: ping @aallan

adc/microphone_adc/microphone_adc.c
```diff
@@ -0,0 +1,47 @@
...
+    
+    adc_init();
+    adc_gpio_init(ADC_PIN + 26);
+    adc_select_input(ADC_PIN);
+
+    uint adc_raw;
+    while (1) {
+        adc_raw = adc_read(); // raw voltage from ADC
+        printf("%.2f\n", adc_raw * ADC_CONVERT);
+        sleep_ms(10);
```
lurch, line 43: I guess the 10ms sleep here implies there's a maximum frequency that this code will be able to detect? (only mentioning this because you talked about FFT analysis earlier)

adc/microphone_adc/plotter.py
```diff
@@ -0,0 +1,77 @@
...
+        t = lastt + 1
+        self.tdata.append(t)
+        self.ydata.append(y)
+        self.line.set_data(self.tdata, self.ydata)
+        return self.line,
+
+
+def serial_getter():
+    # grab fresh ADC values
+    # note sometimes UART drops chars so we try a max of 5 times
```
lurch, line 50: Ahhh, this is probably related to https://github.com/raspberrypi/pico-sdk/pull/504 !

adc/microphone_adc/README.adoc
lurch, line 12: Why say "we're more interested in the amplitude of the audio signal. This is defined as one half the peak-to-peak amplitude." when the Python code just plots the raw voltage?

adc/microphone_adc/README.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+
+[[ics-40180-adc_wiring]]
+[pdfwidth=75%]
+.Wiring Diagram for ICS-40180 microphone breakout board.
+image::ics_40180_adc_bb.png[]
```
lurch, line 29: Image name doesn't match up with the files in this PR?

adc/microphone_adc/README.adoc
aallan: > @aallan Do we need to mention ENOB here, or is it fine to pretend that the ADC is perfect?

Urgh. We should probably say "The Pico has a 12-bit ADC (ENOB of 8.7-bit, see the [RP2040 Datasheet]() for details), meaning that…" with a URL that links into that specific section.

adc/microphone_adc/README.adoc
matiasilva: I don't think so, as the output from this board is 3.3V anyway, so not relevant to this example.

adc/microphone_adc/microphone_adc.c
matiasilva: Yes. Though I'm not 100% sure where the botteneck is/will be, I don't think it's with the ADC but rather on the Python side with polling values from the UART, so I just set 10ms as a sensible value.

adc/microphone_adc/README.adoc
matiasilva: I think it's interesting for the user to be able to extend the code to do that, or even just be aware that in audio, that is what matters most.

adc/microphone_adc/microphone_adc.c
matiasilva: I've tweaked it now to explain what does the plotting and what does the ADC value reading.

adc/microphone_adc/README.adoc
lurch: I think it's the "either the 3.3v or 5v supplied by the Pico" that confused me. (which perhaps implies that you can power the microphone with Pico's 5V, and then feed 5V into Pico's ADC?). Or I might just be being overly cautious :shrug: :laughing:

matiasilva: Ohhh, I see how that could be misleading. I'll clarify things a bit.

adc/microphone_adc/README.adoc
```diff
@@ -4,12 +4,12 @@ This example code shows how to interface the Raspberry Pi Pico with a standard a
 
 [TIP]
 ======
-An analog to digital converter (ADC) is responsible for reading continuous input signals that may range from 0 to a specified reference voltage (in the Pico's case this reference voltage is set by the supply voltage and can be measured on pin 35, ADC_VREF) and converting them into 1s and 0s, ie. a number that can be digitally stored.
+An analog to digital converter (ADC) is responsible for reading continuous input signals that may range from 0 to a specified reference voltage (in the Pico's case this reference voltage is set by the supply voltage and can be measured on pin 35, ADC_VREF) and converting them into binary, ie. a number that can be digitally stored.
```
lurch, line 7: I wonder if "continually varying" might be a good substitute for "continuous" here?

@aallan should the "ie." be "i.e." ?

adc/microphone_adc/README.adoc
```diff
@@ -4,12 +4,12 @@ This example code shows how to interface the Raspberry Pi Pico with a standard a
...
 
-The ADC provides us with a raw voltage value but when dealing with sound, we're more interested in the amplitude of the audio signal. This is defined as one half the peak-to-peak amplitude. Included with this example is a very simple Python script that will plot the values it receives via the serial port. By tweaking the sampling rates, and various other parameters, the data from the microphone can analyzed in various ways, such as in a Fast Fourier Transform to see what frequencies make up the signal.
+The ADC provides us with a raw voltage value but when dealing with sound, we're more interested in the amplitude of the audio signal. This is defined as one half the peak-to-peak amplitude. Included with this example is a very simple Python script that will plot the values it receives via the serial port. By tweaking the sampling rates, and various other parameters, the data from the microphone can be analysed in various ways, such as in a Fast Fourier Transform to see what frequencies make up the signal.
```
lurch, line 12: Perhaps changing "will plot the values it receives" to "will plot the voltage values it receives" will clear up my earlier voltage / amplitude confusion?

adc/microphone_adc/README.adoc
```diff
@@ -20,7 +20,7 @@ image::microphone_adc_plotter.png[]
 
 Wiring up the device requires 3 jumpers, to connect VCC (3.3v), GND, and AOUT. The example here uses ADC0, which is GP26. Power is supplied from the 3.3V pin.
 
-WARNING: Most boards will take a range of voltages from the Pico's default 3.3v to the 5 volts commonly seen on other microcontrollers. Ensure your board works with either the 3.3V or 5V supplied by the Pico.
+WARNING: Most boards will take a range of voltages from the Pico's default 3.3v to the 5 volts commonly seen on other microcontrollers. Ensure your board works with either the 3.3V or 5V supplied by the Pico. If you supply 5V, then the board must have a level shifter to reduce the output voltage to 3.3V to be compatible with the Pico's GPIOs, otherwise you risk damaging the Pico!
```
lurch, line 23: I'm not sure this added sentence is actually that helpful as IIRC level shifters are used for translating digital signals, not constantly-varying analogue signals like we're talking about with this ADC example?

Perhaps it makes sense to just say something like:
"WARNING: Most boards will take a range of VCC voltages from the Pico's default 3.3V to the 5 volts commonly seen on other microcontrollers. Ensure your board doesn't output an analogue signal greater than 3.3V as this may result in permanent damage to the Pico's ADC."

WDYT @aallan ?

adc/microphone_adc/README.adoc
aallan: > @aallan should the "ie." be "i.e." ?

Yes.

adc/microphone_adc/microphone_adc.c
```diff
@@ -0,0 +1,48 @@
...
+
+    GPIO 26/ADC0 (pin 31)-> AOUT or AUD on microphone board
+    3.3v (pin 36) -> VCC on microphone board
+    GND (pin 38)  -> GND on microphone board
+ */
+
+#define ADC_PIN 0
+#define ADC_VREF 3.3
+#define ADC_RANGE (1 << 12)
+#define ADC_CONVERT ADC_VREF / ADC_RANGE
```
lurch, line 27: Should this be `#define ADC_CONVERT ADC_VREF / (ADC_RANGE - 1)` ?

adc/microphone_adc/README.adoc
aallan: > level shifters are used for translating digital signals, not constantly-varying analogue signals like we're talking about with this ADC example

You are correct, a level shifter won't help you here.

> Perhaps it makes sense to just say something like:
"WARNING: Most boards will take a range of VCC voltages from the Pico's default 3.3V to the 5 volts commonly seen on other microcontrollers. Ensure your board doesn't output an analogue signal greater than 3.3V as this may result in permanent damage to the Pico's ADC."

That sounds okay to me!

adc/microphone_adc/microphone_adc.c
```diff
@@ -0,0 +1,48 @@
...
+
+int main() {
+    stdio_init_all();
+    printf("Beep boop, listening...\n");
+
+    bi_decl(bi_program_description("Analog microphone example for Raspberry Pi Pico")); // for picotool
+    bi_decl(bi_1pin_with_name(ADC_PIN, "ADC input pin"));
```
lurch, line 34: @kilograham Is `ADC_PIN` here correct, or should it be `ADC_PIN + 26` as used with `adc_gpio_init` below?

adc/microphone_adc/README.adoc
matiasilva: Oh...right 🤦

adc/microphone_adc/microphone_adc.c
matiasilva: I think that's what I had originally, but there are 4096 different voltage values the output could take, right?

adc/microphone_adc/microphone_adc.c
matiasilva: Amusingly, I faced the same dilemma. `adc_select_input` expects 0-3 while `adc_gpio_init` expects a proper pin number.

adc/microphone_adc/README.adoc
```diff
@@ -20,7 +20,7 @@ image::microphone_adc_plotter.png[]
 
 Wiring up the device requires 3 jumpers, to connect VCC (3.3v), GND, and AOUT. The example here uses ADC0, which is GP26. Power is supplied from the 3.3V pin.
 
-WARNING: Most boards will take a range of voltages from the Pico's default 3.3v to the 5 volts commonly seen on other microcontrollers. Ensure your board works with either the 3.3V or 5V supplied by the Pico. If you supply 5V, then the board must have a level shifter to reduce the output voltage to 3.3V to be compatible with the Pico's GPIOs, otherwise you risk damaging the Pico!
+WARNING: Most boards will take a range of VCC voltages from the Pico's default 3.3V to the 5 volts commonly seen on other microcontrollers. Ensure your board doesn't output an analogue signal greater than 3.3V as this may result in permanent damage to the Pico's ADC.
 
 WARNING: Do not connect a voltage greater than 3.3V to the Pico's ADC (and any of its GPIO) as this may result in permanent damage.
```
lurch, line 25: This can be removed, now it's included in the previous paragraph :wink:

adc/microphone_adc/microphone_adc.c
lurch: Yup, but one of those 4096 values is zero :wink:  I have to confess, I always find this confusing too!

So, the ADC can read values between 0V (which will be a 12-bit 0) and 3.3V (which will be a 12-bit 4095).
Which means we want a reading of 4095 from the ADC to come back out again as 3.3V.

With `#define ADC_CONVERT ADC_VREF / ADC_RANGE`, an ADC value of 4095 would result in a human-readable voltage of 3.299V.
With `#define ADC_CONVERT ADC_VREF / (ADC_RANGE - 1)`, an ADC value of 4095 would result in a human-readable voltage of 3.3V.

(but of course for all practical purposes the difference here will actually be indistinguishable :laughing: )

matiasilva: Ah!

aallan:

JamesH65 (approved):
LGTM

PR #137, opened 2021-07-12
Enhancement - added servo motor example

SEEK-THE-VOID-LIGHT:
This is example is an adapted version from the micropython repository to be used directly in C.


lurch:
Looks like this PR contains commits that you don't want it to. Easiest solution is to PR from a separate branch instead of using `SEEK-THE-VOID-LIGHT:master`. Also, this project only accepts PRs to the `develop` branch (not the master branch).

PR #139, opened 2021-07-12
added servo example

SEEK-THE-VOID-LIGHT:
**Servo motor example**

This example is an adapted version of the Micropython repository to be used easily in C.


SEEK-THE-VOID-LIGHT:
Closed because of selecting wrong branch

PR #140, opened 2021-07-12
added servo example

SEEK-THE-VOID-LIGHT:
**Servo motor example**

This example is an adapted version of the Micropython repository to be used easily in C.

### DIFF
--- a/pwm/servo/main.c
+++ b/pwm/servo/main.c
@@ -1,0 +1,15 @@
+#include "pico/bootrom.h"
+#include "pico/stdlib.h"
+#include "servo.h"
+#include <stdbool.h>
+
+#define ANGLE 90
+
+
+int main()
+{
+    servo_init(1, 50);
+    servo_put(1, ANGLE, true);
+
+    reset_usb_boot(0,0); // reset to BOOTSEL for another flash
+}
--- a/pwm/servo/servo.c
+++ b/pwm/servo/servo.c
@@ -1,0 +1,56 @@
+#include "servo.h"
+#include "pico/stdlib.h"
+#include "hardware/pwm.h"
+#include "hardware/clocks.h"
+#include <stdbool.h>
+
+#define DUTY_MIN 2400
+#define DUTY_MAX 8000
+
+long map(long x, long in_min, long in_max, long out_min, long out_max) {
+  return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
+}
+
+void servo_put(int gpio, int angle, bool wait_write)
+{
+    uint slice = pwm_gpio_to_slice_num(gpio);
+    uint channel = pwm_gpio_to_channel(gpio);
+    uint32_t top = pwm_hw->slice[slice].top;
+    int duty_u16 = map(angle, 0, 180, DUTY_MIN, DUTY_MAX);
+    uint32_t cc = duty_u16 * (top + 1) / 65535;
+    pwm_set_chan_level(slice, channel, cc);
+    pwm_set_enabled(slice, true);
+
+    if(wait_write)
+        sleep_ms(1000);
+}
+
+void servo_init(int gpio, int base_frequency)
+{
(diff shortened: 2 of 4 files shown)


pwm/servo/main.c
```diff
@@ -0,0 +1,15 @@
...
+
+#define ANGLE 90
+
+
+int main()
+{
+    servo_init(1, 50);
+    servo_put(1, ANGLE, true);
+
+    reset_usb_boot(0,0); // reset to BOOTSEL for another flash
```
lurch, line 14: I don't think we want to have any examples that deliberately reboot the Pico? Having an example that continuously sweeps the servo back and forth would probably be more useful?

pwm/servo/servo.c
```diff
@@ -0,0 +1,56 @@
...
+
+void servo_init(int gpio, int base_frequency)
+{
+    gpio_set_function(gpio, GPIO_FUNC_PWM);
+    uint slice = pwm_gpio_to_slice_num(gpio);
+    uint channel = pwm_gpio_to_channel(gpio);
+
+
+    uint32_t source_hz = clock_get_hz(clk_sys);
+        #define TOP_MAX 65534
```
lurch, line 36: This `#define` probably belongs near the top of the file with the others?

pwm/servo/main.c
```diff
@@ -0,0 +1,19 @@
+#include "pico/stdlib.h"
+#include "servo.h"
+#include <stdbool.h>
+
+int main()
+{
+    for(;;)
+    {
+        servo_init(1, 50);
```
lurch, line 9: I've not tried running this code, but I _suspect_ that `servo_init` probably ought to be before the `for` loop, rather than inside it?

SEEK-THE-VOID-LIGHT: oh yeah sorry forgot

pwm/servo/servo.c
```diff
@@ -0,0 +1,56 @@
...
+
+void servo_init(int gpio, int base_frequency)
+{
+    gpio_set_function(gpio, GPIO_FUNC_PWM);
+    uint slice = pwm_gpio_to_slice_num(gpio);
+    uint channel = pwm_gpio_to_channel(gpio);
+
+
+    uint32_t source_hz = clock_get_hz(clk_sys);
+        uint freq = 50;
```
lurch, line 37: Inconsistent indentation.

pwm/servo/servo.c
```diff
@@ -0,0 +1,56 @@
...
+    int duty_u16 = map(angle, 0, 180, DUTY_MIN, DUTY_MAX);
+    uint32_t cc = duty_u16 * (top + 1) / 65535;
+    pwm_set_chan_level(slice, channel, cc);
+    pwm_set_enabled(slice, true);
+
+    if(wait_write)
+        sleep_ms(1000);
+}
+
+void servo_init(int gpio, int base_frequency)
```
lurch, line 29: This `base_frequency` parameter doesn't seem to actually be used anywhere??

pwm/servo/servo.c
```diff
@@ -0,0 +1,56 @@
...
+    uint slice = pwm_gpio_to_slice_num(gpio);
+    uint channel = pwm_gpio_to_channel(gpio);
+    uint32_t top = pwm_hw->slice[slice].top;
+    int duty_u16 = map(angle, 0, 180, DUTY_MIN, DUTY_MAX);
+    uint32_t cc = duty_u16 * (top + 1) / 65535;
+    pwm_set_chan_level(slice, channel, cc);
+    pwm_set_enabled(slice, true);
+
+    if(wait_write)
+        sleep_ms(1000);
```
lurch, line 26: This is documented as "@param wait_write Wait for write to finish", in which case 1000ms seems like an excessively long time? :shrug:

voidlight7718: My motors are slow and with a smaller delay my motor do not finish the turn properly

PR #143, opened 2021-07-15
Add example for the MCP9808 Temperature sensor 

aayushijhaveri:
This PR reads and writes to the registers on the MCP9808 board. It sets the upper, lower and critical temperatures and reads the ambient temperature to 4dp by performing appropriate conversions. It checks to see that none of the limits have been surpassed and displays alerts if otherwise.


aallan:
If you could switch to using a mini-board rather than a full sized one, and change all (5!? those three pins on the right all go to GND? Weird board) the wires connected to the GND rail to be black that'd be great.

mcp9808_i2c

i2c/mcp9808_i2c/README.adoc
```diff
@@ -0,0 +1,37 @@
...
+== Bill of Materials
+
+.A list of materials required for the example
+[[mcp9808-bom-table]]
+[cols=3]
+|===
+| *Item* | *Quantity* | Details
+| Breadboard | 1 | generic part
+| Raspberry Pi Pico | 1 | http://raspberrypi.org/
+| MCP9808 board| 1 | generic part
```
aallan, line 33: You've used the Adafruit part, https://www.adafruit.com/product/1782, you should link to it.

i2c/mcp9808_i2c/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(mcp9808_i2c
+        mcp9808_i2c.c
+        )
+
+# Pull in our (to be renamed) simple get you started dependencies
```
matiasilva, line 5: A quick note here on this comment. In #132 this is being improved to be more descriptive, so would be good to get that in as well here, like this: https://github.com/raspberrypi/pico-examples/pull/132/files#diff-112d56e94629e0ec281a7eee133c1cc02557af71c1cd075466db5727ab7a42eaR5

matiasilva:
There's also the [general README](https://github.com/raspberrypi/pico-examples/blob/develop/README.md) that needs adding to so that people browsing the pico-examples repo on GitHub can find your example.

aallan:
Hey. so I think the Pico needs to be on the **right**, and the breakout on the **left** on this board. We can't really leave the USB socket on top of the breakout like that. Otherwise much cleaner!

image

lurch:
Apologies for the nit-picking, but could you connect the jumper wires _just_ to the breadboard, rather than connecting them directly to the PCB? That way the labels on the PCB don't get obscured by the wires, see https://github.com/raspberrypi/pico-examples/blob/a73bbdcc9d5336059f81c8099a8af0b0061283bc/spi/eink_mono_spi/eink_mono_spi_bb.png for an example.

@matiasilva If you have the time, could you check the wiring diagrams in your other PRs do the same? Consistency is good :wink:

lurch:
> (those three pins on the right all go to GND? Weird board)

Looks like they're just for setting the I2C address?

aallan:
> Looks like they're just for setting the I2C address?

Huh. Interesting.

i2c/mcp9808_i2c/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+# pull in common dependencies and additional i2c hardware support
```
lurch, line 6: Now you've added the second line, you can remove the first line :wink:

i2c/mcp9808_i2c/README.adoc
```diff
@@ -0,0 +1,37 @@
+= Attaching a MCP9808 digital temperature sensor via I2C
+
+This example code shows how to interface the Raspberry Pi Pico to the MCP9808 digital temperature sensor board. 
+======
+This example reads the ambient temperature value each second from the sensor and sets upper, lower and critical limits for the temperature and checks if alerts need to be raised. The CONFIG register can also be used to check for an alert if the critical temperature is surpassed.
+======
+
+== Wiring information
+
+Wiring up the device requires 7 jumpers, to connect VDD, GND, SDA, SCL, A0, A1 and A2 pins. An additional 2 jumpers could also be required to connect the Pico GND and Pico VSYS pin to the power rail on the breadboard. The example here uses I2C port 0, which is assigned to GPIO 4 (SDA) and 5 (SCL) in software. Power is supplied from the VSYS pin.
```
lurch, line 10: "An additional 2 jumpers could also be required..."
It's probably less confusing to a beginner if you _only_ document the wiring that you've used in the wiring diagram, rather than offering them alternatives?

aayushijhaveri:
> > (those three pins on the right all go to GND? Weird board)
> 
> Looks like they're just for setting the I2C address?
> EDIT: Although https://www.adafruit.com/product/1782 says "The PCB includes mounting holes, and pull down resistors for the 3 address pins." so I guess you probably don't need to manually connect them to ground?

Would it be better to remove them from the Fritzing file then? Or leave them in incase they potentially want to use a different address?

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
+/**
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
```
lurch, line 2: @aallan Should we be using a 2021 copyright year for the newly-added examples?

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+#include "hardware/i2c.h"
+
+/* Example code to talk to a MCP9808 ±0.5°C Digital temperature Sensor
+   
+   This reads and writes to registers on the board. 
+
+   Connections on Raspberry Pi Pico board, other boards may vary.
+
+   GPIO PICO_DEFAULT_I2C_SDA_PIN (On Pico this is 4 (pin 6)) -> SDA on MCP9808 board
+   GPIO PICO_DEFAULT_I2C_SCK_PIN (On Pico this is 5 (pin 7)) -> SCL on MCP9808 board
```
lurch, line 20: I think the 4 and 5 here should be GP4 and GP5 ? See https://github.com/raspberrypi/pico-examples/pull/132/commits/add29c9955ff35f2f95551ea76c28f0defcdf5f0

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+   Vsys (pin 39) -> VDD on MCP9808 board
+   GND (pin 38)  -> GND on MCP9808 board
+
+*/
+//The bus address is determined by the state of pins A0, A1 and A2 on the MCP9808 board
+static uint8_t ADDRESS = 0X18;
+
+const uint8_t WRITE_MODE = 0XFE;
+const uint8_t READ_MODE = 0XFF;
```
lurch, line 29: I think we generally use `0x...` rather than `0X...`

i2c/mcp9808_i2c/README.adoc
aayushijhaveri: All done! Requires 9 jumpers now

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+    //Clear flag bits
+    upper_byte = upper_byte & 0x1F; 
+
+
+    //Check if TA <= 0°C and convert to denary accordingly
+    if ((upper_byte & 0x10) == 0x10) { 
+        upper_byte = upper_byte & 0x0F; 
+        temperature = 256 - (((float)upper_byte * 16) + ((float)lower_byte / 16));
+    }
+    else {
```
lurch, line 80: This should be
```
} else {
```
(reference: https://github.com/raspberrypi/pico-examples/pull/127#discussion_r660801824 )

i2c/mcp9808_i2c/mcp9808_i2c.c
aallan: I'd prefer to see something like "GP4 (physical pin 6)"

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+    }
+    if ((upper_byte & 0x20) == 0x20) { //TA < TLOWER
+        printf("Temperature is below the lower temperature limit.\n");
+    }
+    if ((upper_byte & 0x80) == 0x80) { //TA < TCRIT
+        printf("Temperature is above the critical temperature limit.\n");
+    }
+}
+
+float mcp9808_convert_temp(uint8_t buf[]){
```
lurch, line 59: @matiasilva Do you think it might make sense to have this return an integer representation of the temperature, and then only convert it to float when displaying the value, like you did in your temperature-sensor code? :shrug:

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+void mcp9808_check_limits(uint8_t upper_byte) {
+
+    // Check flags and raise alerts accordingly 
+    if ((upper_byte & 0x40) == 0x40){ //TA > TUPPER
+        printf("Temperature is above the upper temperature limit.\n");
+    }
+    if ((upper_byte & 0x20) == 0x20) { //TA < TLOWER
+        printf("Temperature is below the lower temperature limit.\n");
+    }
+    if ((upper_byte & 0x80) == 0x80) { //TA < TCRIT
```
lurch, line 54: Should the comment here be `//TA > TCRIT` ? (which is what the printf is saying?)

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+    }
+    else {
+        temperature = (((float)upper_byte * 16) + ((float)lower_byte / 16));
+  
+    }
+    return temperature;
+}
+
+
+void mcp9808_read_temp(){
```
lurch, line 88: Space between `)` and `{` please.

i2c/mcp9808_i2c/mcp9808_i2c.c
aayushijhaveri: Whoops yes

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+    
+    //Set an upper limit of 30°C for the temperature
+    uint8_t upper_temp_msb = 0x01;
+    uint8_t upper_temp_lsb = 0xE0;
+
+    //Set an lower limit of 30°C for the temperature
+    uint8_t lower_temp_msb = 0x01;
+    uint8_t lower_temp_lsb = 0x40;
+
+    //Set an critical limit of 30°C for the temperature
```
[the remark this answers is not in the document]
aayushijhaveri: Ah no, sorry went too far with the copying and pasting, will change that

aallan:
Just noticed. One of your GND wires goes to the wrong rail. 

image

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+    return temperature;
+}
+
+
+void mcp9808_read_temp(){  
+
+    uint8_t buf[2];
+    float temperature;
+
+    while (1) {
```
lurch, line 93: IMHO it feels kinda odd to have an infinite `while(1) {` "buried" inside a function? I think this loop should probably be done directly inside `main()` ?

aayushijhaveri:
> Just noticed. One of your GND wires goes to the wrong rail.
> 
> image

That's from the old commit, have changed it since and swapped around the 2 boards in the later commits

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+    // Start reading ambient temperature register for 2 bytes
+    i2c_write_blocking (i2c_default, 0x18 & WRITE_MODE ,&REG_TEMP_AMB,1,true);
```
lurch, line 95: Space after commas please

i2c/mcp9808_i2c/mcp9808_i2c.c
aayushijhaveri: Sure, can move the whole read_temp function into main

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+    i2c_read_blocking(i2c_default, 0x18 & READ_MODE, buf, 2, false);
+
+    temperature = mcp9808_convert_temp(buf);
```
lurch, line 98: _Perhaps_ it makes sense to change the signature of `mcp9808_convert_temp(uint8_t buf[])` to `mcp9808_convert_temp(uint8_t upper_byte, uint8_t lower_byte)` and then inside the `while` loop you could do:
```
    mcp9808_check_limits(buf[0] & 0xE0);

    temperature = mcp9808_convert_temp(buf[0] & 0x1F, buf[1]);
```
(makes things more logically separate IMHO)

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,161 @@
...
+    printf("Ambient temperature: %.4f°C\n", temperature);
+
+    sleep_ms(1000); 
+    }
```
lurch, line 102: Looks like there's something gone wrong with the indentation here? We use 4-spaces-per-indent, and we don't use tabs.

lurch:
Please don't be disheartened by the number of little nit-picky comments I've made, this looks like a good start! :+1:

lurch:
> > EDIT: Although https://www.adafruit.com/product/1782 says "The PCB includes mounting holes, and pull down resistors for the 3 address pins." so I guess you probably don't need to manually connect them to ground?
> Would it be better to remove them from the Fritzing file then? Or leave them in incase they potentially want to use a different address?

I think we want to make the examples as easy to follow as possible, i.e. if there's jumpers that the user can get away without using, then we should omit them!

aallan:
> I think we want to make the examples as easy to follow as possible, i.e. if there's jumpers that the user can get away without using, then we should omit them!

Agreed.

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -1,7 +1,7 @@
 /**
  * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
  *
- * SPDX-License-Identifier: BSD-3-Clause
+ * SPDx-License-Identifier: BSD-3-Clause
```
lurch, line 4: Doh, the `SPDX` should remain in uppercase (I guess you must have search'n'replaced `X` to `x`?)

i2c/mpu6050_i2c/README.adoc
```diff
@@ -36,7 +36,7 @@ mpu6050_i2c.c:: The example code.
 | *Item* | *Quantity* | Details
 | Breadboard | 1 | generic part
 | Raspberry Pi Pico | 1 | http://raspberrypi.org/
-| MPU6050 board| 1 | https://www.adafruit.com/product/1782
+| MPU6050 board| 1 | generic part
```
lurch: Huh?

i2c/mcp9808_i2c/mcp9808_i2c.c
lurch: Looks much nicer now :+1:

README.md
```diff
@@ -76,7 +76,8 @@ App|Description
 App|Description
 ---|---
 [bus_scan](i2c/bus_scan) | Scan the I2C bus for devices and display results.
-[lcd_1602_i2c](i2c/lcd_1602_i2c) | Display some text on a generic 16x2 character LCD display, via I2C.
+[lcd_1602_i2c](i2c/lcd_1602_i2c) | Display some text on a generic 16x2 character LCD display, via I2C
```
lurch, line 79: Looks like you (accidentally) removed the full-stop from the end of this line?

aallan:
> Please don't be disheartened by the number of little nit-picky comments I've made, this looks like a good start! 👍

Yup. This is very much normal.

i2c/mcp9808_i2c/README.adoc
```diff
@@ -0,0 +1,37 @@
+= Attaching a MCP9808 digital temperature sensor via I2C
+
+This example code shows how to interface the Raspberry Pi Pico to the MCP9808 digital temperature sensor board. 
+======
+This example reads the ambient temperature value each second from the sensor and sets upper, lower and critical limits for the temperature and checks if alerts need to be raised. The CONFIG register can also be used to check for an alert if the critical temperature is surpassed.
```
lurch, line 5: "and critical limits for the temperature and checks if alerts need to be raised. The CONFIG register can also be used to check for an alert if the critical temperature is surpassed." - is the second sentence redundant here?

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,153 @@
...
+   Connections on Raspberry Pi Pico board, other boards may vary.
+
+   GPIO PICO_DEFAULT_I2C_SDA_PIN (On Pico this is GP4 (physical pin 6)) -> SDA on MCP9808 board
+   GPIO PICO_DEFAULT_I2C_SCK_PIN (On Pico this is GP5 (physcial pin 7)) -> SCL on MCP9808 board
+   Vsys (physical pin 39) -> VDD on MCP9808 board
+   GND (physical pin 38)  -> GND on MCP9808 board
+
+*/
+//The bus address is determined by the state of pins A0, A1 and A2 on the MCP9808 board
+static uint8_t ADDRESS = 0x18;
```
lurch, line 26: Any reason that this is `static` rather than `const` ?

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,153 @@
...
+        printf("Temperature is above the critical temperature limit.\n");
+    }
+}
+
+float mcp9808_convert_temp(uint8_t upper_byte, uint8_t lower_byte) {
+
+    float temperature;
+
+
+    //Check if TA <= 0°C and convert to denary accordingly
```
lurch, line 64: denary = decimal?

i2c/mcp9808_i2c/README.adoc
```diff
@@ -0,0 +1,37 @@
+= Attaching a MCP9808 digital temperature sensor via I2C
+
+This example code shows how to interface the Raspberry Pi Pico to the MCP9808 digital temperature sensor board.
```
lurch, line 3: Is there a publicly-available datasheet available online somewhere that we can link to, for the benefit of people who might want to tweak this example to do other things?

i2c/mcp9808_i2c/README.adoc
aayushijhaveri: Yeah agreed, would be wise to leave it out, I haven't covered that in the program

i2c/mcp9808_i2c/README.adoc
aayushijhaveri: Yup! Will pop that on

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,153 @@
...
+
+    // This example will use I2C0 on the default SDA and SCL pins (4, 5 on a Pico)
+    i2c_init(i2c_default, 400 * 1000);
+    gpio_set_function(PICO_DEFAULT_I2C_SDA_PIN, GPIO_FUNC_I2C);
+    gpio_set_function(PICO_DEFAULT_I2C_SCL_PIN, GPIO_FUNC_I2C);
+    gpio_pull_up(PICO_DEFAULT_I2C_SDA_PIN);
+    gpio_pull_up(PICO_DEFAULT_I2C_SCL_PIN);
+    // Make the I2C pins available to picotool
+    bi_decl(bi_2pins_with_func(PICO_DEFAULT_I2C_SDA_PIN, PICO_DEFAULT_I2C_SCL_PIN, GPIO_FUNC_I2C));
+#endif
```
lurch, line 126: I think this `#endif` might need to go just before the final `}` ?

i2c/mcp9808_i2c/mcp9808_i2c.c
aayushijhaveri: Not particularly, I think I might have picked it up from some other programs I looked at. All changed now :)

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,153 @@
...
+
+    //Set a critical limit of 40°C for the temperature
+    uint8_t crit_temp_msb = 0x02;
+    uint8_t crit_temp_lsb = 0x80;
+
+    uint8_t buf[3];
+    buf[0] = REG_TEMP_UPPER;
+    buf[1] = upper_temp_msb;
+    buf[2] = upper_temp_lsb; 
+    i2c_write_blocking(i2c_default, ADDRESS, buf, 3, false);
```
lurch, line 94: I think the use of `i2c_default` here _might_ mean that the whole of this `mcp9808_set_limits()` function needs to be inside a `#ifdef i2c_default ... #endif` guard? Graham will be able to confirm/deny.

i2c/mcp9808_i2c/mcp9808_i2c.c
aayushijhaveri: Aha yes, fixed up

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,153 @@
...
+//hardware registers
+
+const uint8_t REG_POINTER = 0x00;
+const uint8_t REG_CONFIG = 0x01;
+const uint8_t REG_TEMP_UPPER = 0x02;
+const uint8_t REG_TEMP_LOWER = 0x03;
+const uint8_t REG_TEMP_CRIT = 0x04;
+const uint8_t REG_TEMP_AMB = 0x05;
+const uint8_t REG_MAN_ID = 0x06;
+const uint8_t REG_DEV_ID = 0x07;
```
lurch, line 41: Is it worth including these unused values? :shrug:

i2c/mcp9808_i2c/mcp9808_i2c.c
```diff
@@ -0,0 +1,153 @@
...
+    
+    uint8_t buf[2];
+    uint16_t upper_byte;
+    uint16_t lower_byte;
+
+    float temperature;
+
+    while (1) {
+        // Start reading ambient temperature register for 2 bytes
+        i2c_write_blocking(i2c_default, ADDRESS & WRITE_MODE, &REG_TEMP_AMB, 1, true);
```
matiasilva, line 138: The SDK adds these bits to the end for us. `i2c_write_blocking` expects a 7-bit address, the code then shifts that to the left by 1 bit and adds the appropriate read/write bit.

matiasilva: ie. we can get rid of this of write mode, read mode and the logical ands. on line 104 for example, this isn't done and the code still works, so proof that it's not needed :P

i2c/mcp9808_i2c/mcp9808_i2c.c
matiasilva: Hmm..my code used a fixed point conversion while this uses a floating point conversion, so I guess it makes sense to return a float here.

i2c/mcp9808_i2c/mcp9808_i2c.c
matiasilva: Yep, this is what I gathered as well.

i2c/mcp9808_i2c/mcp9808_i2c.c
aayushijhaveri: Have cut those out now :)

i2c/mcp9808_i2c/mcp9808_i2c.c
aayushijhaveri: Oh nice! Quite handy, gotten rid of all that then

i2c/mcp9808_i2c/mcp9808_i2c.c
aayushijhaveri: What about the config and resolution register? I haven't used them in my code but should I leave them there in case someone wants to use them or are we going for complete minimalism?

lurch: Question for @aallan ...

aallan: Me? Gods no. I'm not making code decisions like that, that's a @kilograham question.

i2c/mcp9808_i2c/mcp9808_i2c.c
lurch: > The SDK adds these bits to the end for us. i2c_write_blocking expects a 7-bit address, the code then shifts that to the left by 1 bit and adds the appropriate read/write bit.

 @matiasilva Maybe you want to send a PR to the SDK clarifying the documentation (in the doxygen) there, to prevent other users running into the same trap? :thinking:

matiasilva: Good call, see https://github.com/raspberrypi/pico-sdk/pull/520 :D

i2c/mcp9808_i2c/mcp9808_i2c.c
matiasilva: If you do include them then perhaps add some commented out example code showing how they *could* be used?

aallan: > If you do include them then perhaps add some commented out example code showing how they could be used?

That I will say "no" to, no commented out code in the examples.

aallan:

JamesH65:
Very difficult to review since the comments above that have been fixed are not flagged as fixed, and again, doesn't seem to be any way of resolving them.

PR #144, opened 2021-07-19
Add example for MMA8451 Accelerometer

aayushijhaveri:
This PR adds an example for MMA8451 accelerometer board. It translates and displays the acceleration values and demonstrates how to set the range and precision of the values.


aallan:
Where are we on this? @lurch can you take a look at this one?

JamesH65:
@aayushijhaveri Hi Aayushi, can you give me write access to all the PR's here in case I need to make change? Thanks.

PR #145, opened 2021-07-20
Added example for slave mode operation with the Pi Pico

SEEK-THE-VOID-LIGHT:
The code was tested with the Jetson Nano B01 as master.

Code for testing:
```python
#curcuitpython basis
import board
from time import sleep

i2c = board.I2C()

#toggle led on the pico
i2c.writeto(0x31, bytes([0x10]))

sleep(1)

#request to read echo
i2c.writeto(0x31, bytes([0x20]))
sleep(0.5)
buffer = bytearray(2)
i2c.readfrom_into(0x31, buffer)
print(buffer)
```


davidedelvento:
@voidlight7718 not sure if you saw my mention of your code in #192 -- if you are still interested in this after so much time, you might want to chime in there.

PR #146, opened 2021-07-24
bus_scan.c example - fix pin number in comment

a-h:
The comment in the code is incorrect. I tested this out, and it doesn't detect the device via those pins.

I was searching and found this comment thread: https://www.raspberrypi.org/forums/viewtopic.php?p=1822258&sid=1ff96911779abcedda7b3ae535d1d0af#p1822258

Sure enough, when I switched to pins 6 and 7, the code worked fine and could detect the device I had plugged into the i2c bus.

```
I2C Bus Scan
   0  1  2  3  4  5  6  7  8  9  A  B  C  D  E  F
00 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .
10 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .
20 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .
30 .  .  .  .  .  .  .  .  .  .  .  .  @  .  .  .
40 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .
50 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .
60 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .
70 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .
Done.
```

Example wiring.

pico

### DIFF
--- a/i2c/bus_scan/bus_scan.c
+++ b/i2c/bus_scan/bus_scan.c
@@ -38,7 +38,7 @@ int main() {
 #warning i2c/bus_scan example requires a board with I2C pins
     puts("Default I2C pins were not defined");
 #else
-    // This example will use I2C0 on the default SDA and SCL pins (4, 5 on a Pico)
+    // This example will use I2C0 on the default SDA and SCL pins (6, 7 on a Pico)
     i2c_init(i2c_default, 100 * 1000);
     gpio_set_function(PICO_DEFAULT_I2C_SDA_PIN, GPIO_FUNC_I2C);
     gpio_set_function(PICO_DEFAULT_I2C_SCL_PIN, GPIO_FUNC_I2C);
--- a/i2c/bus_scan/bus_scan.c
+++ b/i2c/bus_scan/bus_scan.c
@@ -38,7 +38,7 @@ int main() {
 #warning i2c/bus_scan example requires a board with I2C pins
     puts("Default I2C pins were not defined");
 #else
-    // This example will use I2C0 on the default SDA and SCL pins (6, 7 on a Pico)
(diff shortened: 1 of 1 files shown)


lurch:
The comment is correct, it's referring to the GPIO numbers, and the "default I2C pins" are defined in https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/pico.h

You seem to be confusing "GPIO numbers" with "physical pins" - see the diagram at https://www.raspberrypi.org/documentation/rp2040/getting-started/#board-specifications
So if you wanted to submit an alternative PR where you change "(4, 5 on a Pico)" to "(GP4, GP5 on a Pico)" I think we might be willing to accept that. ping @aallan ?

a-h:
Great. As you can see, I wasn't the first person to be confused by that, and wouldn't be the last.

Prefixing GPIO numbers with GP would remove any ambiguity about what the number means and it matches the pinout diagram (https://datasheets.raspberrypi.org/pico/Pico-R3-A4-Pinout.pdf).

a-h:
I've updated based on your suggestion.

lurch (dismissed):

JamesH65:
@lurch Can you approve this so that Github actually thinks its approved? Or don't you have write access? @kilograham?

aallan (dismissed):

aallan:
Done @JamesH65

JamesH65:
Feel free to merge

lurch:
> @lurch Can you approve this so that Github actually thinks its approved? Or don't you have write access?

IIRC the only Pico-related repo I have write access to is `pico-feedback`.

PR #98, opened 2021-04-05
Added Ethernet example(W5500 of WIZnet product) using SPI

irinakim12:
https://youtu.be/MDdt03a3NQU

PR #147, opened 2021-07-28
Add example for the W5500 Ethernet chip

aayushijhaveri:
This PR adds an example that interfaces the Pico with the W5500 using multiple boards. It is a continuation of PR#98. Changes include the importing of dependencies through CMake, a Fritzing diagram and added documentation on pulling in external files


aallan:
This PR is built on top of #98 and replaces it.

ethernet/w5500_spi/README.adoc
```diff
@@ -0,0 +1,95 @@
...
+This example simply sends and receives data from the W5500 chip using the polling approach. After you finish using the communication command* on the WIZ850io, ensure then the communication command set to clear.
+
+*The communication command is S"n"_CR (with "n" being the socket number).
+======
+
+== Importing external libraries
+
+In order to import the libraries needed for this example, you should first download all the files from the following respository: https://github.com/Wiznet/ioLibrary_Driver. Then in the CMakeLists.txt file in this directory replace "${YOUR_LIB_PATH_HERE}" on line 3 with the full path address of where the ioLibrary_driver folder is stored in your local machine.  
+
+The structure used to generally add additional libraries is:
```
JamesH65, line 16: This section might need a bit of rewrite, to make it clearer what is generic and what is specific to this example.

ethernet/w5500_spi/w5500_spi.c
```diff
@@ -0,0 +1,161 @@
...
+#include <string.h>
+#include "pico/stdlib.h"
+#include "pico/binary_info.h"
+#include "hardware/spi.h"
+#include "w5500.h"
+#include "wizchip_conf.h"
+#include "loopback.h"
+
+
+/* Example code to talk to a wiznet MEMS accelerometer and gyroscope.
```
JamesH65, line 17: This this comment is  a hangover from  a previous example?

aayushijhaveri: Ah I forgot to push the changes I made to that file into this branch. I'll commit them now

ethernet/w5500_spi/README.adoc
```diff
@@ -0,0 +1,95 @@
...
+== Bill of Materials
+
+.A list of materials required for the example
+[[W5500-bom-table]]
+[cols=3]
+|===
+| *Item* | *Quantity* | Details
+| Breadboard | 1 | generic part
+| Raspberry Pi Pico | 1 | http://raspberrypi.org/
+| W5500 board | 1 | generic part
```
matiasilva, line 91: Might be worth adding the specific part used in this example?

ethernet/w5500_spi/w5500_spi.c
```diff
@@ -0,0 +1,146 @@
...
+int main() {
+    int8_t phy_link = 0;
+    uint8_t id = 0;
+    stdio_init_all();
+
+    printf("Hello, W5500! Reading data via SPI...\n");
+
+    // This example will use SPI0 at 0.5MHz.
+    spi_init(SPI_PORT, 5000 * 1000);
+    gpio_set_function(PIN_MISO, GPIO_FUNC_SPI);
```
matiasilva, line 102: If we're using SPI0, then use the default SPI definitions: https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/pico.h#L49

ethernet/w5500_spi/w5500_spi.c
```diff
@@ -0,0 +1,146 @@
...
+*/
+
+#define PIN_SCK  2
+#define PIN_MOSI 3
+#define PIN_MISO 4
+#define PIN_CS   5
+#define PIN_RST  6
+
+#define SPI_PORT spi0
+#define READ_BIT 0x00
```
matiasilva, line 45: Not used anywhere

ethernet/w5500_spi/w5500_spi.c
```diff
@@ -0,0 +1,146 @@
...
+
+wiz_NetInfo gWIZNETINFO = { .mac = {0x00, 0x08, 0xdc, 0x78, 0x91, 0x71},
+                            .ip = {192, 168, 0, 15},
+                            .sn = {255, 255, 255, 255},
+                            .gw = {192, 168, 0, 1},
+                            .dns = {168, 126, 63, 1},
+                            .dhcp = NETINFO_STATIC};
+#define ETH_MAX_BUF 2048
+
+unsigned char ethBuf0[ETH_MAX_BUF];
```
matiasilva, line 56: This is ultimately code style, but is there a particular reason why some functions are declared here and then only defined after main while others are defined before main?

ethernet/w5500_spi/w5500_spi.c
```diff
@@ -0,0 +1,146 @@
...
+    asm volatile("nop \n nop \n nop");
+}
+
+static void wiznet_reset() {
+    gpio_set_dir(PIN_RST, GPIO_OUT);
+    gpio_put(PIN_RST, 0);
+    sleep_ms(100);
+    gpio_put(PIN_RST, 1);
+    sleep_ms(100);
+    bi_decl(bi_1pin_with_name(PIN_RST, "W5500 RST"));
```
matiasilva, line 78: Binary stuff should go in main, imo. Functionally I believe it's the same when using picotool, but there's no need to call it every time we reset the chip and all the other `bi_decl`s are in main anyway.

ethernet/w5500_spi/README.adoc
```diff
@@ -0,0 +1,95 @@
...
+   * GPIO 3 (physical pin 5) MOSI/spi0_tx -> MOSIon WIZ850io board
+   * GPIO 4 (physical pin 6) MISO/spi0_rx-> MISO on WIZ850io board
+   * GPIO 5 (physical pin 7) Chip select -> NCS on WIZ850io board
+   * GPIO 6 (physical pin 9) Reset/GPIO -> Reset on WIZ850io board
+   * 3.3v (physical pin 36) -> VCC on WIZ850io board
+   * GND (physical pin 38)  -> GND on WIZ850io board
+
+The example here uses SPI port 0. Power is supplied from the 3.3V pin.
+
+Raspberry Pi Pico Ethernet Shield:
```
matiasilva, line 57: Why all the newlines here?

ethernet/w5500_spi/README.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+More information at https://dokkotai.tistory.com/198.
```
matiasilva, line 59: Might be worth hiding the link with a macro: https://docs.asciidoctor.org/asciidoc/latest/macros/url-macro/

ethernet/w5500_spi/w5500_spi.c
```diff
@@ -0,0 +1,146 @@
...
+    asm volatile("nop \n nop \n nop");
+}
+
+static inline void cs_deselect() {
+    asm volatile("nop \n nop \n nop");
+    gpio_put(PIN_CS, 1);
+    asm volatile("nop \n nop \n nop");
+}
+
+static void wiznet_reset() {
```
matiasilva, line 72: Any particular reason for static?

ethernet/w5500_spi/w5500_spi.c
```diff
@@ -0,0 +1,146 @@
...
+    // Chip select is active-low, therefore initialise it to a driven-high state
+    gpio_init(PIN_CS);
+    gpio_set_dir(PIN_CS, GPIO_OUT);
+    gpio_put(PIN_CS, 1);
+    // Make the CS pin available to picotool
+    bi_decl(bi_1pin_with_name(PIN_CS, "SPI CS"));
+
+    wiznet_reset();
+
+    reg_wizchip_spi_cbfunc(wiznet_read, wiznet_write);
```
matiasilva, line 117: Might be worth adding a comment here for the reader. I can probably guess what this is doing but since this is an example, it would be nice to offer some explanation

ethernet/w5500_spi/w5500_spi.c
```diff
@@ -0,0 +1,146 @@
...
+        // Check phy status
+        if(ctlwizchip(CW_GET_PHYLINK, &phy_link) == -1){
+            printf("Unknow PHY LINK status.\r\n");
+            sleep_ms(10);
+        }
+    } while (phy_link == PHY_LINK_OFF);
+    ctlnetwork(CN_SET_NETINFO, &gWIZNETINFO);
+    print_network_information();
+
+    while(1) {
```
matiasilva, line 132: Same here with the comments.

ethernet/w5500_spi/README.adoc
aallan: Should it *should* work with any W5500 breakout. So the WIZ850io breakout, which I haven't been able to source at this point (seems to be unavailable due to chipageddon) but I think #98 used, or the [Ethernet Shield](https://dokkodai.tistory.com/198) which is what @aayushijhaveri used.

ethernet/w5500_spi/README.adoc
```diff
@@ -0,0 +1,95 @@
...
+      target_include_directories(*library_name ${source_directory_1_path} ${source_directory_2_path}*)
+
+   Finally, remember to add the library_name to the target_link_libraries() command in the CMakeLists.txt file in the same directory as your target.
+   
+
+
+
+== Wiring information
+
+Wiring information for the WIZ850io board and the Raspberry Pi Pico Ethernet shield, both containing the W5500 chip, is provided below.
```
aallan, line 39: So just FYI I have raised a [brand and trademark](https://www.raspberrypi.org/trademark-rules/) issue around the "Raspberry Pi Pico Ethernet Shield" name.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "ioLibrary_Driver"]
+	path = ethernet/w5500_spi/ioLibrary_Driver
+	url = https://github.com/Wiznet/ioLibrary_Driver
```
aallan, line 3: I'm not sure we want to have this library as an actual submodule of pico-examples.

kilograham: yeah i'm fine with that as most people will likely pull in code via a submodule. it is the catch 22 of the examples: how we'd probably ideally do something there is not necessarily a good example of how the user would do it in their standalone project.

So yeah, I think i'm fine with this (if you don't initialize the submodule you don't get the example).... this is of course a minor doc issue

aallan: >  …this is of course a minor doc issue

Yeah. I'm going to have to think how to add a level of subtlety into the documentation around that.

aayushijhaveri: I messed it up in the previous commit but have updated it since; the files shouldn't have copied over to pico-examples too. @kilograham said it would be best to add it in as a submodule. And it would stay sync with any library updates

ethernet/w5500_spi/w5500_spi.c
aayushijhaveri: Raises the following error when not static: "gpio_put' is static but used in inline function 'cs_deselect' which is not static". So might be best to leave it as that

ethernet/w5500_spi/README.adoc
aayushijhaveri: Ah alright, I wasn't too sure what the official name was but that's the closest I could get

ethernet/w5500_spi/README.adoc
```diff
@@ -0,0 +1,88 @@
...
+      target_include_directories(*library_name ${source_directory_path}*)
+
+   Finally, remember to add the library_name to the target_link_libraries() command in the CMakeLists.txt file in the same directory as your target.
+   
+== Wiring information
+
+Wiring information for the WIZ850io board and the Raspberry Pi Pico Ethernet shield, both containing the W5500 chip, is provided below.
+
+WIZ850io board:
+More information is available [here](https://eshop.wiznet.io/shop/module/wiz850io/).
```
aallan, line 41: Remember that this README.adoc will be imported into the PDF documentation. Please don't link using the word "here" or other methods like this, it's quite possible that this will be printed out, and in any case this is poor sentence construction.

aayushijhaveri: Right, should I be putting in link directly (no macro) or using a better descriptor than 'here'?

JamesH65:
I *think* most of the review comments above have been fixed. @aayushijhaveri Would you have time to flick through and resolve all the fixed stuff? Also, can you make the repo writable to me so I can push changes (and all your other ones as well)

The Fritzing diagram could do with some tweaks, but since I cannot push changes to the PR no point in fixing it yet.

aallan:
Closing.

PR #154, opened 2021-08-18
Add example for the LIS3DH accelerometer

aayushijhaveri:
This PR adds an example for the LIS3DH accelerometer and retrieves data from a built in temperature sensor


i2c/lis3dh_i2c/lis3dh_i2c.c
```diff
@@ -0,0 +1,132 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include "pico/stdlib.h"
+#include "pico/binary_info.h"
+#include "hardware/i2c.h"
+
+/* Example code to talk to a LIS3DH Mini GPS module.
```
lurch, line 13: Looks like the "Mini GPS module" here is probably a copy'n'paste typo from a different example?

PR #175, opened 2021-10-26
Another fix for #121 blink freq, but adds onboard LED

F3RR3T:
Also re-ordered the GPIO pins and frequencies to make them easier to probe.


peterharperuk:
Not sure what this is doing. Closing.

PR #183, opened 2021-11-05
Update pio/ir_nec/README.adoc file links

lurch:
The "List of Files" now links (only) to files, not to directories (which prevented this example being included in the C SDK databook)


aallan (approved):
LGTM. Can we get this merged @kilograham which will let me get it integrated into the SDK book.

PR #187, opened 2021-11-19
Fixup a coding error in BME280 example

JamesH65:
Confirmed as fixing the problem by reporter of the issue.


kilograham (commented):
"I'll give you a round f-ing peg"

kilograham (approved):

PR #200, opened 2022-02-18
SRGBmods Pico LED Controller

FeuerSturm:
to apply for being allocated a PID


SRGBmods_Pico_LED_Controller/README.md
```diff
@@ -0,0 +1,28 @@
+![](https://srgbmods.net/picoledcontroller/img/PicoLEDController.png)
+### SRGBmods Pico LED Controller for SignalRGB
+
+This is an open-source DIY (do it yourself) project to create a small, affordable - yet still powerful - addressable RGB controller that can be managed by **SignalRGB** with sending packets via USB HID to the Pico.
+
+In the **/src/** folder you will find:
+- the **SRGBmods_Pico_LED_Controller.ino** sketch to compile with Arduino IDE
```
lurch, line 7: The examples in this `pico-examples` repo are all designed to be used in conjunction with https://github.com/raspberrypi/pico-sdk

I believe any Arduino-based example-projects will need to be added to whatever the appropriate Arduino-projects website would be.

lurch:
Are you looking for https://github.com/raspberrypi/usb-pid ?

FeuerSturm:
Hi, yes, I've been told to do it like this to get a USB PID as this is a non-commercial open-source software only project (NOT a company, just a fan project) with no selling of hardware, I have probably not been clear enough about not using pico-sdk.  
I've filled out the form as well to apply for a PID.

JamesH65:
I think this would need to be targeted at pico-sdk to progress further.

FeuerSturm:
> I think this would need to be targeted at pico-sdk to progress further.

Hey James,
thanks for your reply.
I have filled out the application form as well, or is there no chance to get a PID as long as it's neither a hardware production nor a software solution using pico-sdk?
Currently I am using the CDC UART pid for the project, but that would make "SignalRGB" spam usb packets to all connected Picos that use that pid.
Thanks in advance.
Stefan

ghollingworth:
Ah, didn't realise it wasn't using Pico SDK...

In this case it seems to me that you could just use the CDC UART PID and change the product descriptor name to be "SignalRGB" such that your code will know it is a SignalRGB device by retrieving the product name and comparing...

FeuerSturm:
> I think this would need to be targeted at pico-sdk to progress further.

Hey James,
thanks for your reply.
I have filled out the application form as well, or is there no chance to get a PID as long as it's neither a hardware production nor a software solution using pico-sdk?
Currently I am using the CDC UART pid for the project, but that would make "SignalRGB" spam usb packets to all connected Picos that use that pid.
Thanks in advance.
Stefan

> Ah, didn't realise it wasn't using Pico SDK...
> 
> In this case it seems to me that you could just use the CDC UART PID and change the product descriptor name to be "SignalRGB" such that your code will know it is a SignalRGB device by retrieving the product name and comparing...

I will do so instead then, thanks for your reply!

PR #211, opened 2022-05-13
Fix PIO count

graeme-winter:
Works fine currently for low frequency, breaks down at higher frequency e.g. 1MHz

Chased down with a scope to the 3 extra cycles needed for wait+1; `mov`; `jmp` so account for these in demo program. When running now at 1MHz get the right 1µs delay on a scope.


graeme-winter:
With these changes, demand for 100 cycles / pulse gives 1.25MHz:

F0001TEK

Without, we are ~ 6% off

kilograham (approved):

PR #233, opened 2022-07-01
Improve ap example

peterharperuk:
Adds a DNS server and changes the AP example to use it
The AP example now serves a web page /ledtest that allows you to turn the led on and off.

### DIFF
--- a/pico_w/access_point/picow_access_point.c
+++ b/pico_w/access_point/picow_access_point.c
@@ -6,86 +6,239 @@
 
 #include <string.h>
 
-#include "pico/stdlib.h"
 #include "pico/cyw43_arch.h"
+#include "pico/stdlib.h"
 
 #include "lwip/pbuf.h"
 #include "lwip/tcp.h"
 
 #include "dhcpserver.h"
-
-#ifndef USE_LED
-#define USE_LED 1
-#endif
+#include "dnsserver.h"
 
 #define TCP_PORT 80
 #define DEBUG_printf printf
+#define POLL_TIME_S 5
+#define HTTP_GET "GET"
+#define HTTP_RESPONSE_HEADERS "HTTP/1.1 %d OK\nContent-Length: %d\nContent-Type: text/html; charset=utf-8\nConnection: close\n\n"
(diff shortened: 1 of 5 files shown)


kilograham (commented):
picow_dns_server_test doesn't sound like an example name!

pico_w/access_point/picow_access_point.c
```diff
@@ -6,86 +6,236 @@
...
-#define USE_LED 1
-#endif
+#include "dnsserver.h"
 
 #define TCP_PORT 80
 #define DEBUG_printf printf
+#define POLL_TIME_S 5
+#define HTTP_GET "GET"
+#define HTTP_RESPONSE_HEADERS "HTTP/1.1 %d OK\nContent-Length: %d\nContent-Type: text/html; charset=utf-8\nConnection: close\n\n"
+#define LED_TEST_BODY "<html><body><h1>Hello from Pico.</h1><p>Led is %s</p><p><a href=\"?led=%d\">Turn led %s</a></body></html>"
```
lurch, line 23: Is it worth tweaking this to "Hello from Pico W". ? (for consistency with e.g. the MicroPython examples)

pico_w/access_point/picow_access_point.c
```diff
@@ -6,86 +6,236 @@
...
+                    cyw43_gpio_set(&cyw43_state, 0, true);
+                } else {
+                    // Turn led off
+                    cyw43_gpio_set(&cyw43_state, 0, false);
+                }
+            }
+        }
+        // Generate result
+        len = snprintf(con_state->result, sizeof(con_state->result), LED_TEST_BODY,
+            led_state ? "ON" : "OFF", led_state ? 0 : 1, led_state ? "OFF" : "ON");
```
lurch, line 108: Looks a bit odd having `led_state ? ...` in here 3 times? Would it be "better" to just use a regular `if / else` block?

Also, I've just spotted that this is writing directly to `con_state->result` rather than the `result` pointer which is passed in the function's args?

peterharperuk:
Checked this - still works nicely. Run it and connect to the picow_test wifi network on your phone and it should open a web page that allows you to turn the led on and off.

kilograham (dismissed):

peterharperuk:
Rebased and retested

kilograham (approved):

PR #237, opened 2022-07-11
Use the flash chip ID as the USB serial number

Spegs21:
Simple change to use the flash chip ID as the USB serial number with the TinyUSB example. I can squash/rebase as needed.

### DIFF
--- a/usb/device/dev_hid_composite/usb_descriptors.c
+++ b/usb/device/dev_hid_composite/usb_descriptors.c
@@ -178,7 +178,7 @@ uint8_t const * tud_descriptor_configuration_cb(uint8_t index)
 // String Descriptors
 //--------------------------------------------------------------------+
 
-// buffer to hold flash chip id
+// buffer to hold flash ID
 char serial[2 * PICO_UNIQUE_BOARD_ID_SIZE_BYTES + 1];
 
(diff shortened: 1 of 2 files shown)


usb/device/dev_hid_composite/usb_descriptors.c
```diff
@@ -177,13 +178,16 @@ uint8_t const * tud_descriptor_configuration_cb(uint8_t index)
...
+char serial[2 * PICO_UNIQUE_BOARD_ID_SIZE_BYTES + 1];
+
 // array of pointer to string descriptors
 char const* string_desc_arr [] =
 {
   (const char[]) { 0x09, 0x04 }, // 0: is supported language is English (0x0409)
   "TinyUSB",                     // 1: Manufacturer
   "TinyUSB Device",              // 2: Product
-  "123456",                      // 3: Serials, should use chip ID
+  serial,                        // 3: Serials, should use chip ID
```
lurch, line 190: Might be worth changing "should use chip ID" to "uses flash chip ID"? :wink:

Spegs21: Good call, I overlooked the comment.

kilograham (approved):

PR #246, opened 2022-07-21
Add Python test scripts and rename MicroPython scripts

mucahitkurtlar:
In Pico W TCP example test script was written in MicroPython. I needed to test on my desktop so I wrote tests in Python (not MicroPython) and think other people may need these test scripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,13 +114,13 @@ These eaxmples are for the Pico W, and are only available for `PICO_BOARD=pico_w
 
 App|Description
 ---|---
-[picow_access_point](pico_w/picow_access_point)| Starts a WiFi access point, and fields DHCP requests.
-[picow_blink](pico_w/picow_blink)| Blinks the on-board LED (which is connected via the WiFi chip).
-[picow_iperf_server](pico_w/picow_iperf)| Runs an "iperf" server for WiFi speed testing.
-[picow_ntp_client](pico_w/picow_ntp_client)| Connects to an NTP server to fetch and display the current time.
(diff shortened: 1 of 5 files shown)


lurch:
Because of the simultaneous rename-and-add, it can be a bit tricky to see exactly what this PR is doing, so just to help make things clearer:
* `pico_w/python_test_tcp/python_test_tcp_client.py` has been renamed to `pico_w/python_test_tcp/micropython_test_tcp_client.py` (without any extra edits)
* `pico_w/python_test_tcp/python_test_tcp_server.py` has been renamed to `pico_w/python_test_tcp/micropython_test_tcp_server.py` (without any extra edits)
* A new `pico_w/python_test_tcp/python_test_tcp_client.py` script has been added (which uses "regular Python" rather than MicroPython)
* A new `pico_w/python_test_tcp/python_test_tcp_server.py` script has been added (which uses "regular Python" rather than MicroPython)

@peterharperuk @kilograham @aallan I wonder if it _might_ make sense to move the `micropython_*.py` scripts into the https://github.com/raspberrypi/pico-micropython-examples repo?

peterharperuk:
yeah - we could move the micropython versions. We just need something to take the place of the other end of the C code examples. It was remiss of me to make micropython only versions. Bonus points if you can make a version that is OS agnostic.

peterharperuk:
Bit confused why there are 3 commits in here. It looks like they need squashing into one before they can be merged?

kilograham:
> Bit confused why there are 3 commits in here. It looks like they need squashing into one before they can be merged?

number of commits doesn't matter, I will squash when i merge

mucahitkurtlar:
So... Why are we still waiting? (:

peterharperuk (approved):
All looks good to me.

kilograham (approved):

PR #256, opened 2022-08-15
Add iperf client example

### LINKED ISSUE #255: iperf example fails to build when compiled as a client
Complains that xstr is not defined.

peterharperuk:
Requires you to specify the server ip address when running cmake,
e.g. -DIPERF_SERVER_IP=1.2.3.4

Fixes #255

### DIFF
--- a/pico_w/iperf/picow_iperf.c
+++ b/pico_w/iperf/picow_iperf.c
@@ -15,7 +15,7 @@
 #define USE_LED 1
 #endif
 
-#if CLIENT_TEST && !defined(IPERF_SERVER_IP)
+#if IPERF_CLIENT && !defined(IPERF_SERVER_IP)
 #error IPERF_SERVER_IP not defined
 #endif
 
@@ -38,7 +38,6 @@ static void iperf_report(void *arg, enum lwiperf_report_type report_type,
 
 int main() {
     stdio_init_all();
-
     if (cyw43_arch_init()) {
         printf("failed to initialise\n");
         return 1;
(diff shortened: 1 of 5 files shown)


lurch:
Perhaps the top-level `README.md` also wants updating to match?

peterharperuk:
> Perhaps the top-level `README.md` also wants updating to match?

ok - tweak the wording

pico_w/iperf/picow_iperf.c
```diff
@@ -53,9 +52,9 @@ int main() {
     }
 
 #if CLIENT_TEST
```
kilograham, line 54: maybe we could rename this vestigial "_TEST" from when these were tests ;-)

lurch:
I know nothing about FreeRTOS, but looking at the content of the diff in this PR, does `xstr` need removing from `picow_freertos_iperf.c` too? :shrug:

peterharperuk:
> I know nothing about FreeRTOS, but looking at the content of the diff in this PR, does `xstr` need removing from `picow_freertos_iperf.c` too? 🤷

Oops - yes. I should test it really!

lurch:
@kilograham / @liamfraser I wonder if it might be worth modifying https://github.com/raspberrypi/pico-examples/blob/master/.github/workflows/cmake.yml to do several different builds, with different combinations of `#define` s set? :man_shrugging:

peterharperuk:
Added iperf client to freertos and checked they work.

peterharperuk:
> @kilograham / @liamfraser I wonder if it might be worth modifying https://github.com/raspberrypi/pico-examples/blob/master/.github/workflows/cmake.yml to do several different builds, with different combinations of `#define` s set? 🤷‍♂️

Maybe this should be a separate issue? I don't really understand how the builds are configured but it looks a bit like pico_w examples are not built? To build everything...

-DPICO_BOARD=pico_w  -DWIFI_SSID='xxx' -DWIFI_PASSWORD='xxx' -DFREERTOS_KERNEL_PATH=/source/FreeRTOS-Kernel

lurch:
> Maybe this should be a separate issue?

I'll let @kilograham decide if he wants to create an issue for it, since he knows _far_ more about the SDK (and how its builds work) than I do :wink:

peterharperuk:
True - it should build for all boards

peterharperuk:
I can't be bothered with this

PR #259, opened 2022-08-25
Add isr example with the pio

dangarbri:
Someone on the pico forum asked [how interrupts work with only 2 interrupt lines and 4 state machines](https://forums.raspberrypi.com/viewtopic.php?p=2032559#p2032390).

I put together this example, which only runs with one isr line, but shows how you could identify which state machine the irq is coming from.

The pio program simply stalls waiting for input to the tx fifo from firmware, and then fires an interrupt when it gets anything.

From main.c, I load the program into each pio and fire the interrupts one by one. The code shows how you can identify the state machine the interrupt came from and then dispatch some action.

### DIFF
--- a/pio/isr/CMakeLists.txt
+++ b/pio/isr/CMakeLists.txt
@@ -1,0 +1,18 @@
+add_executable(isr)
+
+pico_generate_pio_header(isr ${CMAKE_CURRENT_LIST_DIR}/simply_isr.pio)
+
+target_sources(isr PRIVATE main.c)
+
+pico_enable_stdio_usb(isr 1)
+pico_enable_stdio_uart(isr 0)
+
+target_link_libraries(isr PRIVATE
+        pico_stdlib
+        hardware_pio
+        )
+
+pico_add_extra_outputs(isr)
+
+# add url via pico_set_program_url
+example_auto_set_url(isr)
--- a/pio/isr/main.c
+++ b/pio/isr/main.c
@@ -1,0 +1,193 @@
+/**
+ * Copyright 2022 (c) Daniel Garcia-Briseno
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+ * following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+ *    disclaimer.
+ * 
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+ *    disclaimer in the documentation and/or other materials provided with the distribution.
+ * 
+ * 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products
+ *    derived from this software without specific prior written permission.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+ * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <stdio.h>
+#include "pico/stdlib.h"
+
+#include "simply_isr.pio.h"
+// the sdk doesn't give us functions for figuring out which
+// isr source we're using, so we have to do it manually.
+#include "hardware/regs/pio.h"
+
+/**
+ * These defines represent each state machine.
+ * The value is the bit in the IRQ register that
+ * will be set by each state machine thanks to "irq wait 0 rel"
+ */
+#define PIO_SM_0_IRQ 0b0001
+#define PIO_SM_1_IRQ 0b0010
+#define PIO_SM_2_IRQ 0b0100
+#define PIO_SM_3_IRQ 0b1000
+
+/**
+ * This variable will shadow the IRQ flags set by the PIO state machines.
+ * Typically you do not want to do work in ISRs because the main thread
+ * has more important things to do. Because of that, when we get the ISR
+ * I'm simply going to copy the state machine that fired the ISR into
+ * this variable.
+ *
+ * Variable is volatile so that it doesn't get cached in a CPU register
+ * in the main thread. Without this it's possible that you never see
+ * irq_flags get set even though the ISR is firing all the time.
+ *
+ * Of course, you can really do whatever you want in the ISR, it's up to you.
+ */
+volatile uint32_t irq_flags = 0;
+
+/**
+ * Reads a hardware register
+ *
+ * @note See datasheet and/or pico-sdk/src/rp2040/include/hardware/regs
+ *       for base addresses and offsets
+ * @param[in] bank The base address of the register set usually <MODULE>_BASE
+ * @param[in] offset The register offset, usually <REG_NAME>_OFFSET
+ */
+uint32_t read_register(uint32_t bank, uint32_t offset) {
+    return *((volatile io_rw_32*)(bank + offset));
+}
+
+/**
+ * Writes a hardware register
+ *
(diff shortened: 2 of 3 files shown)


pio/isr/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+add_executable(isr)
```
lurch, line 1: Looking at some of the other examples in https://github.com/raspberrypi/pico-examples/tree/master/pio it probably makes sense to use `pio_isr` in `CMakeLists.txt` rather than just `isr` ? :shrug:

pio/isr/main.c
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/**
+ * This function is called when the IRQ is fired by the state machine.
+ * @note See enable_pio_isrs for how to register this function to be called
+ */
+void simply_isr_handler() {
+    // Read the IRQ register to get the IRQ flags from the state machine
+    // This tells me which state machine sent the IRQ
+    irq_flags = read_register(PIO0_BASE, PIO_IRQ_OFFSET);
```
lurch, line 90: Looks like you can also get this from https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2040/hardware_structs/include/hardware/structs/pio.h#L127-L130 which might be "less lowlevel" than defining your own `read_register` and `write_register` functions? :shrug: 

Or there's also https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_pio/include/hardware/pio.h#L799-L822 which is probably even better?

pio/isr/main.c
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/**
+ * Loads simply_isr pio program into PIO memory
+ */
+void load_pio_programs() {
+    PIO pio = pio0;
+
+    // Load the program into PIO memory
+    uint offset = pio_add_program(pio, &simply_isr_program);
```
lurch, line 121: Only a small nitpick, but I wonder if `simple_isr_program` is a better name than `simply_isr_program` ?

pio/isr/main.c
dangarbri: Oh I didn't notice the `pio_interrupt_get`. I think that will work better for this example.

pio/isr/main.c
dangarbri: That is a better name. I think in my mind I was going for something like the orange juice brand name.

pio/isr/main.c
dangarbri: Updated to use pio_interrupt_get.
I think personally I would prefer the struct version to read the IRQ, just because it's 1 read vs. 4 separate reads for `pio_interrupt_get`. But as an sdk example I think this is good.

pio/isr/main.c
dangarbri: Updated

lurch:
Nice, thanks for addressing my comments :+1: 
Would you mind also adding this new example to https://github.com/raspberrypi/pico-examples/blob/master/README.md#pio please?

dangarbri:
> Nice, thanks for addressing my comments +1 Would you mind also adding this new example to https://github.com/raspberrypi/pico-examples/blob/master/README.md#pio please?

done

README.md
```diff
@@ -156,6 +156,7 @@ App|Description
 [uart_tx](pio/uart_tx)| Implement the transmit component of a UART serial port, and print hello world.
 [ws2812](pio/ws2812)| Examples of driving WS2812 addressable RGB LEDs.
 [addition](pio/addition)| Add two integers together using PIO. Only around 8 billion times slower than Cortex-M0+.
+[pio_isr](pio/isr)| Respond to interrupts generated by a PIO state machine.
```
lurch, line 159: For consistency with the other examples listed here, should this be `[isr](pio/isr)` ? Sorry for being picky!

pio/isr/main.c
```diff
@@ -0,0 +1,178 @@
...
+ * I'm going to copy the state machine that fired the ISR into
+ * this variable using the above defines.
+ *
+ * Variable is volatile so that it doesn't get cached in a CPU register
+ * in the main thread. Without this it's possible that you never see
+ * irq_flags get set even though the ISR is firing all the time.
+ *
+ * Of course, you can really do whatever you want in the ISR, it's up to you.
+ */
+volatile uint32_t irq_flags = 0;
```
lurch, line 53: Probably makes no difference, but I wonder if a `uint8_t` would be more appropriate here, given that there are only 8 PIO SMs in total?

README.md
dangarbri: No problem, I understand! I used pio_isr because it's the name of the program that gets compiled, but I can change it to isr if you'd like.

pio/isr/main.c
dangarbri: u8 would be more efficient here. I had a u32 because originally I had that register read function which returned a u32 even though I only cared about the lower 8 bits.

README.md
dangarbri: Updated, but also renamed.

pio/isr/main.c
dangarbri: Updated to use u8

dangarbri:
@lurch I made your requested review changes, and also a couple more.

I renamed it to IRQ since that's a more standard name. I'm used to calling them ISR (interrupt service routine) which gets confusing because in pico world, ISR means Input Shift Register. So I renamed everything to IRQ.

Also I updated my IRQ handler to use the lower level read/clear method. In this case I think it really does make it simpler for this use case than `pio_interrupt_get`. It becomes one read to check the IRQ flags instead of iterating over each bit checking if its set.

pio/irq/main.c
```diff
@@ -0,0 +1,174 @@
...
+    simple_irq_program_init(pio, 2, offset);
+    simple_irq_program_init(pio, 3, offset);
+}
+
+/**
+ * Writes to the tx fifo of the given state machine.
+ * This will make the simple_irq program send an ISR to us!
+ */
+void trigger_irq(int sm) {
+    printf("Triggering ISR from state machine %d\n", sm);
```
lurch, line 113: ISR -> IRQ here? There's also a few comments that still say "ISR", where perhaps "IRQ" would be more appropriate? (although there _are_ some comments where it might be appropriate to stick to "ISR"?)

In my understanding (which may still be wrong, as I've not done much low-level coding), IRQ = Interrupt ReQuest, i.e. the actual hardware signal, and ISR = Interrupt Service Routine i.e. the callback function you've registered to handle the IRQ.

pio/irq/main.c
```diff
@@ -0,0 +1,174 @@
...
+    pio_sm_put_blocking(pio0, sm, 1);
+    // Interrupt will fire from the pio right here since they trigger
+    // whenever data goes into the TX FIFO, see simple_irq.pio
+
+    // Print the irq we expect based on the given state machine
+    printf("Expected IRQ flags: 0x%08X\n", (1 << sm));
+    printf("Actual IRQ Flags: 0x%08X\n", irq_flags);
+
+    // Here you could dispatch work for the irq depending one
```
lurch, line 122: "depending one" ? Looks like you might have missed some words out?

pio/irq/main.c
```diff
@@ -0,0 +1,174 @@
...
+    }
+    if (irq_flags & PIO_SM_2_IRQ) {
+        // handle_sm2_irq();
+    }
+    if (irq_flags & PIO_SM_3_IRQ) {
+        // handle_sm3_irq();
+    }
+    
+    // clear irq flags now.
+    irq_flags = 0;
```
lurch, line 141: Rather than setting _all_ `irq_flags` to 0 here at the end, I wonder if it might be more idiomatic to clear the flags individually e.g. by doing something like `irq_flags &= !PIO_SM_1_IRQ` inside the `if (irq_flags & PIO_SM_1_IRQ) {` block? Or perhaps that's unnecessary complication? :shrug:

pio/irq/simple_irq.pio
```diff
@@ -0,0 +1,72 @@
...
+; continuing. rel means let the irq be relative to the state machine.
+; by using "0 rel" on all state machines, software will see a different
+; interrupt source for each state machine. Technically state machines
+; can set any irq source, so in order to know where the irq is coming from
+; it's best to set 0 rel
+irq wait 0 rel
+
+% c-sdk {
+    /**
+     * Initializer for the fake irq program program
```
lurch, line 43: "program program" ?

pio/irq/main.c
```diff
@@ -0,0 +1,174 @@
...
+ * @note See enable_pio_irqs for how to register this function to be called
+ */
+void simple_irq_handler() {
+    // Check the interrupt source and set the flag accordingly
+    // This method can handle multiple simultaneous interrupts from the PIO
+    // since it checks all the irq flags. The advantage to this is that the
+    // ISR would execute only once when simultaneous interrupts happen instead
+    // of being executed for each interrupt individually.
+    // For other applications, make sure to check the correct pio (pio0/pio1)
+    irq_flags = pio0_hw->irq;
```
lurch, line 66: Perhaps some compilers will need an explicit cast here, to avoid warnings? :shrug:  (alternatively, I guess you could make `irq_flags` a `uint32_t` again)

pio/irq/main.c
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * Lets the pico know that we want it to notify us of the PIO ISRs.
+ * @note in simple_irq.pio we enable irq0. This tells the state machine
+ *       to send the ISRs to the core, we still need to tell the core
+ *       to send them to our program.
+ */
+void enable_pio_irqs() {
+    // Set the function that will be called when the PIO IRQ comes in.
+    irq_set_exclusive_handler(PIO0_IRQ_0, simple_irq_handler);
```
henrygab, line 82: I think this should be:

```C
irq_add_shared_handler(PIO0_IRQ_0, simple_irq_handler, e PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
```

After all, each state machine may have its own handler attached, as they could each run entirely distinct programs.

Issue #262, opened 2022-08-31
VSCode launch.json templates "runToMain" deprecated


jamd315:
In the templates launch-raspberrypi-swd.json and launch-remote-openocd.json under /ide/vscode, both use the deprecated debug attribute `runToMain: true`, which causes a warning to display in VSCode.  See also [cortex-debug debug attributes](https://github.com/Marus/cortex-debug/blob/master/debug_attributes.md)

![Screenshot from 2022-08-31 11-15-28](https://user-images.githubusercontent.com/6594497/187732528-db888452-b24d-426e-bdba-251ddd82fcc5.png)

lurch:
ping @JamesH65

JamesH65:
Yes, definitely a valid issue. Not sure if I can add the new command, which would be backwards compatible, or need to replace the runtomain completely, which is no so backwards compatible.

PR #263, opened 2022-08-31
Updated deprecated cortex-debug debug attributes

jamd315:
Attempts to fix issue #262, section 7.3 in the getting started PDF may also need to be updated but I wasn't sure how to do that.


lurch:
> section 7.3 in the getting started PDF may also need to be updated but I wasn't sure how to do that.

Probably best to file an issue at https://github.com/raspberrypi/pico-feedback

P33M:
LGTM - this matches what we have been testing internally

andygpz11 (commented):
This looks good to me, thanks.
I think the changes here are a sub-set of those I just tested and approved in #313
So, I'm not sure we need this if we merge #313 (but I'm happy).

kilograham:
closed in favor of #313

PR #264, opened 2022-09-01
wifi scan: Enable USB output

jepler:
It tripped me up that these examples required the use of a serial adapter, rather than working via usb serial.

### DIFF
--- a/pico_w/wifi_scan/CMakeLists.txt
+++ b/pico_w/wifi_scan/CMakeLists.txt
@@ -25,3 +25,6 @@ target_link_libraries(picow_wifi_scan_poll
         )
 pico_add_extra_outputs(picow_wifi_scan_poll)
 
+pico_enable_stdio_usb(picow_wifi_scan_background 1)
+pico_enable_stdio_usb(picow_wifi_scan_poll 1)
+


lurch:
This should be targeted against the `develop` branch rather than `master`.

kilograham:
Yeah, i mean we haven't really gone thru and made all the examples with USB stdio.

You could try passing -DPICO_STDIO_USB=1 to your cmake build and see what breaks; that should add stdio USB to everything (some things presumably will not work - we would need to at least disable it for the TinyUSB examples), but worth looking at as an option

peterharperuk:
It's probably a fair comment that we (or is it just me) wrongly default the examples to using the uart. If anything requires the user to look at the output it should default to usb? This example certainly works with this change but maybe we need to check for other issues like this?

peterharperuk:
It's probably a fair comment that we (or is it just me) wrongly default the examples to using the uart. If anything requires the user to look at the output it should default to usb? This example certainly works with this change but maybe we need to check for other issues like this?

aallan:
> It's probably a fair comment that we (or is it just me) wrongly default the examples to using the uart. If anything requires the user to look at the output it should default to usb? This example certainly works with this change but maybe we need to check for other issues like this?

Yes, it would be better if it defaulted to USB rather than Serial UART.

peterharperuk:
https://github.com/raspberrypi/pico-examples/issues/321

peterharperuk:
Closed in favour of the above issue which will deal with the issue

jepler:
Thanks for your attention to this. subscribed to the new bug.

PR #275, opened 2022-09-12
Example to send simple UDP packets, such as a beacon

pastcompute:
Hi, I needed to do this for myself and realised there was no simple example, so I created one


peterharperuk:
Looks ok to me. The ntp example was supposed to demonstrate udp but I guess it's too complicated?

pico_w/udp_beacon/picow_udp_beacon.c
```diff
@@ -0,0 +1,80 @@
...
+#include <stdlib.h>
+
+#include "pico/stdlib.h"
+#include "pico/cyw43_arch.h"
+
+#include "lwip/pbuf.h"
+#include "lwip/udp.h"
+
+#define UDP_PORT 4444
+#define DEBUG_printf printf
```
lurch, line 17: This `DEBUG_printf` define isn't actually used?

pastcompute: OK, fixed

pastcompute:
> Looks ok to me. The ntp example was supposed to demonstrate udp but I guess it's too complicated?

I figure simply sending packets is something many playing with the pico w  for the first time will often want to do (like I did), I did run the ntp example and it is very useful but being able to immediately see your own packets in short order is very satisfying

pico_w/udp_beacon/picow_udp_beacon.c
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        // Note in practice for this simple UDP transmitter,
+        // the end result for both background and poll is the same
+
+#if PICO_CYW43_ARCH_POLL
+        // if you are using pico_cyw43_arch_poll, then you must poll periodically from your
+        // main loop (not from a timer) to check for WiFi driver or lwIP work that needs to be done.
+        cyw43_arch_poll();
+        sleep_ms(BEACON_INTERVAL_MS);
```
peterharperuk, line 49: Looking at this again - it's a bit of bad form to block when polling. While it works, we might not want anyone to copy this...

pastcompute: I figured it would be fit for purpose for such a simple example... but I'll have a look and come up with something more advanced if that is needed

peterharperuk: Lets stick with what we've got for now. I can push another change to improve things later.

peterharperuk:
@kilograham It would be nice to get this example into develop

kilograham:
and yet you don't add to the README.txt ;-)

peterharperuk:
I pushed a new review with the readme added https://github.com/raspberrypi/pico-examples/pull/306

lurch:
Is this defunct, now that #306 has been merged?

peterharperuk:
Yes - these changes were merged via that pull request

PR #298, opened 2022-12-06
Fix build issues with spaces and parentheses

### LINKED ISSUE #297: Build issues with examples when path contains spaces or brackets

peterharperuk:
Add VERBATIM to add_custom_command

Fixes #297


lurch:
Are similar fixes also needed for pico-extras and pico-playground?

peterharperuk:
> Are similar fixes also needed for pico-extras and pico-playground?

I've not tested them, but they don't use add_custom_command, so I assume not.

kilograham (approved):

PR #307, opened 2023-01-03
Pico Probe friendly pinout; generalized LCD initialization sequence

kromych:
Faced two issues:
1. Could not use pico probe with the pinout from this example, as picoprobe requires the pins that are used for the example
2. My display has the height of 320, had to edit the code in two places.

Tested with Waveshare 240x320 2" LCD.

https://user-images.githubusercontent.com/745493/210317707-8a506e21-1171-4f5c-9275-6de3815f6f7f.mov

### DIFF
--- a/pio/st7789_lcd/st7789_lcd.c
+++ b/pio/st7789_lcd/st7789_lcd.c
@@ -20,12 +20,15 @@
 #define IMAGE_SIZE 256
 #define LOG_IMAGE_SIZE 8
 
-#define PIN_DIN 0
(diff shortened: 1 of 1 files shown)


lurch:
I _believe_ all the examples in this repo are intended to be standalone and (relatively) simple, so if you wanted to change which pins are used (so that you can modify the example to do other things at the same time) we'd expect you to do that in your local version, rather than making that change in this repo? :man_shrugging:

kromych:
I appreciate the advice very much! Will do.

PR #309, opened 2023-01-04
ping example app throws lwip assert

### LINKED ISSUE #308: ping example app throws lwip assert
The sys version.

peterharperuk:
Now LWIP_PLATFORM_ASSERT works it asserts for the ping example app. It sets the SO_RCVTIMEO socket option which is not implemented.

Fixes #308

### DIFF
--- a/pico_w/freertos/ping/lwipopts.h
+++ b/pico_w/freertos/ping/lwipopts.h
@@ -16,6 +16,10 @@
 
 // not necessary, can be done either way
 #define LWIP_TCPIP_CORE_LOCKING_INPUT 1
+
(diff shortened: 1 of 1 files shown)


pico_w/freertos/ping/lwipopts.h
```diff
@@ -16,6 +16,10 @@
 
 // not necessary, can be done either way
 #define LWIP_TCPIP_CORE_LOCKING_INPUT 1
+
+// ping_thread sets socket receive timeout to 1s
```
kilograham, line 20: this comment is a bit confusing since the value is 1 ... the #define is an enable, right?

kilograham (approved):

PR #314, opened 2023-01-09
Feature/display

cyndi5:
blink is now a digital display driver
a few CMakeList files tweaked to add default LED pin to 0 and enable USB serial

### DIFF
--- a/blink/CMakeLists.txt
+++ b/blink/CMakeLists.txt
@@ -2,6 +2,8 @@ add_executable(blink
         blink.c
         )
 
+target_compile_definitions(blink PUBLIC PICO_DEFAULT_LED_PIN=0)
+
 # pull in common dependencies
 target_link_libraries(blink pico_stdlib)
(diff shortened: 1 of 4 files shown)


lurch:
I'm not sure what the intention here was, but you might want to look at https://github.com/raspberrypi/pico-examples/tree/master/gpio/hello_7segment

cyndi5:
> I'm not sure what the intention here was, but you might want to look at https://github.com/raspberrypi/pico-examples/tree/master/gpio/hello_7segment

Thanks so much; that is very similar. I like the way that gpio/hello_7segment/hello_7segment.c uses the `bits[]` array to store a bitmap in hex form. Very compact. I was thinking it would be cool to map the ordinal ASCII characters (the ones that are discernable in 7 segments) to the bitmaps. Also, to extend to 15-segment alphanumeric, which I sadly don't have but could test on paper...

PR #319, opened 2023-01-23
Update on reset of mpu6050_i2c.c

akhodeir:
To reset the MPU6050, the device_reset bit shall be set to '1'.

reference from "MPU-6000/MPU-6050 Register Map and Descriptions"
DEVICE_RESET : When set to 1, this bit resets all internal registers to their default values. The bit automatically clears to 0 once the reset is done.

### DIFF
--- a/i2c/mpu6050_i2c/mpu6050_i2c.c
+++ b/i2c/mpu6050_i2c/mpu6050_i2c.c
@@ -37,7 +37,7 @@ static int addr = 0x68;
 static void mpu6050_reset() {
     // Two byte reset. First byte register, second byte data
     // There are a load more options to set up the device in different ways that could be added here
-    uint8_t buf[] = {0x6B, 0x00};
+    uint8_t buf[] = {0x6B, 0x80};
     i2c_write_blocking(i2c_default, addr, buf, 2, false);
 }
 


peterharperuk:
Did you test this change?

kilograham (approved):

akhodeir:
I just tried it.
it didn't work !!! and I don't know what is the reason. 
I found that for SPI the SIGNAL_PATH_RESET shall be reset as well. it seems it is interface dependant. 
I am investigating 

snippet from the spec : 
When using SPI interface, user should use DEVICE_RESET (register 107) as well as
SIGNAL_PATH_RESET (register 104) to ensure the reset is performed properly. The sequence
used should be:
1. Set DEVICE_RESET = 1 (register PWR_MGMT_1)
2. Wait 100ms
3. Set GYRO_RESET = ACCEL_RESET = TEMP_RESET = 1 (register SIGNAL_PATH_RESET)
4. Wait 100ms

akhodeir:
I made a test using this code : 

    uint8_t res[] = {0x6B, 0x80};
    i2c_write_blocking(i2c_default, addr, res, 2, false);
    sleep_ms(200);
    uint8_t buf[] = {0x6B, 0x00};
    i2c_write_blocking(i2c_default, addr, buf, 2, false);
    sleep_ms(200);

res will be used to rest the MPU then wait 200 ms then put the reset bit again to zero.

if the reset bit was not set back to zero the MPU will not work.

The above solution is tested and verfied.

kilograham:
tsk; see #352 ... I assumed you tested your change and it worked, but what you say in the above comment is not the code in the commit. can you please clarify, and submit a new PR

kilograham:
and explain what happens if `0x80` is not set first

vexown:
Hi, just wanted to add an explanation to this solution. 

0x80 is the correct value that should be written to the PWR_MGMT_1 (0x6B) register to cause a reset of all internal registers of the MPU6050 to their default values. 
The reason the value "0x00" needs to be written after the reset is that the reset value of PWR_MGMT_1 is 0x40 (sleep mode) - this causes the device to go into sleep mode by default after a reset. So this is simply waking up the device.

Also if you really care about timing, you could reduce the delays quite a bit, I've found even a 1us after reset is enough. I suppose this doesn't matter much for most applications but just wanted to mention it.

Below is my slightly modified version of the code:

```
static void mpu6050_reset() 
{
	size_t length;

	/* Register: PWR_MGMT_1 (0x6B), Value: 0x80, Action: Reset all internal registers (of MPU6050) to default values */
    uint8_t outputData_Reset[] = {0x6B, 0x80};
	length = sizeof(outputData_Reset);
	i2c_write_blocking(i2c_default, MPU6050_I2C_ADDRESS, outputData_Reset, length, false);
	sleep_us(1); /* Give the slave device some time to perform the reset */
	
	/* Register: PWR_MGMT_1 (0x6B), Value: 0x00, Action: Take MPU6050 out of sleep mode (which is the default state after reset) */
	uint8_t outputData_WakeUp[] = {0x6B, 0x00};
	length = sizeof(outputData_WakeUp);
	i2c_write_blocking(i2c_default, MPU6050_I2C_ADDRESS, outputData_WakeUp, length, false);
}
```

civcode:
Hi @kilograham, I can confirm the explanation of @vexown. See MPU-6050 datasheet below.

The example code in the current version results in the device only transmitting 0x00 for each acceleration, gyro and the temperature value because it's going right to SLEEP state. 

Actually, just writing 0x00 to register 0x6b results in a minimal working example because the device goes to SLEEP state on power-up as the default.

```
static void mpu6050_reset() {
    uint8_t buf[] = {0x6B, 0x00};
    i2c_write_blocking(i2c_default, mpu6050_addr, buf, 2, false);
}
```

BR, 
civ

2023-04-06_13h18_09

vacoff:
It is not merged yet I guess? Problem is still occuring

rafaelcorsi:
i can confirm that 0x80 wont work, and 0x00 works!

akhodeir:
Did you try this code:

https://github.com/akhodeir/pico-examples/pull/1/files

PR #320, opened 2023-02-02
Add i2c slave example

peterharperuk:
from https://github.com/vmilea/pico_i2c_slave

To test you need to wire pin GP4 to GP6 (SDA), and pin GP5 to GP7 (SCL)


peterharperuk:
Depends on https://github.com/raspberrypi/pico-sdk/pull/1205

andygpz11 (changes requested):
This looks good to me and works (when, as noted, it is used along with the new pico_i2c_slave library from PR #1205).

I have one minor change request:  Please can you change the licensing for BSD-3-Clause and also add Raspberry Pi as an author (like kilograham did in 1205)

For the record, we have an email from Valentin which includes "Feel free to relicense under BSD-3-Clause should you prefer that instead." which I know kilograham wanted/wants.

Thanks.

andygpz11 (approved):
LGTM

PR #326, opened 2023-02-06
basic example code for Issue/324 on reading VSYS/VBUS while using wifi on Pico_W

pspeybro:
This is a basic example on how to read VSYS and VBUS using ADC3 and WL_GPIO2 which share pins with wifi, so extra measures need to be taken to make sure wifi keeps working.
See also discussion on https://github.com/raspberrypi/pico-sdk/issues/1222

There are probably some minor updates needed to align it with the other examples (for example put the wifi SSID and password and host ip address in Cmake file), but putting it up here so other people can find it as well.

It seems like I pulled in a few other commits, possibly because I created the branch from the master.

### DIFF
--- a/pico_w/wifi_vsys/wifi_vsys.cpp
+++ b/pico_w/wifi_vsys/wifi_vsys.cpp
@@ -1,0 +1,162 @@
+#include <stdio.h>
+#include "pico/stdlib.h"
+#include "pico/cyw43_arch.h"
+#include "hardware/gpio.h"
+#include "hardware/adc.h"           //to read ADC for soil moisture sensor, connected to pin 36 (3.3V), pin 33 (GND/AGND), pin 34 (GPIO28/ADC2)
+#include "lwip/apps/http_client.h"  //http client
+#include "pico/cyw43_arch/arch_threadsafe_background.h"  //to get access to void cyw43_thread_enter(void);  and void cyw43_thread_exit(void);
+
+#ifndef CYW43_WL_GPIO2
+#define CYW43_WL_GPIO2 2  //this is not defined in pico_w.h
+#endif
+
+const float conversion_factor = 3.3f / (1 << 12); //ADC_VREF=3.3, so this is the ADC resolution. 
+const char ssid[] = "myWifiNetwork";
+const char pass[] = "myWifiPassword";
+const uint32_t auth = CYW43_AUTH_WPA2_MIXED_PSK;
+
+//Http client variables
+bool httpBusy = false;
+char httpBuff[1000]; //if not big enough to read the result or headers, pico will hang.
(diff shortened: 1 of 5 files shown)


kilograham (changes requested):
needs to rebase onto latest pico-examples develop (you seem to have picked up ws2812 changes somehow), and use latest pico-sdk develop (because you will want to include `pico/cyw43_arch.h` instead of anything more specific)

Also please look at the latest updates to pico_w examples in terms of how polling is done in examples

pico_w/wifi_vsys/wifi_vsys.cpp
```diff
@@ -0,0 +1,162 @@
...
+                apicall,    //api call
+                &settings,  //handlers for result/headers
+                body,       //handler for body
+                NULL,
+                NULL
+            );  
+            printf("[1] http status %d \n", err);
+        }
+        else{
+            printf("[1] http client is busy...\n");
```
lurch, line 140: What are all these `[1]` supposed to indicate?

pico_w/wifi_vsys/wifi_vsys.cpp
```diff
@@ -0,0 +1,162 @@
+#include <stdio.h>
+#include "pico/stdlib.h"
+#include "pico/cyw43_arch.h"
+#include "hardware/gpio.h"
+#include "hardware/adc.h"           //to read ADC for soil moisture sensor, connected to pin 36 (3.3V), pin 33 (GND/AGND), pin 34 (GPIO28/ADC2)
```
lurch, line 5: soil moisture sensor - huh?

pico_w/wifi_vsys/wifi_vsys.cpp
pspeybro: ah, some leftover from debugging, totally irrelevant

pico_w/wifi_vsys/wifi_vsys.cpp
pspeybro: one of these: https://arduino-tutorials.net/tutorial/capacitive-soil-moisture-sensor-arduino
I left the sensor reading in there to make the example at least a little more interesting.

pspeybro:
> 

I will be quite busy the following days, so if you have time to clean it up, be my guest. I mainly already wanted to share my solution so that other people can benefit from it.

pico_w/wifi_vsys/wifi_vsys.cpp
pspeybro: leftover, as in, I was running the http request on core1 initially, so it was indicating from which core the printf was coming...

kilograham:
is this superseded by #331 ?

PR #327, opened 2023-02-09
Add example which uses the in-built 'DMA sniff' capability to check a…

andygpz11:
… CRC32


dma/sniff_crc/sniff_crc.c
```diff
@@ -0,0 +1,88 @@
...
+
+    printf("\n");
+    printf("\n");
+
+    // cacluate and append the crc
+    crc_res = soft_crc32_block(CRC32_INIT, src, DATA_TO_CHECK_LEN);
+    *((uint32_t *)&src[DATA_TO_CHECK_LEN]) = crc_res;
+
+    printf("Buffer to DMA: ");
+    for (int i = 0; i < 12; i++) {
```
peterharperuk, line 49: magic number 12. Not obvious why 12 is used here either.

dma/sniff_crc/sniff_crc.c
```diff
@@ -0,0 +1,88 @@
...
+        true           // Start immediately.
+    );
+
+    // We could choose to go and do something else whilst the DMA is doing its
+    // thing. In this case the processor has nothing else to do, so we just
+    // wait for the DMA to finish.
+    dma_channel_wait_for_finish_blocking(chan);
+
+    uint32_t sniffed_crc = dma_sniffer_get_data_accumulator();
+    printf("Completed DMA copy of %d byte buffer, sniffed crc: 0x%08lx\n", TOTAL_LEN, sniffed_crc);
```
peterharperuk, line 87: Maybe print success/fail? And or panic on failure rather than just printf?

peterharperuk: Some comments, but looks useful to me. You need to add to the readme as well.

peterharperuk (changes requested):

dma/sniff_crc/sniff_crc.c
andygpz11: Thanks, that crept through.  It will be TOTAL_LEN

peterharperuk:
You need to add to the readme as well.

peterharperuk:
Looks good to me. Lets see if Graham wants it.

kilograham (commented):
1. Seems like it might make more sense to set the destination increment to 0 unless this is intended to mimic some already happening copy operation
2. Does this calculate the common CRC32 or some variant

andygpz11:
> Seems like it might make more sense to set the destination increment to 0 unless this is intended to
> mimic some already happening copy operation.

Ah - ok.  The code is a "copy 'n change" of hello_dma - hence the copy.
Agreed, If a copy _less_ version would work, that would be better.

> Does this calculate the common CRC32 or some variant

It's the bit reversed version of one of the standard polynomials.
I'm not sure why the non-reversed version did not work for me (I thought I had everything in step).
I'm not saying it can't but it needs more fiddling I think.

dma/sniff_crc/sniff_crc.c
```diff
@@ -0,0 +1,98 @@
...
+    dma_channel_wait_for_finish_blocking(chan);
+
+    uint32_t sniffed_crc = dma_sniffer_get_data_accumulator();
+    printf("Completed DMA copy of %d byte buffer, DMA sniff accumulator value: 0x%lx\n", TOTAL_LEN, sniffed_crc);
+
+    if (0ul == sniffed_crc) {
+        printf("CCRC32 check is good\n");
+    }
+    else {
+        printf("ERROR - CCRC32 check FAILED!\n");
```
lurch, line 96: Is "CCRC32" a typo which should be "CRC32" ?

dma/sniff_crc/sniff_crc.c
```diff
@@ -0,0 +1,98 @@
...
+
+    // We could choose to go and do something else whilst the DMA is doing its
+    // thing. In this case the processor has nothing else to do, so we just
+    // wait for the DMA to finish.
+    dma_channel_wait_for_finish_blocking(chan);
+
+    uint32_t sniffed_crc = dma_sniffer_get_data_accumulator();
+    printf("Completed DMA copy of %d byte buffer, DMA sniff accumulator value: 0x%lx\n", TOTAL_LEN, sniffed_crc);
+
+    if (0ul == sniffed_crc) {
```
lurch, line 92: I know very little about "DMA sniffing" but comparing the result against 0 looks a bit fishy? :shrug:

andygpz11: Good question.  I'll reword this.
You probably know:  If you calculate a CRC32 value over data, and append that "result" the data and run the CRC32 over the expanded buffer [i.e data and the extra 4 bytes], a 0 result means the buffer checks as good,

lurch: No, I didn't - but now I do! :grinning:

andygpz11: Ah, ok.  That's the beauty of CRC algos (like this).  When you detect "end of packet" just just need to read the CRC register [or accumulated value in s/w] - if it is zero you are good. 

Having re-read the code I'm not sure how I can sensibly improve it.  The register is called `sniff_data ` since it operates for both 16 and 32 bit CRC algos.

What do you suggest please?

lurch: Perhaps it doesn't need any code-changes, but just an explanatory comment like you've got here?

lurch:
You seem to have picked up unintentional changes? You probably want to rebase this branch onto the latest `develop`.

kilograham (approved):

peterharperuk (approved):
Looks good to me

PR #331, opened 2023-02-13
Add an example to read VBUS and VSYS

### LINKED ISSUE #324: Write an example to demonstrate how to read vbus and vsys on Pico_W
As these GPIOs are shared with wifi it's not trivial or obvious, see https://github.com/raspberrypi/pico-sdk/issues/1222

peterharperuk:
The process is different on Pico and Pico W so demonstrate how to do it.

Fixes #324


adc/read_vsys/power_status.c
```diff
@@ -0,0 +1,48 @@
...
+#include "power_status.h"
+
+#if CYW43_USES_VSYS_GPIO_PIN
+#include "pico/cyw43_arch.h"
+#endif
+
+// Get power status
+// Must have called adc_init
+// On Pico W must also have called cyw43_arch_init
+void power_status(bool *mains_powered, float *voltage_result) {
```
lurch, line 18: To me "mains" implies 240V, which we're **definitely** not powering Pico directly from! :joy: 
So perhaps `usb_powered` is a better name than `mains_powered`?

adc/read_vsys/power_status.c
```diff
@@ -0,0 +1,48 @@
...
+#if defined CYW43_WL_GPIO_VBUS_PIN
+    *mains_powered = cyw43_arch_gpio_get(CYW43_WL_GPIO_VBUS_PIN);
+#elif defined PICO_VBUS_GPIO_PIN
+    gpio_set_function(PICO_VBUS_GPIO_PIN, GPIO_FUNC_SIO);
```
lurch, line 22: If this function is changing the GPIO_FUNC, should it also restore it before exiting?

adc/read_vsys/power_status.c
peterharperuk: Yes, I didn't really like it. I don't like "usb powered" either as you don't have to be powered by usb. It would probably be better to change it to "battery_powered".

adc/read_vsys/power_status.c
peterharperuk: I think anything using this pin should change the function before using it. The driver now does this.

adc/read_vsys/power_status.c
lurch: > as you don't have to be powered by usb

Although it probably is 90% of the time? :shrug:

peterharperuk: if you're using a power supply you'll be using usb 0% of the time

lurch:
Hmmm, just occurred to me that maybe this example should be using the binary_info stuff too?

adc/read_vsys/power_status.h
```diff
@@ -0,0 +1,33 @@
...
+ * \brief Get batttery status
+ *
+ * Returns whether battery powered
+ * \note On Pico W must have called cyw43_arch_init
+ *
+ * \param battery_powered True if powered by battery, False if powered by USB or another means
+ * \return Zero if the battery status can be determined, an error code otherwise \see pico_error_codes
+ */
+
+int power_battery(bool *battery_powered);
```
lurch, line 20: Sorry to nitpick, but `power_battery` feels like a very weird function-name? Perhaps `power_is_battery` or `power_source` or `power_status` or something? :shrug: 
(as always, please take my comments with a large pinch of salt!)

adc/read_vsys/power_status.c
```diff
@@ -0,0 +1,58 @@
...
+    return PICO_ERROR_NO_DATA;
+#endif
+}
+
+int power_voltage(float *voltage_result) {
+#ifndef PICO_VSYS_PIN
+    return PICO_ERROR_NO_DATA;
+#endif
+#ifndef PICO_VSYS_ADC_INPUT
+    #error If PICO_VSYS_PIN is defined PICO_VSYS_ADC_INPUT must also be defined
```
lurch, line 37: This error message _seems_ to imply that this second `#ifndef` ought to be nested inside the previous `#ifndef` ?

lurch:
LGTM (although I've not actually tried testing it).
Just adding a note here that this can't be merged until after https://github.com/raspberrypi/pico-sdk/pull/1252 has.

kilograham (approved):

kilograham (approved):

adc/read_vsys/power_status.c
```diff
@@ -0,0 +1,78 @@
...
+}
+
+int power_voltage(float *voltage_result) {
+#ifndef PICO_VSYS_PIN
+    return PICO_ERROR_NO_DATA;
+#endif
+#if CYW43_USES_VSYS_PIN
+    cyw43_thread_enter();
+    // Make sure cyw43 is awake
+    cyw43_arch_gpio_get(CYW43_WL_GPIO_VBUS_PIN);
```
oddstr13, line 42: `cyw43_arch_gpio_get` communicates with the wifi chip, and also calls to `cyw43_thread_enter` and `cyw43_thread_exit`.

Would that not cause conflicts and or release our lock too soon?

peterharperuk: Are you seeing an issue? You can nest calls to enter and exit so I wouldn't expect this to be a problem.

PR #334, opened 2023-02-18
Add pio onewire example

mjcross:
Add a PIO library for interfacing to 1-Wire devices, with an example for the widely used DS18B20 temperature sensors


mjcross:
@kilograham apologies, but I'm unfamiliar with the workflow: is this awaiting any action by me?

kilograham (approved):

mjcross (commented):
@kilograham re your [e8d977e](https://github.com/raspberrypi/pico-examples/pull/334/commits/e8d977e6010eccc64d283766100dc18cf317fb29): `target_sources()` call in line 2 of `pio/onewire/onewire_library/CMakeLists.txt` is incorrect I think because it is missing a PUBLIC/PRIVATE/INTERFACE qualifier. Perhaps it should be something like this:

`target_sources(onewire_library INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/onewire_library.c)`

Issue #321, opened 2023-02-02
Should the examples default to using stdio USB?


peterharperuk:
Some of the examples only really "work" if you can see the output. But by default we enabled UART which means you have to own a USB adapter.

Maybe we should default to using USB, or have some clearer explicit way for the user to specify their choice? I have a load of aliases to make the switch for me

alias cmaked="cmake -DCMAKE_BUILD_TYPE=Debug"
alias cmakeu="cmake -DPICO_STDIO_UART=0 -DPICO_STDIO_USB=1"
alias cmakedu="cmaked -DPICO_STDIO_UART=0 -DPICO_STDIO_USB=1"
alias cmaken="cmake -G Ninja"
alias cmakedn="cmaken -DCMAKE_BUILD_TYPE=Debug"
alias cmakeun="cmaken -DPICO_STDIO_UART=0 -DPICO_STDIO_USB=1"
alias cmakedun="cmakedn -DPICO_STDIO_UART=0 -DPICO_STDIO_USB=1"

kilograham:
yeah, we need to go thru and try them to make sure they do indeed actually work without deadlocks ;-)

peterharperuk:
I've had no problems so far on develop

lurch:
I _suspect_ that the reason the examples all default to using UART-stdio is that using the hardware UART has much less software-overhead on the RP2040 than using USB-stdio, and so when doing low-level debugging it's much easier to see what's going on when the USB device stack isn't being used on the RP2040 at the same time. But the software running on the RP2040 has now probably matured to a point where that's less of a concern, and using USB-stdio by default is certainly _much_ more user-friendly for those people that don't have a USB-UART adapter. For maximum-compatibility reasons, would there be any reason not to do "-DPICO_STDIO_UART=1 -DPICO_STDIO_USB=1" (wherever possible) rather than "-DPICO_STDIO_UART=0 -DPICO_STDIO_USB=1" ? With https://github.com/raspberrypi/pico-examples/tree/master/hello_world being an obvious exception to the rule!

[...]

PR #338, opened 2023-03-05
A2ds source demo add usb control

ianabc:
This PR makes the [a2ds_source_demo](../blob/master/pico_w/bt/a2dp_source_demo) example a bit more accessible by exposing the UI over a USB/Serial connection.

[btstack/examples/a2ds_source_demo](https://github.com/bluekitchen/btstack/blob/master/example/a2dp_source_demo.c) provides a UI for paring with an audio sink, controlling volume, starting and stopping playback etc. This works straightforwardly on desktop since `stdout/stdin` are generally connected, but not on a RPi pico_w. The modifications in this PR expose that same UI over a USB/serial connection (borrowing code from some of the other examples).

When the application is run via an RPi/w connected by USB, you can connect to the UI with e.g. `screen /dev/tty.usbmodem*`, then hitting `space` on MacOS. Next, put any audio sink (e.g. a pair of Bluetooth headphones) in pairing mode, and type `a` in the UI to scan for devices and connect. Once connected the audio from the example should start playing in the sink device, and you can control playback in the UI.

I think these modifications make the example a bit more useful, but I don't know either Bluetooth, Raspberry PI or this project very well so if they're inappropriate my apologies and please close the PR.

### DIFF
--- a/pico_w/bt/a2dp_source_demo/CMakeLists.txt
+++ b/pico_w/bt/a2dp_source_demo/CMakeLists.txt
@@ -1,1 +1,5 @@
 picow_bt_example(a2dp_source_demo pico_btstack_hxcmod_player pico_btstack_sbc_encoder)
+
+# enable usb output, disable uart output
+pico_enable_stdio_usb(picow_bt_example_a2dp_source_demo_background 1)
+pico_enable_stdio_uart(picow_bt_example_a2dp_source_demo_background 0)


peterharperuk:
Probably a duplicate of https://github.com/raspberrypi/pico-examples/issues/321

ianabc:
Yes, I tried it out with `-DPICO_STDIO_UART=0 -DPICO_STDIO_USB=1` and got the same behaviour. I'll close this in favour of #321

Issue #348, opened 2023-03-08
VSCode Launch.json example configurations for more types of probes


d-adamson:
Propose a launch.json file that supports more types of debug probes which can be selected from the Debug pull-down menu.
![PicoDebugProbes](https://user-images.githubusercontent.com/7835990/223779759-33521ad3-bf87-4c6e-9ee6-bbbf0742cf8a.png)

Attached is an "overly commented" example. :) 
I've tested each probe from a Linux machine (except for the original Raspberry Pi configuration).

[launch.json.txt](https://github.com/raspberrypi/pico-examples/files/10923065/launch.json.txt)

d-adamson:
See pull request #349

PR #349, opened 2023-03-08
launch.json - debug probes ( Issue #348)

d-adamson:
Propose a modified lauch.json that provides the user with choices for a debug probe.

PicoDebugProbes


kilograham (approved):
probably worth updating the readme

PR #350, opened 2023-03-08
Blink gpio10

d-adamson:
Propose an additional blink example that
1) Uses and external led to work with both the pico and pico w
2) Provides a serial out message
3) Shows how to declare a variable used for debugging with "volatile" to avoid the inspected value shows as "optimized out"


lurch:
I'm guessing this PR shouldn't include the JSON file?

d-adamson:
Correct. Sorry, my mistake.

What's the best way to correct this?

aallan:
As far as I know the existing `blink.c` using the on-board LED works on both Pico and Pico W? https://github.com/raspberrypi/pico-examples/blob/master/blink/blink.c

However, if we were going to include something like this — an additional blink example using an external LED — just like the two "hello world" examples (`hello_usb` and `hello_serial`) it should live in a sub-directory of the existing `blink` example, which should then be renamed and moved to its own subdirectory inside the existing `blink/` directory. 

In addition, since it's a getting started example, I think it should have a wiring diagram as it requires external wiring, e.g. see https://github.com/raspberrypi/pico-examples/tree/master/i2c/bmp280_i2c where there is a `README.md` file with an included wiring example.

That said, a blink example needs to be the simplest thing possible, parsable by folks that don't even know C, and are just familiar with other languages, and [this](https://github.com/raspberrypi/pico-examples/pull/350/files#diff-f91eae1432ceff6ed79b31a746046ff5cf0be0421015e0b01b020c0c82685a86) is far longer and much more complex than the [existing blink example](https://github.com/raspberrypi/pico-examples/blob/master/blink/blink.c). I don't believe this is suitable for folks that'll be looking to a `blink` example as their very first experience of the SDK.

peterharperuk:
That blink example doesn't work on Pico W. There's a separate one for Pico W https://github.com/raspberrypi/pico-examples/blob/master/pico_w/wifi/blink/picow_blink.c

d-adamson:
In the spirit of keeping it simple, someone just starting may purchase the pico w as their first board (I did because I want to do an IOT project). 
* It is not immediately obvious that the current blink example does not work on the pico w. 
* The pico w blink example is pretty well buried in the examples directory tree. 
* picow_blink it is a slightly more complex program because the onboard led is tied to the wifi module rather than being a standard gpio => not a good teaching example.

For those reasons,  the current blink example should be updated to be picow w friendly.

No problem leaving out the serial coms and the "volatile" declaration for simplicity.

urbasus:
For the record I fell into this trap getting started, flashing the pico examples on a pico w, scratching my head and flailing about until I realized my mistake. A few more breadcrumbs and safety nets wouldn't hurt the experience.

peterharperuk:
Is it going to be any more obvious when you need a led and a resistor to make it work?

CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ pico_sdk_init()
 include(example_auto_set_url.cmake)
 # Add blink example
 add_subdirectory(blink)
+add_subdirectory(blink-10)
```
lurch, line 24: `blink_external_led` is probably a more useful name than `blink-10` ?

PR #353, opened 2023-03-17
Improve on github workflow uses of make

urbasus:
- Fix use of jobs count for throttling parallelism.
  - Would appear the prior variable expanded to nothing. 
  - Option -j should be followed by jobs count without space in between, e.g. -j24.
- Synchronize output per target (-O).
  - Makes reading logs easier when command and output are adjacent.
  - "Deeper" synchronization may be preferable.
- Omit built-in variables (-R) and rules (-r).
  - Saves a few milliseconds starting up make, in my experience.
  - CMake defines all rules and variables from scratch for what I know, superceding any built-ins.

### DIFF
--- a/.github/workflows/multi-gcc.yml
+++ b/.github/workflows/multi-gcc.yml
@@ -40,146 +40,142 @@ jobs:
       working-directory: ${{github.workspace}}/pico-examples
       run:  cmake -E make_directory ${{github.workspace}}/pico-examples/build
 
-    - name: Get core count
-      id: core_count
-      run : cat /proc/cpuinfo  | grep processor | wc -l
-
     - name: GCC 6.2.1 Debug
       if: always()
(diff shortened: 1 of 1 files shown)


urbasus:
I suppose I should remove the unused core_count step.

urbasus:
The workflow looks the same as for pico-sdk. Should probably keep them in sync. Will open issue and PR on SDK should this PR be accepted.

urbasus:
> I suppose I should remove the unused core_count step.

Done.

liamfraser:
Thanks. This workflow is generated by a script so I will update the script and do the same for pico-sdk

liamfraser:
Thanks for this PR. It isn't quite right for our systems as we have a different version of make. I will raise a new PR with the correct make options and credit you.

PR #368, opened 2023-04-18
Fix standalone bt client

peterharperuk:


pico_w/bt/standalone/client.c
```diff
@@ -5,6 +5,7 @@
  */
 
 #include <stdio.h>
+#include <hci_dump_embedded_stdout.h>
```
kilograham, line 8: did you mean to include this? if so, perhaps " for consitency

peterharperuk: No. Now removed

PR #372, opened 2023-05-01
Add new example pio/7_segment

mjcross:
1. Adds a new example showing how to control a multiplexed 7-segment LED display with the PIO.
2. Includes my previous fix for `pio/onewire/onewire_library/CMakeLists.txt`, as I inadvertently deleted my previous fork (apologies if this is unnecessary and/or duplicative)


lurch:
> Adds a new example showing how to control a multiplexed 7-segment LED display with the PIO.

I've just had a quick read through the code; IMHO this looks quite elegant, and is certainly a good demo of why using PIO is better than the bit-bashing that'd be needed on other MCUs :sunglasses:

PR #365, opened 2023-04-14
fix pio/onewire_library CMakeLists.txt

mjcross:
Add missing INTERFACE keyword and path to `target_sources()` in pio/onewire/onewire_library/CMakeLists.txt.
Add the `onewire` subdirectory to pio/CMakeLists.

PR #374, opened 2023-05-05
fix build for pio/onewire/onewire_library

mjcross:
In `pio/onewire/onewire_library/CMakeLists.txt` I think _target_sources()_ requires a keyword (INTERFACE | PUBLIC | PRIVATE).
Also the onewire subdirectory seems to be missing from `pio/CMakeLists.txt`


lurch:
Replaces #365

mjcross:
bump

kilograham (approved):

PR #375, opened 2023-05-05
add pio example for 7_segment display

mjcross:
Add an example showing how to use the PIO to control a multiplexed 7-segment display.


lurch:
Replaces #372

pio/7_segment/pio_7_segment_library/7_segment_lib.pio
```diff
@@ -0,0 +1,50 @@
...
+                                          uint segment_pinbase, uint digit_pinbase) {
+    // create new sm config
+    pio_sm_config config = seven_segment_program_get_default_config (offset);
+
+    // configure the common cathodes and segment pin groups
+    sm_config_set_out_pins (&config, segment_pinbase, 8);
+    sm_config_set_sideset_pins (&config, digit_pinbase);
+
+    // configure the clock divider
+    float div = clock_get_hz (clk_sys) * 10e-3;
```
pmarques-dev, line 41: I may be missing something, but it seems that if clk_sys is 125MHz (for instance) then div will become 125000. This is higher than 65536, which is the maximum supported by the clock divider. I think this needs a check and if the clock is too high, the divisor would just be set to 65536. This will produce a refresh rate of around 2kHz on the display, which should still be ok.

mjcross: Good shout - fixed with https://github.com/mjcross/pico-examples/commit/25ac8f45374d1bacec096dae20ec09623164a2f7

mjcross:
HI guys - is there some reason this one isn't progressing?

mjcross:
Bump

peterharperuk:
No. It'll be considered for the next sdk release.

mjcross:
Thanks - I think it's a useful example because it shows how to keep outputting the same pattern until something arrives on the FIFO

lurch:
Weird - GitHub says "[mjcross](https://github.com/mjcross) closed this by deleting the head repository" and yet the repository still seems to be there? https://github.com/mjcross/pico-examples/

mjcross:
@lurch I think it's probably because I deleted the original repo and created a new one with the same name: I'd forgotten this old PR was still open
If for some reason you still want the example let me know and I'll repost it :-)

lurch:
I still think that this is a useful example to have.
Would using a Darlington array simplify the wiring?

mjcross:
Good idea, I'll have a look.

mjcross:
On reflection the example should probably be about using the PIO rather than wiring up the display.
How about I just put a square box on the diagram between the GPIOs and the 7-Seg labelled "High current driver (if required)" ?

lurch:
I'm afraid that we expect all the examples to be easily followable (which is the whole reason that we ask for Fritzing diagrams), and an "abstract box" in the wiring diagram would make this example impossible to follow for many people.
Feel free to stick to using individual transistors, if you feel that using a Darlington array would overcomplicate things.

lurch:
Although having just had another look at your Fritzing diagram, IMHO it'd be better if you used the normal convention of using the bottom blue rail for 0V, rather than using the bottom red rail :slightly_smiling_face:

mjcross:
Fair enough!
Unfortunately I don't have Fritzing any more and as it's now a paid download I'm afraid that's unlikely to change...
I've found the source so I'll see if GitHub will let me restore the original repo...

mjcross:
Sorry Andrew - I can't work out for the life of me how to persuade GitHub to let me restore the original Repo so I'll open a new PR

PR #381, opened 2023-05-16
Add an example for how to enable TLS verification.

### LINKED ISSUE #337: How to use TLS with certificate check
Instead of using `tlsConfig = altcp_tls_create_config_client(NULL, 0);`. I tried "thousand" ways and no positive result.

With NULL and zero values it works great, but I think it is not a proper way to connect our pico to our server.

https://github.com/raspberrypi/pico-sdk/issues/1299

peterharperuk:
TLS should really be used with verification enabled, as otherwise you can still suffer from a "man in the middle" attack. Add an example that demonstrates how to do this.

Fixes #337


kilograham (changes requested):
probably worth adding a tiny extra bit to the README entry

peterharperuk:
Added an entry for the new example to the readme

kilograham (approved):

PR #387, opened 2023-05-22
Call disable function for STA/AP modes.

### LINKED ISSUE #386: Add some tests for cyw43_arch_disable_ap_mode cyw43_arch_disable_sta_mode
See https://github.com/raspberrypi/pico-sdk/issues/1371

peterharperuk:
If the user presses "d" wifi is disconnected and the example exits. Tests new "disable" functions.

Fixes #386


kilograham (approved):
Ha, i was confused what `state->complete` was doing before, but it wasn't!

PR #388, opened 2023-05-26
Comments about btstack_run_loop_execute

peterharperuk:
Add some comments about how btstack_run_loop_execute does nothing in threadsafe backrgound mode.

### DIFF
--- a/pico_w/bt/standalone/client.c
+++ b/pico_w/bt/standalone/client.c
@@ -268,13 +268,16 @@ int main() {
     // turn on!
     hci_power_control(HCI_POWER_ON);
 
-    // We're using threadsafe background where poll does nothing and work is performed in a low priority IRQ
-    // So btstack_run_loop_execute effectively does nothing and you're free to just loop and do your own stuff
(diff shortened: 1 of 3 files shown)


pico_w/bt/standalone/client.c
```diff
@@ -268,6 +268,15 @@ int main() {
     // turn on!
     hci_power_control(HCI_POWER_ON);
 
+    // We're using threadsafe background where poll does nothing and work is performed in a low priority IRQ
+    // So btstack_run_loop_execute effectively does nothing and you're free to just loop and do your own stuff
+    // You would have to call btstack_run_loop_execute if you were using pico_cyw43_arch_poll
+#if 1
```
lurch, line 274: Apologies for the stupid question, but why does this PR add `#if 1` in `client.c` but `#if 0` in `server.c` ?

lurch: Fair enough :slightly_smiling_face:

kilograham: yeah, i also found it a bit odd(confusing); i'm planning to make some comment tweaks when i merge

peterharperuk:
The changes look ok to me

PR #393, opened 2023-06-07
Fix PICO_VBUS_PIN

peterharperuk:
Inexplicably using the wrongly named constant. Should be PICO_VBUS_PIN.
This is used for Pico, Pico W uses CYW43_WL_GPIO_VBUS_PIN

### DIFF
--- a/adc/read_vsys/power_status.c
+++ b/adc/read_vsys/power_status.c
@@ -49,13 +49,11 @@ int power_voltage(float *voltage_result) {
     adc_fifo_setup(true, false, 0, false, false);
     adc_run(true);
 
-#if CYW43_USES_VSYS_PIN
-    // We seem to read low values from cyw43 sometimes - this seems to fix it
+    // We seem to read low values initially - this seems to fix it
     int ignore_count = PICO_POWER_SAMPLE_COUNT;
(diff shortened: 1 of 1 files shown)


peterharperuk:
Note: I don't currently have a PICO on me, so haven't tested this yet.

peterharperuk:
Tested this on Pico. VBUS is correct with this. I'm seeing instability in VSYS but I guess that's a different problem.

adc/read_vsys/power_status.c
```diff
@@ -49,13 +49,11 @@ int power_voltage(float *voltage_result) {
     adc_fifo_setup(true, false, 0, false, false);
     adc_run(true);
 
-#if CYW43_USES_VSYS_PIN
-    // We seem to read low values from cyw43 sometimes - this seems to fix it
+    // We seem to read low values initially - this seems to fix it
     int ignore_count = PICO_POWER_SAMPLE_COUNT;
```
lurch, line 53: I wonder if it might be worth making a separate `PICO_POWER_IGNORE_SAMPLE_COUNT` define for this, given that it _always_ seems to be needed?

kilograham (approved):

Issue #380, opened 2023-05-13
Can't use GDB in VS Code


SlazyHW:
Hello everyone! I followed the guide [(on https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf)](url) to start debugging my Pico with VS Code. Everytime I start trying to debug the code with the green debug button in the upper left corner, I get following error in the Output Window:

Reading symbols from objdump-multiarch --syms -C -h -w /home/jan/pico/pico-examples/build/hello_world/usb/hello_usb.elf
Reading symbols from nm-multiarch --defined-only -S -l -C -p /home/jan/pico/pico-examples/build/hello_world/usb/hello_usb.elf
Launching GDB: gdb-multiarch -q --interpreter=mi2
    IMPORTANT: Set "showDevDebugOutput": "raw" in "launch.json" to see verbose GDB transactions here. Very helpful to debug issues or report problems
Launching gdb-server: openocd -c "gdb_port 50000" -c "tcl_port 50001" -c "telnet_port 50002" -s /home/jan/pico/pico-examples -f /home/jan/.vscode/extensions/marus25.cortex-debug-1.10.0/support/openocd-helpers.tcl -f interface/cmsis-dap.cfg -f target/rp2040.cfg
    Please check TERMINAL tab (gdb-server) for output from openocd
Finished reading symbols from objdump: Time: 25 ms
Finished reading symbols from nm: Time: 37 ms
OpenOCD: GDB Server Quit Unexpectedly. See gdb-server output in TERMINAL tab for more details.

I am using the Raspberry Pi Debug Probe, so I looked at this tutorial: [https://www.raspberrypi.com/documentation/microcontrollers/debug-probe.html](url). 
I changed the interface from raspberrypi-swd to cmsis-dap, set the right path to my pico sdk folder. 

my launch.json looks like this:

"version": "0.2.0",
    "configurations": [
        {
            "cwd": "${workspaceRoot}",
            "executable": "./bin/executable.elf",
            "name": "Debug with OpenOCD",
            "request": "launch",
            "type": "cortex-debug",
            "servertype": "openocd",
            "configFiles": [],
            "searchDir": [],
            "runToEntryPoint": "main",
            "showDevDebugOutput": "none"
        },
        {
            "name": "Pico Debug",
            "cwd": "${workspaceRoot}",
            "executable": "${command:cmake.launchTargetPath}",
            "request": "launch",
            "type": "cortex-debug",
            "servertype": "openocd",
[...]

PR #401, opened 2023-06-28
Adding launch.json for the debug probe

noutram:
Currently, none of the candidate launch.json files in the ide/vscode folder work with the (relatively new) official Debug Probe. Although similar to the pico-probe, there are some small differences. It took a considerable time to work out the changes needed to get it to work.
 
The proposal here is to add an additional file (and reference in the readme) with these changes. It is also labelled launch-debug-probe.json to match the product name.
The README is also updated (although I am not sure about including the URL of the product page - is that a stable URL?)

### DIFF
--- a/ide/vscode/launch-debug-probe.json
+++ b/ide/vscode/launch-debug-probe.json
@@ -1,0 +1,28 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Pico Debug",
+            "cwd": "${workspaceRoot}",
+            "executable": "${command:cmake.launchTargetPath}",
+            "request": "launch",
+            "type": "cortex-debug",
+            "servertype": "openocd",
+            // This may need to be "arm-none-eabi-gdb" for some previous builds
+            "gdbPath" : "gdb-multiarch",
+            "device": "RP2040",
+            "configFiles": [
+                // This may need to be "interface/picoprobe.cfg" for some previous builds
+                "interface/cmsis-dap.cfg",
+                "target/rp2040.cfg"
+            ],
+            "svdFile": "${env:PICO_SDK_PATH}/src/rp2040/hardware_regs/rp2040.svd",
+            "runToEntryPoint": "main",
+            // Work around for stopping at main on restart
+            "postRestartCommands": [
+                "break main",
+                "continue"
+            ]
+        }
+    ]
+}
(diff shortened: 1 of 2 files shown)


lurch:
See also #349 and #380 and https://github.com/raspberrypi/picoprobe/issues/49#issuecomment-1331748165 
ping @P33M

lurch:
> The README is also updated (although I am not sure about including the URL of the product page - is that a stable URL?)

Just to be pedantic, https://www.raspberrypi.com/documentation/microcontrollers/debug-probe.html is actually the _documentation_ URL; the product URL is https://www.raspberrypi.com/products/debug-probe/ :wink: 
But yes, both of those URLs are stable.

noutram:
Slight aside, but I would be happy to contribute small PRs to the documentation - or is that internal only?

lurch:
> Slight aside, I would be happy to contribute small PRs to the documentation - or is that internal only?

The documentation at https://www.raspberrypi.com/documentation/ is open for PRs at https://github.com/raspberrypi/documentation/tree/develop

The PDF documentation at https://datasheets.raspberrypi.com/ isn't open, but you can submit suggestions at https://github.com/raspberrypi/pico-feedback/issues

ide/vscode/launch-debug-probe.json
```diff
@@ -0,0 +1,31 @@
...
+            "name": "Pico Debug Probe",
+            "cwd": "${workspaceRoot}",
+            "executable": "${command:cmake.launchTargetPath}",
+            "request": "launch",
+            "type": "cortex-debug",
+            "servertype": "openocd",
+            // This may need to be "arm-none-eabi-gdb" for some previous builds
+            "gdbPath" : "gdb-multiarch",
+            "serverArgs": [
+                "-c", "adapter speed 5000"
```
vidadonizete, line 14: This way https://github.com/vidadonizete/pivro/blob/main/environments/pico/.vscode/launch.json#L20C39-L20C39 it works too

vidadonizete:
Also, would be great to document about udev rules: https://github.com/raspberrypi/openocd/blob/rp2040-v0.12.0/contrib/60-openocd.rules when using openocd with cortex-debug extension.
You need to move udev rules and add your user into group: sudo useradd -G plugdev $(whoami).

PR #405, opened 2023-07-04
add high resolution quadrature encoder example

pmarques-dev:
This code solves a problem that anyone doing robotics or automation faces on a regular basis: estimate the speed of a motor using a quadrature encoder.

It uses the step counts **and** the timing of the step transitions to get the best possible speed estimate, which may be a fractional step count per sample (see README.md for more details).


peterharperuk:
Seems impressive. Am I right in thinking that you get revolutions by multiplying speed by Counts Per Revolution? Presumably this could be tested with something like this? https://shop.pimoroni.com/products/magnetic-encoder-pair-kit-for-micro-metal-gearmotors-12-cpr-2-7-18v-hpcb-compatible?variant=30271550914643

pmarques-dev:
The output is "substeps per second" where a substep is 1/64 of an hardware encoder step. So if  you know how many steps your encoder has per rotation, it's easy to get the speed in RPM:  rpm = ((<substeps_per_second>/64) / <encoder steps per rotation> * 64) * 60. This should be done with floating point numbers or you have to be careful not to have numerical issues.

Sometimes you don't really need the speed in rpm. For instance, if you have a wheel connected to that motor and you want the speed in meters per second. In that case, you can take your robot, move it one meter and read the position in substeps at the start and end of that meter. Subtract the two and you have the <substeps per meter> constant that will allow you to convert from <subteps per second> into <meters per second> directly.

And yes, it can be tested with something like that.

pmarques-dev:
This has been sitting here since July. Anything I can do to help review this pull request?

peterharperuk:
I want to test it here rather than just merge it

peterharperuk:
Finally. Looks good to me. I would like to improve the output slightly if that's ok.

peterharperuk:
Needs to be changed to develop branch

pmarques-dev:
sorry about that. How do you want to proceed: do you want to just close this MR and I'll submit a new one using the correct branch? or something else?

PR #412, opened 2023-08-16
update tcp_server to allow for usb CMakeLists.txt

horner:
without picow_tcpip_server_poll the printf's like this: https://github.com/raspberrypi/pico-examples/blob/eca13acf57916a0bd5961028314006983894fc84/pico_w/wifi/tcp_server/picow_tcp_server.c#L258
do not show anywhere. I burned hours wondering if the code was running.

Should we think about making all of the examples turn on usb stdio by default? Since you have to used the USB to install the app it sure would make it easy to see debug every example was set to show stdio to the usb serial.


lurch:
See section 4.1 of https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf

kilograham:
duplicate of #321

Issue #197, opened 2022-02-08
Pulse Position Modulation (PPM)


adaptivercsailing:
An invitation to use RPi Pico to develop RC (remote radio frequency control) pulse position modulation (PPM) waveforms. The RPi Zero and RPi Pico appear to be ideal candidates to create a new approach for providing alternate solutions to the **Adaptive RC Sailer Project** https://github.com/adaptivercsailing/adaptivercsailing. Request help duplicating the effects of the **Arduino Pulse Width Modulation Utility** https://github.com/ElectricRCAircraftGuy/eRCaGuy_PPM_Writer for use with RPi Zero and Pico.

elmot:
Hello @adaptivercsailing, would you like test my demo?
I have done that but I have no real hardware to test.
https://github.com/elmot/pico-examples/tree/master/pio/ppm

adaptivercsailing:
elmot,
I just saw your reply/offer from August 2023.
I'll start testing your code on Monday 10-09-2023.
Please correspond via dkender@wright.edu
Thank you - I am looking forward to working with you.
Dave K

PR #414, opened 2023-08-20
pio_ppm - RC PPM protocol example works on GPIO 3

elmot:
An example of RC PPM protocol encoder using PIO
Resolves raspberrypi/pico-examples#197


pio/ppm/ppm.pio
```diff
@@ -0,0 +1,79 @@
...
+void ppm_handler() {
+    while (!pio_sm_is_tx_fifo_full(ppm_data.pio, 0)) {
+        pio_sm_put(ppm_data.pio, 0, ppm_data.ch_values[ppm_data.value_idx]);
+        ppm_data.value_idx = (ppm_data.value_idx + 1) % (PPM_CHANNELS + 1);
+    }
+}
+
+static inline void ppm_program_init(PIO pio, uint pin) {
+    ppm_data.pio = pio == NULL ? pio0 : pio;
+    ppm_data.ch_values[0] = 499 * 0x10000 +  4999;
```
lurch, line 44: _Perhaps_ it'd be useful to have a macro which did
```c
#define BUILD_CH_VALUE(pulse) (((500 - 1) << 16) + (pulse - 1))
```
and then you could replace
```c
ppm_data.ch_values[0] = 499 * 0x10000 +  4999;
```
(and other similar lines) with
```c
ppm_data.ch_values[0] = BUILD_CH_VALUE(5000);
```
?

pio/ppm/ppm.c
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * 8-channel Radio Control PPM example
+ * Output on GPIO3
+ *
+ * 0.5ms   5ms         0.5ms      1-2mc      0.5ms      1-2mc
```
lurch, line 11: Are the `1-2mc` here supposed to be `1-2ms`?
Thanks for incorporating my other suggestion :slightly_smiling_face:

elmot: Thanks for the review and for the suggestion.
Yes, that is the idea of PPM protocol for radio-controlled models. 
See here, for instance: https://sourceforge.net/p/arduinorclib/wiki/PPM%20Signal/

lurch: I wasn't querying the timing, I was only querying the use of `mc` versus `ms` ?

elmot: omg... thanks

elmot: Ok, that was just a documentation glitch. Now everything is fixed.

elmot:
Needs to be fixed after testing with real hardware

pio/ppm/ppm.c
```diff
@@ -8,10 +8,10 @@
...
  * Output on GPIO3
  *
- *  >=5ms       0.5ms    1-2ms       0.5ms     1-2ms         0.5ms    1-2ms       0.5ms
- *  synchro     strobe   ch0 value   strobe    ch1 value     strobe   ch8 value   strobe
- *              _______               _______                _______              _______
- * _______...__|       |_____________|       |___________...|       |____________|       |_
+
+ *  synchro     /  ch0 value 1-2ms  \/  ch0 value 1-2ms  \   /  ch8 value 1-2ms  \
+ *  >=5ms       _______    -500ms     _______                 _______              _______
+ * _______...__| 0.5ms |_____________| 0.5ms |___________|...| 0.5ms |____________| 0.5ms |_
```
lurch, line 14: I don't have hardware to test this on, but is there definitely a 0.5ms pulse after the ch8 value, or does it just go straight into the next 5ms synchro pulse?
Also, I suspect the "-500ms" should probably have been "-500us"?

elmot: My friend tested that with a real flight controller, ch8 worked.
There is a pulse after ch8, it's drawn. 
"500ms" should not be here at all.
Excuse me for being so inaccurate.
I have changed the comment once again, hopefully it's better now

pio/ppm/ppm.pio
```diff
@@ -0,0 +1,83 @@
...
+        ppm_data.value_idx = (ppm_data.value_idx + 1) % (PPM_CHANNELS + 1);
+    }
+}
+
+// 500 microseconds strobe + gap = value (microseconds)
+#define BUILD_CH_VALUE(value_us) (((500 - 1) << 16) + (value_us - 500 - 1))
+
+static inline void ppm_program_init(PIO pio, uint pin) {
+    ppm_data.pio = pio == NULL ? pio0 : pio;
+    ppm_data.ch_values[0] = BUILD_CH_VALUE(5000);
```
lurch, line 47: Should this now be bumped to 5500? (I think otherwise you'd only get a 4.5ms synchro)

elmot: You are right, yes

lurch:
Oh, this should be targeted at the `develop` branch, rather than the `master` branch.

elmot:
Redirected

elmot:
@lurch any progress here?

jschwob:
Today I tested your code and found an issue. Using the ToolkitRC M6 Smart charger which includes a ppm monitor I found that the output could not be read. Since the device can also generate a PPM output I simply compared the waveforms and determined that that the signal was inverted. (This is common in the RC world. Blame futuba - Sbus is also inverted)
I'm not too good with github so here is the update I made to invert the signal.

.program ppm
.side_set 1

    set pindirs, 1  side 0; Set pin to output
.wrap_target
    out x, 16 side 1
delay0:
    jmp x-- delay0 side 1
    out x, 16 side 0
delay1:
    jmp x-- delay1 side 0
    .wrap

I'm working on a project that will use this code to interface on the trainer port of RC Transmitters. I haven't performed any value tests yet. Will post an update if I find any timing issues which would cause an incorrect value to be returned to the rc transmitter.

lurch:
> determined that that the signal was inverted.

It might be easier to use [gpio_set_outover](https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#ga829129eed5d42f3455e9da339219d814) with GPIO_OVERRIDE_INVERT, than to rewrite the PIO code? :shrug:

jschwob:
That sounds like a great suggestion. I haven't had time to test it yet - my changes simply flipped the output bits. I also need to generate an inverted uart for Sbus so I'll give GPIO_Override_invert a shot and report my findings.

jschwob:
I added this to the init portion:
 gpio_set_outover(pin, GPIO_OVERRIDE_INVERT); //Invert output as rc transmitters expect inverted signal.

worked as expected and my tester read the signal w/o issue. No changes are required to the assembly code. Just need to add this to the init portion. 

Testing: Initial tests show values requested were output +2 high. I have only tested it on one device so far. I don't have an accurate oscilloscope to validate the signal properly however I do have multiple flight controllers and other rc transmitters to read the values to validate the ppm signal with. I'll post more as I continue testing. So far this is working great and +2 seems reasonable all things considered.

elmot:
@jschwob that story with inverted signals looks a bit strange. Existing code works well with Ardupilot flight controller, and I never seen any articles where PPM signal looks inverted. Can you, please, provide the link?

jschwob:
Like I commented here is my testing device. It's a standard hobby charger and signal tester. ToolkitRC M6 Smart charger which includes a ppm monitor

Ardupilot may accept non inverted signals. Not sure. Inverted and non inverted signals are a common issue in the hobby world and many devices are starting to use non inverted signals on see radio master transmitters.

jschwob:
I'm still testing it and will post results when I connect it to a transmitter trainer port on one of my transmitters. For my application I will add a setting to allow the user to choose inverted as default or non inverted.

Timot05:
I got this to work by replacing `PIO_INTR_SM0_TXNFULL_LSB` with `pis_sm0_tx_fifo_not_full`. Apparently `PIO_INTR_SM0_TXNFULL_LSB` is defined in multiple place and it was causing issues. See the PPM frame below.

Will report back later on whether is can be read by the [ExpressLRS](https://www.expresslrs.org) module I'm using.

IMG_8161

Timot05:
I can confirm that this is working fine. Here is the code I'm using: [link](https://github.com/Timot05/PIO_PPM_RP2040).

One note: the library implies that you can use either the `pio0` or `pio1` hardware but the interrupts are hard coded to `pio0`.

PR #419, opened 2023-09-05
Update mpu6050_i2c.c

akhodeir:
The reset process requires to set DEVICE_RESET bit in PWR_MGMT_1 (0x6B) bits to one ===> 0x80. then the we must release the reset bit or the device will remain always in reset mode. 

detailed explanation is here: #319


akhodeir:
If the reset bit is not set again to zero === 0x00 , the MPU6050 will remain in reset state forever and will not function at all.

lurch:
This should be targeted against the `develop` branch, rather than `master`.

akhodeir:
> This should be targeted against the `develop` branch, rather than `master`.

Ok, I moved it to develop branch

lurch:
According to Gihub, this PR now has conflicts :confused:

akhodeir:
triggered a new PR

PR #424, opened 2023-10-03
BTstack enable common features

mringwal:
BTstack uses feature flags when adding new features. As a consequence, the default configuration should enable most of them, e.g. LE Data Length extension or LE Secure Connections should be enabled. This aligns the Pico SDK port with libusb.


peterharperuk:
I notice Micropython doesn't enable ENABLE_LE_DATA_LENGTH_EXTENSION - maybe it should?

mringwal:
> I notice Micropython doesn't enable ENABLE_LE_DATA_LENGTH_EXTENSION - maybe it should?

Well, from my side, it should/could. You could suggest it.

PR #427, opened 2023-10-03
Add a README for bluetooth debugging

### LINKED ISSUE #426: Add a readme for bluetooth examples
See https://github.com/raspberrypi/pico-examples/pull/425

peterharperuk:
Fixes #426

### DIFF
--- a/pico_w/bt/README.md
+++ b/pico_w/bt/README.md
@@ -1,0 +1,18 @@
+# Pico W Bluetooth Examples
+
+The source for most of the Bluetooth examples is stored in `pico-sdk/lib/btstack/example`.
+There's a standalone example in `pico-examples/pico_w/bt/standalone`.
+
+## Debugging
+
+To debug Bluetooth issues you can enable [btstack](https://github.com/bluekitchen/btstack) debug output which also enables packet logging.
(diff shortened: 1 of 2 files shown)


mringwal:
It is! I think the Bluetooth SIG insists on always writing Bluetooth with a capital 'B' (and I'm used to that as well).
Could you use "The Bluetooth examples that use audio requires code in .." to indicate why pico-extras might be needed (I think it's only used for audio demos, isn't it?)

pico_w/bt/README.md
```diff
@@ -0,0 +1,18 @@
...
+To debug Bluetooth issues you can enable [btstack](https://github.com/bluekitchen/btstack) debug output which also enables packet logging.
+Define `WANT_HCI_DUMP=1` in your CMakeLists.txt file. Uncomment this line to enable debug in the btstack examples.
+
+	target_compile_definitions(picow_bt_example_common INTERFACE
+	    #WANT_HCI_DUMP=1 # This enables btstack debug
+	    )
+
+## Packet logging
+
+To view packet logs, save the output from the debug port (e.g. the uart) to a file and run `pico-sdk/btstack/tools/create_packet_log.py <filename>`.
```
lurch, line 17: `lib/btstack/` instead of `btstack/`, and `tool/` instead of `tools/` ?

PR #429, opened 2023-10-04
Fix freertos compilation

peterharperuk:
Issues with the latest version of the FreeRTOS-Kernel


ex-fil:
This worked for me :top:

lurch:
I guess this conflicts with #460 ?

lurch:
I guess this also relates to https://github.com/raspberrypi/pico-sdk/pull/1530 ?

peterharperuk:
@kilograham I know we need to see if freertos have broken something else, but are you ok with me merging this? As it's better than nothing?

PR #445, opened 2023-12-03
Fix for composite report HIDs.

Franticware:
There was an extraneous byte in boot mouse report for composite device type mice. In existing code, this was only handled in process_generic_report, not tuh_hid_report_received_cb. This is a fix for that.
I also tested it on pi pico and it works fine.

### DIFF
--- a/usb/host/host_cdc_msc_hid/hid_app.c
+++ b/usb/host/host_cdc_msc_hid/hid_app.c
@@ -99,11 +99,11 @@ void tuh_hid_umount_cb(uint8_t dev_addr, uint8_t instance)
 void tuh_hid_report_received_cb(uint8_t dev_addr, uint8_t instance, uint8_t const* report, uint16_t len)
 {
   uint8_t const itf_protocol = tuh_hid_interface_protocol(dev_addr, instance);
-  uint8_t const rpt_count = tuh_hid_instance_count(dev_addr);
+  uint8_t const report_count = tuh_hid_instance_count(dev_addr);
 
   uint8_t const* reportAdj = report;
 
-  if ( rpt_count != 1 )
+  if ( report_count != 1 )
   {
     // Composite report, 1st byte is report ID, data starts from 2nd byte
     reportAdj++;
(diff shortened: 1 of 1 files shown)


usb/host/host_cdc_msc_hid/hid_app.c
```diff
@@ -99,17 +99,26 @@ void tuh_hid_umount_cb(uint8_t dev_addr, uint8_t instance)
 void tuh_hid_report_received_cb(uint8_t dev_addr, uint8_t instance, uint8_t const* report, uint16_t len)
 {
   uint8_t const itf_protocol = tuh_hid_interface_protocol(dev_addr, instance);
+  uint8_t const rpt_count = tuh_hid_instance_count(dev_addr);
```
lurch, line 102: _Tiny_ nitpick, but perhaps `report_count` would be a better variable name? When I first read this I thought that `rpt_count` meant "repeat count" :wink:

lurch:
ping @hathach just in case a similar problem affects TinyUSB too?

usb/host/host_cdc_msc_hid/hid_app.c
Franticware: Done.

Franticware:
> ping @hathach just in case a similar problem affects TinyUSB too?

Maybe, is he able to see this PR because he is tagged?

lurch:
Sorry, that was a question aimed at the TinyUSB developer, not at yourself Vojtěch :wink: 
(i.e. I wasn't saying "Should we ping Ha Thach?", I was saying "I'm pinging Ha Thach here (by mentioning his GitHub username), just in case TinyUSB is also affected".)

Franticware:
@lurch Should I do something to move forward the approval/merge process?

lurch:
@Franticware Nope, I think the only thing to do at this point is be patient... :slightly_smiling_face:

Franticware:
I now realize this fix is not good enough. Unfortunately there are other issues (the data passed to tuh_hid_report_received_cb callback by tinyusb is incomplete) and I am not able to fix it at this point.

Franticware:
The key to solving this is using the following as part of TinyUSB initialization:
tuh_hid_set_default_protocol(HID_PROTOCOL_REPORT);

Then the report data must be interpreted according to the descriptor.

The default HID_PROTOCOL_BOOT does not work very well with many USB mice that I have tested so far.

lurch:
> The key to solving this is using the following as part of TinyUSB initialization: tuh_hid_set_default_protocol(HID_PROTOCOL_REPORT);
> 
> Then the report data must be interpreted according to the descriptor.
> 
> The default HID_PROTOCOL_BOOT does not work very well with many USB mice that I have tested so far.

Would you like to create a new issue and/or PR? I suspect that commenting on a closed PR will mean that your comment will quickly get forgotten / lost.

PR #450, opened 2023-12-19
Add additional C++ Example for AM2302-Sensor

hasenradball:
# Add a C++ Example for AM2302-Sensor.

- usage of class instantiation
- based on [AM2302-Sensor Lib](https://github.com/hasenradball/AM2302-Sensor)

### DIFF
--- a/gpio/AM2302_sensor/AM2302-Sensor.cpp
+++ b/gpio/AM2302_sensor/AM2302-Sensor.cpp
@@ -1,0 +1,191 @@
+/*
+*	AM2302-Sensor.cpp
+*
+*	Author: Frank Häfele
+*	Date:	21.11.2023
+*
+*	Object: Measure Sensor Data of AM2302-Sensor
+*
+*/
+#include "AM2302-Sensor.hpp"
+
+/**
+ * @brief Construct a new am2302::am2302 sensor::am2302 sensor object
+ * 
+ * @param pin Pin for AM2302 sensor
+ */
+AM2302::AM2302_Sensor::AM2302_Sensor(uint8_t pin) : _millis_last_read{0}, _pin{pin}
+{}
+
+/**
+ * @brief begin function setup pin and run sensor check.
+ * 
+ * @return true if sensor check is successful.
+ * @return false if sensor check failed.
+ */
+bool AM2302::AM2302_Sensor::begin() {
+   gpio_init(_pin);
+   // required delay() for a secure sensor check,
+   // if you reset the mcu very fast one after another
+   auto tic{time_us_64()};
+   while ( time_us_64() - tic < READ_FREQUENCY * 1000U ) {
+      sleep_ms(1U);
+   }
+   auto status{read()};
+   _millis_last_read = time_us_64();
+   if (status == AM2302_READ_OK) {
+      return true;
+   }
+   else {
+      return false;
+   }
+}
+
+
+int8_t AM2302::AM2302_Sensor::read() {
+   // check read frequency
+   if ( time_us_64() - _millis_last_read < READ_FREQUENCY * 1000U) {
+      return AM2302_ERROR_READ_FREQ;
+   }
+   _millis_last_read = time_us_64();
+   // *****************************
+   //  === send start sequence ===
+   // ****************************
+   // start from HIGH ==> switch to LOW for min. of 1 ms
+   // Set pin to Output
+   gpio_set_dir(_pin, GPIO_OUT);
+   // set Pin to LOW 
+   gpio_put(_pin, 0);
+   // wait min. 1,0 ms
+   sleep_us(1200U);
+   // Set port to HIGH ==> INPUT_PULLUP with PullUp
+   gpio_put(_pin, 1);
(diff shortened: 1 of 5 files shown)


hasenradball:
@lurch Hi,

added in short a cpp example for the AM2302-sensor.
It is possible to add to the pico examples?

lurch:
As I might have mentioned before, I only provide "(hopefully) helpful comments", it's not my decision whether to merge PRs in `pico-sdk` or `pico-examples`. But my first helpful comment would be that you need to target your PR against `develop` instead of `master` :wink:

hasenradball:
@lurch Thanks for the hint.
You are totally right.

:-)

peterharperuk:
How is this different from the DHT22 code you've added? Or are you adding this as an example of how to use C++?

hasenradball:
@peterharperuk 
Hi Peter, the idea was to add this as C++ Example.
I got the info in the morning that cpp in gerneral is possible… using classes and instatiation and all this stuff.
Then I tried it directly.
It makes the main file a little cleaner.
Could be an example how it looks like in cpp style.
One thing is different here.
When call method begin() there is a check if sensor is connected.

And maybee here could also the check placed to distinguish between DHT11 or DHT22.

Best regards
Frank

hasenradball:
@peterharperuk 
One other thing ist also improved.
If you poll the sensor too often which means < 2 s .
Then status is -3 and the read function is not called. 

So code is slightly improved.
please try it.

gpio/AM2302_sensor/AM2302.cpp
```diff
@@ -8,7 +8,7 @@
 #include "pico/stdlib.h"
 #include "AM2302-Sensor.hpp"
 
-const uint LED = 13U;
+const uint LED_PIN = 13U;
```
lurch, line 11: Maybe you'd like to make use of the `PICO_DEFAULT_LED_PIN` define, so that your example doesn't require an external LED? https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/pico.h#L33
(see also e.g. https://github.com/raspberrypi/pico-examples/blob/master/blink/blink.c )

hasenradball:
@lurch can you please give a review for this PR.

lurch:
> As I might have mentioned before, I only provide "(hopefully) helpful comments", it's not my decision whether to merge PRs in `pico-sdk` or `pico-examples`.

hasenradball:
@lurch I thought so because you are mentioned as reviewer, sorry for the missunderstanding.

lurch:
I think GitHib just treats _any_ comment on the code as a "review" :facepalm:

hasenradball:
@hasenradball 

Added the following features:
- better human readable error output
- example with sensor array

Added screenshot how example looks: 

20240425_16h59m19s_grim

gpio/AM2302_sensor/AM2302-Sensor.cpp
```diff
@@ -0,0 +1,252 @@
...
+   gpio_init(_pin);
+   gpio_pull_up(_pin);
+   // required delay() for a secure sensor check,
+   // if you reset the mcu very fast one after another
+   auto tic{time_us_64()};
+   while ( time_us_64() - tic < READ_FREQUENCY * 1000U ) {
+      sleep_ms(1U);
+   }
+   auto status{read()};
+   _millis_last_read = time_us_64();
```
lurch, line 36: `time_us_64` returns the time in microseconds https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#ga7bbcd996dbd3dd4adc03a80272fb2fc3 so calling the variable `_millis_last_read` might be confusing to whoever reads this code in future? :wink:

gpio/AM2302_sensor/AM2302-Sensor.cpp
```diff
@@ -0,0 +1,252 @@
+/*
+*	AM2302-Sensor.cpp
+*
+*	Author: Frank Häfele
```
lurch, line 4: Could you add some licensing-information please?

hasenradball: @lurch 
any prefererations?

gpio/AM2302_sensor/AM2302-Sensor.cpp
hasenradball: sure youre right,
I should adapt this
👍

gpio/AM2302_sensor/AM2302-Sensor.cpp
lurch: As you're the author, it's obviously up to you. But if you choose the same licence as all of our other examples, that's probably easiest.

hasenradball: @lurch fixed, hope it is fine.

lurch: For the avoidance of doubt, it's probably best to add the `License-Identifier` to each of the header-blocks in which you've listed yourself as the Author?

PR #452, opened 2023-12-26
Include urgbw_u32 function and document rgb vs rgbw

### LINKED ISSUE #451: Include some description about RGBW vs RGB in ws2812 example
Maybe it was just my lack of experience with WS2812, but when I was running the `ws2812` example, it was not working as expected.

I modify the code to try to understand the logic more in deep and at the end of the day the "fix" was very simple. The WS2812 that I was using was a `RGB` module not an `RGBW`.

It would be great to add a note that this is something to take into account when trying to run the example.

jcarranz97:
Fixes #451 

I have also changed the `IS_RGBW` to false because the example is using `urgb_u32` which does not include the `white` color.

### DIFF
--- a/pio/ws2812/ws2812.c
+++ b/pio/ws2812/ws2812.c
@@ -12,7 +12,20 @@
 #include "hardware/clocks.h"
 #include "ws2812.pio.h"
 
-#define IS_RGBW true
+/**
+ * NOTE:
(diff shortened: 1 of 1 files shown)


peterharperuk:
Ha. Thanks. That explains why my led strip looked so bad.

peterharperuk (approved):

PR #458, opened 2024-01-04
Add lwip httpd examples

### LINKED ISSUE #266: add freertos httpd example
I added this for some reason - may as well have another free rtos example?

### LINKED ISSUE #272: Add a test for mDns
I don't fully understand mDns but I think it gives me the ability to register a local domain name, so a web server can register and you can surf to it like http://picow1a28.local/?

Anyway - there's code in the SDK wrapped in #if LWIP_MDNS_RESPONDER, so we should test it works?

peterharperuk:
Fixes #266
Fixes #272


peterharperuk:
Depends on https://github.com/raspberrypi/pico-sdk/pull/1600

pico_w/wifi/freertos/httpd/FreeRTOSConfig.h
```diff
@@ -0,0 +1,145 @@
+/*
```
kilograham, line 1: i have no problem with this here (and possibly it is best here), just wasn't sure why/if this is different given that we use a common header elsewhere in the examples

kilograham (changes requested):
should update README.md

pico_w/wifi/freertos/httpd/FreeRTOSConfig.h
peterharperuk: All the wifi freertos example seems to duplicate FreeRTOSConfig.h. Maybe I should stop them doing that?

README.md
```diff
@@ -130,6 +130,7 @@ App|Description
 [picow_tls_verify](pico_w/wifi/tls_client) | Demonstrates how to make a HTTPS request using TLS with certificate verification.
 [picow_wifi_scan](pico_w/wifi/wifi_scan) | Scans for WiFi networks and prints the results.
 [picow_udp_beacon](pico_w/wifi/udp_beacon) | A simple UDP transmitter.
```
Hax2022, line 132: uzùi

Hax2022 (commented):
# 55rr₩44

peterharperuk:
> should update README.md

Done

kilograham (approved):

PR #459, opened 2024-01-04
Use the function to set the IP address

peterharperuk:
In access point example


peterharperuk:
Depends on https://github.com/raspberrypi/pico-sdk/pull/1424

kilograham (approved):

PR #460, opened 2024-01-05
Remove duplicate FreeRTOSConfig.h files

peterharperuk:
Include FreeRTOSConfig_examples_common.h, similar to what we do for lwipopts.h


pico_w/wifi/freertos/FreeRTOSConfig_examples_common.h
```diff
@@ -0,0 +1,145 @@
...
+#define configTIMER_TASK_STACK_DEPTH            1024
+
+/* Interrupt nesting behaviour configuration. */
+/*
+#define configKERNEL_INTERRUPT_PRIORITY         [dependent of processor]
+#define configMAX_SYSCALL_INTERRUPT_PRIORITY    [dependent on processor and application]
+#define configMAX_API_CALL_INTERRUPT_PRIORITY   [dependent on processor and application]
+*/
+
+#if FREE_RTOS_KERNEL_SMP // set by the RP2040 SMP port of FreeRTOS
```
eamars, line 105: Should this line be removed too? I don't see where this macro is defined.

peterharperuk: I see this being set in FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/library.cmake
So I think the answer is no.

PR #463, opened 2024-01-17
Some audio fixes

peterharperuk:
If you initialise audio twice it fails
Initialise the gpios used


peterharperuk:
Seems to be a duplicate

PR #465, opened 2024-01-20
Add extern "C" to isr_usbctrl in dev_lowlevel

supsm:
The USB dev_lowlevel example silently fails in C++ because `isr_usbctrl` is name-mangled and therefore not registered as an interrupt handler. This PR adds `extern "C"` to use C linkage for this function and allow compiling as C++.  

It is certainly not the prettiest, but it is necessary for correct behavior (with C++ linkage, none of the host's packets are responded to and the device is useless). The alternative to this is to manually specify `irq_set_exclusive_handler`, probably within `usb_device_init`. The pico-sdk advertises itself as C/C++, so C++ compatibility should be taken into consideration in its examples.

### DIFF
--- a/usb/device/dev_lowlevel/dev_lowlevel.c
+++ b/usb/device/dev_lowlevel/dev_lowlevel.c
@@ -489,6 +489,9 @@ static void usb_handle_buff_status() {
  *
  */
 /// \tag::isr_setup_packet[]
+#ifdef __cplusplus
+extern "C" {
+#endif
 void isr_usbctrl(void) {
     // USB interrupt handler
     uint32_t status = usb_hw->ints;
@@ -520,6 +523,9 @@ void isr_usbctrl(void) {
         panic("Unhandled IRQ 0x%x\n", (uint) (status ^ handled));
     }
 }
+#ifdef __cplusplus
+}
+#endif
 
 /**
  * @brief EP0 in transfer complete. Either finish the SET_ADDRESS process, or receive a zero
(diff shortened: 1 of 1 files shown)


peterharperuk (approved):

usb/device/dev_lowlevel/dev_lowlevel.c
```diff
@@ -489,6 +489,9 @@ static void usb_handle_buff_status() {
  *
  */
 /// \tag::isr_setup_packet[]
+#ifdef __cplusplus
+extern "C" {
+#endif
```
lurch, line 494: Should this snippet be moved **above** the `/// \tag::isr_setup_packet[]` line? I don't think we want this `#ifdef` stuff appearing in section 4.1.3.2.3. of https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf ?

peterharperuk (changes requested):
See comment from @lurch

kilograham:
I'm not sure I understand why/how are you trying to compile this example as C++.(or what you mean by it).

usb/device/dev_lowlevel/dev_lowlevel.c
supsm: I wasn't aware of what that did. It should be fixed now.

peterharperuk: Me neither!

supsm:
> I'm not sure I understand why/how are you trying to compile this example as C++

I initially saved the file as .cpp to modify it in a more familiar language (not realizing isr_usbctrl is a special function). I suspect other developers may also attempt to unsuccessfully incorporate isr_usbctrl into C++ code.

supsm:
Is there any update on this pr? I have already made the requested changes.

peterharperuk (approved):

PR #476, opened 2024-02-28
reply through the same PCB packet came in

mtz29:
`d->udp` should be used to set up callbacks and start up UDP server only. Replying via the same PCB instance is freezing Picow.


lurch:
Can you change the base branch to `develop` please?

mtz29:
Raised https://github.com/raspberrypi/pico-examples/pull/477

PR #440, opened 2023-11-29
[Bugfix] Refracture not working `dht.c` Example

hasenradball:
# Refracture `dht.c` Example
- improve functionality
- add constants for:
  - Timeout Error
  - Checksum Error
  - Read OK

This code is adapted to the Pico c files and was mainly taken from here:
[AM2302-sensor](https://github.com/hasenradball/AM2302-Sensor)
  
  ## Tested 
  Code is tesed on Arduino NANO, ESP8266, and also on Pi Pico 

### Screeshot
2023-11-29-223802_1920x1200_scrot

PR #485, opened 2024-03-28
fix dht.c error: can read data from DHT 11

ZhongUncle:
Hello,
I notice when I use the example `gpio/dht_sensor` in right way, Pico can't read data from DHT11. It toke a night and after reading datasheet [https://www.mouser.com/datasheet/2/758/DHT11-Technical-Data-Sheet-Translated-Version-1143054.pdf](https://www.mouser.com/datasheet/2/758/DHT11-Technical-Data-Sheet-Translated-Version-1143054.pdf), I find this error according 2 errors:
1. In Line 49~53, it doesn't set high signal in some time.
2. In Line 73, the range of `count` (accumulated time) is 26～28, so need a number larger than 28, I choose 35 for redundancy.
 
This is first time I find this error [[Why DHT11 not work after some times using C in pico - Raspberry Pi Forums](https://forums.raspberrypi.com/viewtopic.php?t=355462#p2130252)](https://forums.raspberrypi.com/viewtopic.php?t=355462l), in this post, I use USB and minicom on Mac.

Recently, I buy a Raspberry Pi 4B to re-test this error following the document [Raspberry Pi Pico C/C++ SDK](https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf) by UART, this error still exists. It displays:

截屏2024-03-28 23 07 16

The print's dislocation may be according to I ssh to raspberry pi. But the data from DHT11  also are `0`, and print `Bad data`.

After fixing, it displays:
截屏2024-03-28 23 30 45

It works well.

### DIFF
--- a/gpio/dht_sensor/dht.c
+++ b/gpio/dht_sensor/dht.c
@@ -48,8 +48,12 @@ void read_from_dht(dht_reading *result) {
 
     gpio_set_dir(DHT_PIN, GPIO_OUT);
     gpio_put(DHT_PIN, 0);
-    sleep_ms(20);
+    sleep_ms(18);
+    gpio_put(DHT_PIN,1);
+    sleep_us(40);
+    
     gpio_set_dir(DHT_PIN, GPIO_IN);
+    
 
 #ifdef LED_PIN
     gpio_put(LED_PIN, 1);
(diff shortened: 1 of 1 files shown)


hasenradball:
@ZhongUncle have a look at:

I completly reworked this example, see: 

a) PR #440
b) PR #450 
c) https://github.com/hasenradball/AM2302-Sensor

Issue #352, opened 2023-03-11
The Pico i2c C example for the MPU6050 only returns 0x00 as gyro, accel and temp values


svkers:
The provided [Pico C i2c example for the MPU6050 ](https://github.com/raspberrypi/pico-examples/blob/master/i2c/mpu6050_i2c/mpu6050_i2c.c) does not work as expacted: It only returns 0x00 instead of the real gyro, accel and temp values.

After struggeling several time I finally managed to get real values from the MPU6050 by changing the following code line:

<pre>
```
static void mpu6050_reset() {
    // Two byte reset. First byte register, second byte data
    // There are a load more options to set up the device in different ways that could be added here
    uint8_t buf[] = {0x6B, <b>0x80</b>};
    i2c_write_blocking(i2c_default, addr, buf, 2, false);
}
```
</pre>

to 

<pre>
```
static void mpu6050_reset() {
    // Two byte reset. First byte register, second byte data
    // There are a load more options to set up the device in different ways that could be added here
    uint8_t buf[] = {0x6B, <b>0x00</b>};
    i2c_write_blocking(i2c_default, addr, buf, 2, false);
}
```
</pre>

The [MPU-6500 Register Map and Descriptions Revision 2.1](https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6500-Register-Map2.pdf) states that register 0x6B is the PWR_MGMT_1 register:
     
DEVICE_RESET SLEEP CYCLE GYRO_STANDBY TEMP_DIS CLKSEL[2:0]

So setting this register to 0x80 (binary: 1000 0000) means that the bit DEVICE_RESET is set to 1 whereas all other bits are set to 0.
The document also states for this register and bit the following:
DEVICE_RESET 1 – Reset the internal registers and restores the default settings. Write a 1 to set the reset, the bit will auto clear.

Since I am quite new to Pico and electronics I would be really glad for further clarification. Thanks!

Note: I faced this issue with various MPU6050 boards that I tested.

vacoff:
Man I can't tell you how much time I wasted on this topic. I didn't ever think the implementation could be the source of the problem. I spent 6 days in span of 3 months try to get reading from 3 different sensors, the last one which I ordered new try to narrow down the problem.

PR #494, opened 2024-04-30
fix mpu6050 sensor initialization issue

Bartosz89k:
In my case the sensor (mpu6050), was not initialized correctly and returns erroneous data. 
The proposed changes works fine.
In particular, I'm not sure why the second byte (0x80 == 1000 0000) is messing up the initialization.
As I understand from the documentation (page 40):
https://pdf1.alldatasheet.com/datasheet-pdf/view/1132807/TDK/MPU-6050.html
the seventh bit of register 0x6b should be set to one:
0x80 = 1000 0000
but it doesn't work that way.


lurch:
Related to #352 and the big discussion in #319 ?

PR #495, opened 2024-05-01
Simplify example of stdio_set_chars_available_callback

### LINKED ISSUE #461: Make sure there's an example of using stdio_set_chars_available_callback
Make sure it works with USB
See https://github.com/raspberrypi/pico-sdk/issues/1603

peterharperuk:
The example in here uses an async callback to avoid a race when using stdio_usb. But there's a fix available for this issue so simplify the example.

To test pass -DPICO_STDIO_USB=1 on the cmake command line.

Fixes #461

### DIFF
--- a/pico_w/wifi/iperf/picow_iperf.c
+++ b/pico_w/wifi/iperf/picow_iperf.c
@@ -36,21 +36,12 @@ static void iperf_report(void *arg, enum lwiperf_report_type report_type,
 #endif
 }
 
-// This "worker" function is called to safely perform work when instructed by key_pressed_func
-void key_pressed_worker_func(async_context_t *context, async_when_pending_worker_t *worker) {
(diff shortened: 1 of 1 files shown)


peterharperuk:
Requires this fix in the sdk https://github.com/raspberrypi/pico-sdk/pull/1649

peterharperuk:
@kilograham Simple fix to remove some of my sillyness. Will merge if that's ok.

kilograham (approved):
prepared to believe you ;-) though curious what / why of the disable_sta_mode?

peterharperuk:
This is "demonstrating" clean shutdown of the wifi stack. The loop keeps going while the link is up so pressing "d" calls disable_sta_mode which disables wifi and causes the program to exit.

PR #496, opened 2024-05-01
Set PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS

peterharperuk:
If you use stdio_usb for the examples you miss some of the output while USB is connected. Set PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS to a sensible default so you see all the output


CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ endif()
 
 set(PICO_EXAMPLES_PATH ${PROJECT_SOURCE_DIR})
 
+# If you want debug output from USB (pass -DPICO_STDIO_USB=1) this ensures you don't lose any debug output while USB is set up
+set(PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS 3000)
```
peterharperuk, line 19: We should probably allow it to be overridden

peterharperuk:
@kilograham Is this change ok? Just sets PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS for the examples to 3000ms

kilograham:
sounds reasonable

kilograham (approved):

PR #501, opened 2024-06-16
fix error img_to_array.py

wahidshirdel:
when the last part of the image generate  in .h file it does not involve ";" at the end ,and will make error  when we start make the project


peterharperuk (approved):
Certainly looks wrong. Thanks

PR #507, opened 2024-07-22
Make the blink example work on Pico W

### LINKED ISSUE #421: blink does not blink on a Pico W
The Pico W 'hello world' (aka blink) out-of-the-box experience is terrible as of [sdk-1.5.1](https://github.com/raspberrypi/pico-examples/tree/sdk-1.5.1).

Please add at least an additional `#warning if using a pico w, blink has moved to pico-examples/pico_w/wifi/blink` in `pico-examples\blink\blink.c`. Preferred would be to update `blink.c` to include both the vanilla and W code paths. I suspect beginners to the Pico (W)(H) platform using the C SDK will be able to figure out what is going on, and it could be a good place to include some comments about other potential gotcha differences between the two boards.

While truly the LED is now connected to the wireless chip and therefore pedantically sample code for it should be in the `pico-examples/pico_w` directory, I highly doubt anyone looking for `blink` cares about that distinction when looking for a first thing to run to figure out if their board is working _at all_.

The official [overview documentation](https://www.raspberrypi.com/documentation/microcontrollers/raspberry-pi-pico.html#raspberry-pi-pico-w-and-pico-wh) has a few W-specific notes; none of them describe the moved LED. It is only section 3.8 of [the Pico W datasheet](https://datasheets.raspberrypi.com/picow/pico-w-datasheet.pdf) that gets around to rather dryly describing this move that then *might* point someone in the correct direction in the examples directory:

> There are three GPIO pins from the CYW43439 that are used for other board functions and can easily be accessed via the SDK:
> [...]
> WL_GPIO0
> OP connected to **user LED**

#350 has drifted off topic to somewhat address this, but it seems like a new issue is more appropriate.

peterharperuk:
It seems to confuse people that the blink example doesn't work on PicoW. To avoid further confusion, make it work, rather than build and silently do nothing.

Fixes #421

### DIFF
--- a/blink/CMakeLists.txt
+++ b/blink/CMakeLists.txt
@@ -5,6 +5,10 @@ add_executable(blink
 # pull in common dependencies
 target_link_libraries(blink pico_stdlib)
 
+if (PICO_CYW43_SUPPORTED)
+        target_link_libraries(blink pico_cyw43_arch_none)
+endif()
+
 # create map/bin/hex file etc.
 pico_add_extra_outputs(blink)
 
--- a/blink/CMakeLists.txt
+++ b/blink/CMakeLists.txt
@@ -1,12 +1,12 @@
 add_executable(blink
-        blink.c
-        )
+    blink.c
+    )
 
 # pull in common dependencies
 target_link_libraries(blink pico_stdlib)
 
 if (PICO_CYW43_SUPPORTED)
-        target_link_libraries(blink pico_cyw43_arch_none)
+    target_link_libraries(blink pico_cyw43_arch_none)
 endif()
 
 # create map/bin/hex file etc.
--- a/blink/blink.c
+++ b/blink/blink.c
@@ -6,18 +6,39 @@
 
 #include "pico/stdlib.h"
 
-int main() {
 #ifndef PICO_DEFAULT_LED_PIN
-#warning blink example requires a board with a regular LED
-#else
-    const uint LED_PIN = PICO_DEFAULT_LED_PIN;
-    gpio_init(LED_PIN);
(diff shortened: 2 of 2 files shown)


blink/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ add_executable(blink
 # pull in common dependencies
 target_link_libraries(blink pico_stdlib)
 
+if (PICO_CYW43_SUPPORTED)
+        target_link_libraries(blink pico_cyw43_arch_none)
```
kilograham, line 9: 8 spaces upsets me ;-) 4 please

kilograham (commented):
Maybe add a few comments

blink/CMakeLists.txt
peterharperuk: Fixed

kilograham (approved):

saipovab83-png:
@startuml
participant "Transport Sensor" as Sensor
participant "Card Reader" as Reader
participant "Barrier Controller" as Controller

autonumber "<b>[ ]"

Sensor -> Reader : E1 (Detection) [0 ms]
activate Sensor
activate Reader

Reader -> Reader : E3 (Start Reading) [50 ms]
activate Reader

Reader -> Reader : E4 - E5 (Card Reading) [60-110 ms]
deactivate Reader

Reader -> Controller : E5 (Card Read Complete) [110 ms]
activate Controller

Controller -> Controller : E8 (Start Opening) [310 ms]
activate Controller

Controller -> Controller : Barrier Opening [330 ms]
deactivate Controller

Controller --> Reader : Operation Complete (mantiqiy)
deactivate Controller
deactivate Sensor

@enduml

PR #518, opened 2024-08-18
Update README.md

chrisboross:
@kilograham A few proposed edits to the README, backing out some early access copy and calling out which examples are RP235x only.


chrisboross:
Missed a few

PR #522, opened 2024-08-20
Skip encrypted bootloader with no mbedtls

will-v-pi:
Currently, a pico-examples build without the mbedtls submodule checked out in the SDK will fail, as picotool throws
```
ERROR: Cannot sign/hash partition table with no mbedtls
```
when making `enc_bootloader`, due to being compiled without mbedtls

Solve this by skipping `enc_bootloader` when there is no `pico_mbedtls`

### DIFF
--- a/bootloaders/CMakeLists.txt
+++ b/bootloaders/CMakeLists.txt
@@ -1,1 +1,6 @@
-add_subdirectory_exclude_platforms(encrypted host rp2040 rp2350-riscv)
+if (TARGET pico_mbedtls)
+    add_subdirectory_exclude_platforms(encrypted host rp2040 rp2350-riscv)
+else()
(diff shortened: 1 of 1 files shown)


kilograham:
small nit, but the convention is to put the guard in the individual example's CMakeLists.txt unless it affects multiple examples

lurch (commented):
This example now gets correctly skipped inside a Docker image where I've cloned pico-sdk but not init-ed its submodules

will-v-pi:
> small nit, but the convention is to put the guard in the individual example's CMakeLists.txt unless it affects multiple examples

This will probably end up applying to multiple bootloader examples in the long run, so I think it's ok? It applies to anything that requires signing/hashing with picotool

kilograham (approved):

PR #525, opened 2024-08-22
Fix compilation for PICO_BOARD=pico_w when cyw43_arch is missing

lurch:
Currently, if you try to build all of `pico-examples` for the Pico W and you haven't yet init-ed the `cyw43-driver` submodule in `pico-sdk`, then the build of `pico-examples` will halt with a fatal build-error. This PR fixes that :slightly_smiling_face: 

I'm not very familiar with CMake though, so perhaps there's a "neater" way of doing what I'm doing here?


kilograham:
pretty sure this is supposed to work so @peterharperuk may want to revist this

peterharperuk:
If ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE} is missing then we don't add the cyw43_driver_picow library. If cyw43_driver_picow doesn't exist we don't add the pico_cyw43_arch library. We have this guarding the pico_w examples, so I guess we should have it here as well.

if (PICO_CYW43_SUPPORTED) # set by PICO_BOARD=pico_w
    if (NOT TARGET pico_cyw43_arch)
        message("Skipping Pico W examples as support is not available")
    else()

CMakeLists.txt
```diff
@@ -43,7 +43,15 @@ function(add_subdirectory_exclude_platforms NAME)
 endfunction()
 
 # Add blink example
-add_subdirectory_exclude_platforms(blink)
+if (PICO_CYW43_SUPPORTED) # set by PICO_BOARD=pico_w
```
kilograham, line 46: Can we make this `if (PICO_CYW43_SUPPORTED AND NOT TARGET pico_cyw43_arch)`

peterharperuk: That might be better

CMakeLists.txt
```diff
@@ -43,12 +43,8 @@ function(add_subdirectory_exclude_platforms NAME)
 endfunction()
 
 # Add blink example
-if (PICO_CYW43_SUPPORTED) # set by PICO_BOARD=pico_w
-    if (NOT TARGET pico_cyw43_arch)
-        message("Skipping blink for Pico W as support is not available")
-    else()
-        add_subdirectory_exclude_platforms(blink)
-    endif()
+if (PICO_CYW43_SUPPORTED AND NOT TARGET pico_cyw43_arch)
```
kilograham, line 46: maybe "Wi-Fi driver is not available"

lurch: I simply re-used the wording I found in another CMakeLists.txt, so I'll let @peterharperuk decide what to do here.

peterharperuk (dismissed):

CMakeLists.txt
```diff
@@ -43,7 +43,11 @@ function(add_subdirectory_exclude_platforms NAME)
 endfunction()
 
 # Add blink example
-add_subdirectory_exclude_platforms(blink)
+if (PICO_CYW43_SUPPORTED AND NOT TARGET pico_cyw43_arch)
+    message("Wi-Fi driver is not available")
```
kilograham, line 47: I had meant "Skipping blink example as Wi-Fi driver is not available"

Can you also move this check into the blink CMakeLists.txt

adc/CMakeLists.txt
```diff
@@ -5,7 +5,11 @@ if (TARGET hardware_adc)
     add_subdirectory_exclude_platforms(joystick_display)
     add_subdirectory_exclude_platforms(onboard_temperature)
     add_subdirectory_exclude_platforms(microphone_adc)
-    add_subdirectory_exclude_platforms(read_vsys)
+    if (PICO_CYW43_SUPPORTED AND NOT TARGET pico_cyw43_arch)
+        message("Skipping read_vsys for Pico W as support is not available")
```
kilograham, line 9: and fix this up to be "as Wi Fi driver is not available"

CMakeLists.txt
lurch: > Can you also move this check into the blink CMakeLists.txt

By the time we're "in" the blink CMakeLists.txt, is it not then too late to call `add_subdirectory_exclude_platforms(blink)` ?

seamusdemora:
just a "throw-away" comment: 
Not sure what all the comments were about, but as someone who got "tripped up" with the `cyw43` thing, I hope this issue has been fixed - and thanks!

lurch:
> as someone who got "tripped up" with the `cyw43` thing, I hope this issue has been fixed - and thanks!

The issue will be fixed _after_ this PR is merged, but we're still discussing the finer details :slightly_smiling_face:

kilograham (changes requested):
marked as unapproved since there are comments

peterharperuk:
@kilograham I've addressed the comments for this and it seems to work now.

binary_info/blink_any/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+if (PICO_CYW43_SUPPORTED AND NOT TARGET pico_cyw43_arch)
+    return()
```
kilograham, line 2: why do some of these have a message (there is one without below too), and some not?

peterharperuk: So you don't see the same message repeatedly?

PR #527, opened 2024-08-23
Fix SSD1306 example to display all text on small (32px) screens

kaycebasques:
The last 4 lines of `text` are never displayed on the 32px-tall versions of SSD1306. This PR makes sure all of `text` is always displayed (at the cost of making the logic a bit more complex).

### DIFF
--- a/i2c/ssd1306_i2c/ssd1306_i2c.c
+++ b/i2c/ssd1306_i2c/ssd1306_i2c.c
@@ -387,30 +387,22 @@ restart:
         "by the name of",
         "    PICO"
     };
-    int max_lines = SSD1306_HEIGHT / 8;  // Each line is 8px tall.
     int y = 0;
     for (uint i = 0; i < count_of(text); i++) {
-        int line_number = i + 1;
         WriteString(buf, 5, y, text[i]);
         y += 8;
-        // On the 32px-tall versions of SSD1306, we can't display
-        // all the text at once. We need to display 4 lines at a time.
-        bool is_end_of_screen = (line_number % max_lines) == 0;
-        if (is_end_of_screen) {
+        // Height limit reached. Show some lines.
+        if (y == SSD1306_HEIGHT) {
             render(buf, &frame_area);
             sleep_ms(3000);
             memset(buf, 0, SSD1306_BUF_LEN);
             y = 0;
         }
-        // If it's the last line...
-        if (line_number == count_of(text)) {
-            // And the last line is not a multiple of max_lines...
-            if (line_number % max_lines != 0) {
-                // Then we will have a little more text to display.
-                render(buf, &frame_area);
-                sleep_ms(3000);
-            }
-        }
+    }
+    // Check if there's any more text left to display.
+    if (y != 0) {
+        render(buf, &frame_area);
+        sleep_ms(3000);
     }
 
     // Test the display invert function
--- a/i2c/ssd1306_i2c/ssd1306_i2c.c
+++ b/i2c/ssd1306_i2c/ssd1306_i2c.c
@@ -387,6 +387,7 @@ restart:
         "by the name of",
         "    PICO"
     };
+
     int y = 0;
     for (uint i = 0; i < count_of(text); i++) {
         WriteString(buf, 5, y, text[i]);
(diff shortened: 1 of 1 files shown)


i2c/ssd1306_i2c/ssd1306_i2c.c
```diff
@@ -387,13 +387,31 @@ int main() {
...
+        if (is_end_of_screen) {
+            render(buf, &frame_area);
+            sleep_ms(3000);
+            memset(buf, 0, SSD1306_BUF_LEN);
+            y = 0;
+        }
+        // If it's the last line...
+        if (line_number == count_of(text)) {
+            // And the last line is not a multiple of max_lines...
+            if (line_number % max_lines != 0) {
```
lurch, line 408: **If** I'm understanding the code correctly, I guess
```c
            // And the last line is not a multiple of max_lines...
            if (line_number % max_lines != 0) {
```
could be replaced by:
```c
            // And there's pending text in buf...
            if (y != 0) {
```
?
And that'd then allow this part of the code to be moved to _after_ the `for` loop? :thinking: (which would remove the `if (line_number == count_of(text))` test on every loop-iteration)

kaycebasques:
accidentally closed the pr by deleting my local branch, re-opening

i2c/ssd1306_i2c/ssd1306_i2c.c
kaycebasques: Ah yes this can probably be simplified by relying on `y`, let me revisit

lurch:
Also, can you change the base branch to `develop` please?

i2c/ssd1306_i2c/ssd1306_i2c.c
kaycebasques: OK, much simpler now. Please take another look @lurch

i2c/ssd1306_i2c/ssd1306_i2c.c
```diff
@@ -387,13 +387,23 @@ int main() {
...
+            render(buf, &frame_area);
+            sleep_ms(3000);
+            memset(buf, 0, SSD1306_BUF_LEN);
+            y = 0;
+        }
+    }
+    // Check if there's any more text left to display.
+    if (y != 0) {
+        render(buf, &frame_area);
+        sleep_ms(3000);
```
lurch, line 405: Probably don't need the `sleep_ms(3000);` here, because there's already a `sleep_ms(3000);` on line 409 ?

i2c/ssd1306_i2c/ssd1306_i2c.c
```diff
@@ -387,13 +387,23 @@ int main() {
         "by the name of",
         "    PICO"
     };
-
```
lurch, line 390: Minor nitpick, but could you add in this blank line again please?

kaycebasques: Done

i2c/ssd1306_i2c/ssd1306_i2c.c
kaycebasques: I think we should leave this one and remove the one on line 409

Let me leave another comment on line 399 to explain

i2c/ssd1306_i2c/ssd1306_i2c.c
```diff
@@ -387,13 +387,23 @@ int main() {
...
     int y = 0;
-    for (uint i = 0 ;i < count_of(text); i++) {
+    for (uint i = 0; i < count_of(text); i++) {
         WriteString(buf, 5, y, text[i]);
-        y+=8;
+        y += 8;
+        // Height limit reached. Show some lines.
+        if (y == SSD1306_HEIGHT) {
+            render(buf, &frame_area);
+            sleep_ms(3000);
```
kaycebasques, line 397: So what happens when you run this app is:

1. Lines 0 to 3 get printed
1. Wait 3s
1. Lines 4 to 7 get printed
1. Wait 3s
1. The extra check on lines 403 to 406 doesn't run with the default text because y is 0
1. Wait another 3s (this is the extra sleep on line 409 we should remove)

Leaving the sleep on line 397 and 405 ensures that text always (and only) displays for exactly 3s

lurch: Sounds good!

lurch:
I don't have a SSD1306 with which to test, but the code looks fairly plausible. Thanks for incorporating all my feedback @kaycebasques !

PR #534, opened 2024-08-27
Fix picotool_DIR and pioasm_DIR passing to universal build

### LINKED ISSUE #531: Universal examples ignore `picotool_DIR` setting
If I try and build all of `pico-examples` with something like this:
```console
cd pico-examples
mkdir build
cd build
cmake .. -DPICO_SDK_PATH=../../pico-sdk -DPICO_RISCV_TOOLCHAIN_PATH=$(realpath ../../corev-openhw-gcc-ubuntu2204-20240530) -DPICO_ARM_TOOLCHAIN_PATH=/usr -Dpicotool_DIR=$(realpath ../../picotool.install/picotool)
```
(where `../../picotool.install` is the location that I've done a "flat install" of `picotool` into), then when it gets as far as building the universal examples, it says "No installed picotool with version 2.0.0 found - building from source", and then downloads and builds a fresh copy of `picotool`! :facepalm: 
By looking at https://github.com/raspberrypi/pico-examples/blob/develop/universal/CMakeLists.txt#L55 I _was_ able to get this working with:
```console
cmake .. -DPICO_SDK_PATH=../../pico-sdk -DPICO_RISCV_TOOLCHAIN_PATH=$(realpath ../../corev-openhw-gcc-ubuntu2204-20240530) -DPICO_ARM_TOOLCHAIN_PATH=/usr -Dpicotool_DIR=$(realpath ../../picotool.install/picotool) -Dpicotool_INSTALL_DIR=$(realpath ../../picotool.install)
```
i.e. by supplying both `-Dpicotool_DIR=$(realpath ../../picotool.install/picotool)` **and** `-Dpicotool_INSTALL_DIR=$(realpath ../../picotool.install)`, but that's obviously a bit awkward / unexpected.

(And as an aside, I have to use `realpath` in a few places, because some parts of the build don't work properly with relative paths?)

will-v-pi:
If pictool_DIR is set then pass that, otherwise pass the picotool_INSTALL_DIR, which is only set when picotool is built as part of the SDK

Fixes #531 I think - although @lurch should probably double check

### DIFF
--- a/universal/CMakeLists.txt
+++ b/universal/CMakeLists.txt
@@ -48,6 +48,11 @@ function (add_universal_target TARGET SOURCE)
     endif()
 
     add_custom_target(${TARGET} ALL)
+    if (picotool_DIR)
+        set(universal_picotool_DIR ${picotool_DIR})
+    else()
+        set(universal_picotool_DIR ${picotool_INSTALL_DIR}/picotool)
+    endif()
 
     set(DEPS "")
     set(BINS "")
@@ -70,7 +75,7 @@ function (add_universal_target TARGET SOURCE)
                 "-DUNIVERSAL_PROJECT_DIR:FILEPATH=${SOURCE}"
                 "-DUNIVERSAL_BINARY_DIR:FILEPATH=${CMAKE_CURRENT_BINARY_DIR}/${TARGET}/${platform}"
                 "-DSOURCE_TARGET=${SOURCE_TARGET}"
-                "-Dpicotool_DIR=${picotool_INSTALL_DIR}/picotool"
+                "-Dpicotool_DIR=${universal_picotool_DIR}"
                 "-Dpioasm_DIR=${PIOASM_INSTALL_DIR}/pioasm"
             BUILD_ALWAYS 1 # force dependency checking
             INSTALL_COMMAND ""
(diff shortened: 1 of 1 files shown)


lurch:
Yup, I can conform this fixes #531 thanks!

universal/CMakeLists.txt
```diff
@@ -70,7 +75,7 @@ function (add_universal_target TARGET SOURCE)
                 "-DUNIVERSAL_PROJECT_DIR:FILEPATH=${SOURCE}"
                 "-DUNIVERSAL_BINARY_DIR:FILEPATH=${CMAKE_CURRENT_BINARY_DIR}/${TARGET}/${platform}"
                 "-DSOURCE_TARGET=${SOURCE_TARGET}"
-                "-Dpicotool_DIR=${picotool_INSTALL_DIR}/picotool"
+                "-Dpicotool_DIR=${universal_picotool_DIR}"
                 "-Dpioasm_DIR=${PIOASM_INSTALL_DIR}/pioasm"
```
lurch, line 79: Is a similar "fix" also needed here for `pioasm_DIR` ? (I've not experimented with a standalone-pioasm yet)

lurch: I've pushed a commit to your branch to make a similar change for `pioasm_DIR`, I hope that was okay @will-v-pi !

kilograham (approved):

universal/CMakeLists.txt
will-v-pi: Yep, that's great - thanks

PR #536, opened 2024-08-27
Add cmake script to update otp.json from privateaes.bin

### LINKED ISSUE #528: Output OTP decryption key JSON file for `encrypted` example always uses example key (`00010203...`)
In the example at `bootloaders/encrypted`, the README.md file gives instructions on how to generate a new AES key with the command:
```
dd if=/dev/urandom of=privateaes.bin bs=1 count=32
```
However, the `otp.json` file in the CMake build binary directory, which contains the data to be programmed into the decryption key OTP rows , is simply a copy of the `otp.json` in the example directory.
This can be confusing to the user since, if (the CMake binary directory copy of) `otp.json` is used to program OTP, the example key will be permanently burned into OTP, when they might expect the new key to be used instead.

This is particularly confusing as the signing key is correctly filed out (by picotool) to `hello_serial_enc.otp.json`.
I believe that there should either be a note indicating that the key in `otp.json` will not be updated with a newly-generated key, or some script/instructions for generating a decryption key OTP file (from `privateaes.bin`) should be added.

will-v-pi:
Currently, any changes to the key in privateaes.bin will not be reflected in the otp.json file, which could lead to issues if that is written to OTP.

This adds a cmake script using the JSON string handling to update the otp.json file if privateaes.bin is modified, before copying it to the binary directory. This fixes #528 

This will not work on cmake versions prior to 3.19, as those don't support JSON string handling, so instead it throws a warning if the file has been changed. This should only throw the warning if privateaes.json has been changed from the default.


bootloaders/encrypted/update-key.cmake
```diff
@@ -0,0 +1,23 @@
+if (CMAKE_VERSION VERSION_LESS 3.19)
+    # Check if keyfile is not the default, and print warning
+    file(READ ${CMAKE_CURRENT_LIST_DIR}/privateaes.bin key_file HEX)
+    if (NOT ${key_file} STREQUAL "000102030405060708090a0b0c0d0e0f00102030405060708090a0b0c0d0e0f0")
+        message(WARNING
+            "Encrypted bootloader AES key not updated in otp.json file, as CMake version is < 3.19"
```
kilograham, line 6: maybe this should be a fatal error?

will-v-pi: If this is a fatal error then I don't think it will work if the key file has been changed with older CMake versions, as this script gets run whenever the file is changed (and on first build), and on older CMake versions it has no way to parse the JSON file to check if the key in the JSON file matches the privateaes.bin file

kilograham (approved):

PR #537, opened 2024-08-27
Copy hello_serial into bootloaders/encrypted

will-v-pi:
Removes the dependence on a file from a separate folder, which other examples don't have

This would allow the example to be used in the VS Code extension, as that needs examples to be contained in a single directory

### DIFF
--- a/bootloaders/encrypted/CMakeLists.txt
+++ b/bootloaders/encrypted/CMakeLists.txt
@@ -64,7 +64,7 @@ example_auto_set_url(enc_bootloader)
 
 # Example binary to load
 add_executable(hello_serial_enc
-        ../../hello_world/serial/hello_serial.c
+        hello_serial.c
         )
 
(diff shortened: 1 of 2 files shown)


lurch:
> This would allow the example to be used in the VS Code extension, as that needs examples to be contained in a single directory

Does this also mean that some of the "universal examples" aren't buildable with VS Code? Or is that a non-issue as the VS Code extension doesn't support all the other "infrastructure" needed by those examples anyway?

will-v-pi:
Yeah, those aren't included in the VS Code extension yet, as they need both Arm and Risc-V toolchains. I think it'll be simpler to just add an option to make any binary universal in the extension, rather than including the examples

kilograham (approved):

kilograham:
I don't know that "universal" binary is that common a feature in general.

PR #538, opened 2024-08-27
Fix windows builds

will-v-pi:
Apply the same fix as in raspberrypi/pico-sdk#1846


will-v-pi:
I think that this needs merging to develop before the fix actually gets run

kilograham (approved):

lurch:
Fantastic! :tada: 
https://github.com/raspberrypi/pico-examples/commits/develop/

Issue #357, opened 2023-03-18
Skipping TinyUSB dual examples


approxdec:
g'day.
why - having done `git submodule update --init` for the sdk
do i get this? in particular the `TinyUSB available at` ... 
but then `Skipping TinyUSB dual examples, as TinyUSB hw/mcu/raspberry_pi/Pico-PIO-USB submodule unavailable`

this apprently makes \pico-examples\build\flash\program\flash_program.uf2
not work

regards
joe

I:\pico\pico-examples\build>cmake -G "NMake Makefiles" ..
Using PICO_SDK_PATH from environment ('I:\pico\pico-sdk')
PICO_SDK_PATH is I:/pico/pico-sdk
Defaulting PICO_PLATFORM to rp2040 since not specified.
Defaulting PICO platform compiler to pico_arm_gcc since not specified.
-- Defaulting build type to 'Release' since not specified.
PICO compiler is pico_arm_gcc
-- The C compiler identification is GNU 10.3.1
-- The CXX compiler identification is GNU 10.3.1
-- The ASM compiler identification is GNU
-- Found assembler: C:/ST/STM32CubeCLT/GNU-tools-for-STM32/bin/arm-none-eabi-gcc.exe
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/ST/STM32CubeCLT/GNU-tools-for-STM32/bin/arm-none-eabi-gcc.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/ST/STM32CubeCLT/GNU-tools-for-STM32/bin/arm-none-eabi-g++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
Build type is Release
Defaulting PICO target board to pico since not specified.
Using board configuration from I:/pico/pico-sdk/src/boards/include/boards/pico.h
-- Found Python3: C:/Python311/python.exe (found version "3.11.1") found components: Interpreter
TinyUSB available at I:/pico/pico-sdk/lib/tinyusb/src/portable/raspberrypi/rp2040; enabling build support for USB.
BTstack available at I:/pico/pico-sdk/lib/btstack
cyw43-driver available at I:/pico/pico-sdk/lib/cyw43-driver
Pico W Bluetooth build support available.
[...]

PR #540, opened 2024-08-29
Remove check for GCC version in usb dual example

lurch:
https://github.com/sekigon-gonnoc/Pico-PIO-USB/pull/70 already fixed Pico-PIO-USB to compile with GCC 11.3 and above


kilograham (approved):

arturo182:
I guess that resolves #357 :)

lurch:
@arturo182 But see also https://github.com/raspberrypi/pico-sdk/issues/1849

PR #541, opened 2024-08-30
Ignore build_* dirs

kaycebasques:


.gitignore
```diff
@@ -4,5 +4,6 @@ _deps
 cmake-*
 build
 build-*
+build_*
```
kaycebasques, line 7: `pico_setup.sh` seems to create `build_pico` and `build_pico2`: https://github.com/raspberrypi/pico-setup/blob/master/pico_setup.sh#L82

lurch: I guess that suggests that an alternative approach might be to modify `pico_setup.sh` to use build-directory names that **are** already covered by the `.gitignore` in `pico-examples`? ping @liamfraser 
The advantage of modifying the `pico-setup` repo instead of the `pico-examples` repo, is that edits to `pico-setup` are made directly on `master` (and so get "released" immediately); whereas edits to `pico-examples` get made on the `develop` branch (and so don't get "released" until @kilograham does the next merge from `develop` to `master`).

kilograham (approved):

Issue #552, opened 2024-09-25
bootloaders/encrypted routines are not functioning properly


Octopus1633:
Hello, 

I successfully compiled the bootloaders/encrypted example and generated two firmwares: enc_bootloader.uf2 and hello_serial_enc.uf2. I dragged enc_bootloader.uf2 into the device first, and then dragged hello_serial_enc.uf2 into the device, but it seems that the encryption program did not run in the end. Is this normal? What is the problem?
![image](https://github.com/user-attachments/assets/c019b7ee-d4ff-4757-9010-9d7431071410)

will-v-pi:
Did you write the generated otp.json file (in the build folder with the UF2s) to your OTP with `picotool otp load otp.json`? This example requires the AES key to be stored in OTP, else the decryption will not work - the generated otp.json file will write this key, lock the relevant OTP page, and enable secure boot on your device. I'll add this to the readme for this example

Note that writing the One-Time-Programable memory this way will only allow binaries signed with the private key used to run on your device.

lurch:
So this issue can be closed?

Octopus1633:
of course.

PR #553, opened 2024-09-25
Add note on writing OTP to enc bootloader

will-v-pi:
Add a note on how to write the AES key to OTP to the encrypted bootloader readme - prompted by #552


bootloaders/encrypted/README.md
```diff
@@ -4,12 +4,17 @@ Replace private.pem and privateaes.bin with your own keys - your signing key mus
...
 ```
 
-The AES key is just be a 32 byte binary file - you can create one with
+The AES key is just a 32 byte binary file - you can create one with
 
 ```bash
 dd if=/dev/urandom of=privateaes.bin bs=1 count=32
 ```
 
+You will need to program your OTP using the generated `otp.json` file in the build folder. Note that this will enable secure boot on your device, so only signed binaries can run, and will also lock down the page the AES key is stored in. If you wish to test without enabling secure boot then you can load the `otp.json` file in the source folder, which will just program the AES key and lock down that page.
```
lurch, line 13: "the page" -> "the OTP page" ?

lurch:
Would it be sensible for this README to refer people to chapter 10 of https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf for more information about secure boot ?

bootloaders/encrypted/README.md
will-v-pi:
```suggestion
You will need to program your OTP using the generated `otp.json` file in the build folder. Note that this will enable secure boot on your device, so only signed binaries can run, and will also lock down the OTP page the AES key is stored in. If you wish to test without enabling secure boot then you can load the `otp.json` file in the source folder, which will just program the AES key and lock down that OTP page.
```

will-v-pi:
Maybe chapter 5.10.1 instead, as that has more details on how to actually sign binaries? Or could point to both chapters

lurch:
> Maybe chapter 5.10.1 instead, as that has more details on how to actually sign binaries? Or could point to both chapters

I'll leave that decision entirely up to you, as you understand all of this _much_ better than I do!

bootloaders/encrypted/README.md
```diff
@@ -4,12 +4,17 @@ Replace private.pem and privateaes.bin with your own keys - your signing key mus
...
 
 ```bash
 dd if=/dev/urandom of=privateaes.bin bs=1 count=32
 ```
 
+You will need to program your OTP using the generated `otp.json` file in the build folder. Note that this will enable secure boot on your device, so only signed binaries can run, and will also lock down the OTP page the AES key is stored in. If you wish to test without enabling secure boot then you can load the `otp.json` file in the source folder, which will just program the AES key and lock down that OTP page.
+```bash
+picotool otp load otp.json
+```
+
```
will-v-pi, line 17:
```suggestion

For more information on security see chapter 10 of the [RP2350 datasheet](https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf), and for information on how to sign other binaries to run on a secure chip see chapter 5.10

```

bootloaders/encrypted/README.md
```diff
@@ -4,12 +4,19 @@ Replace private.pem and privateaes.bin with your own keys - your signing key mus
 openssl ecparam -name secp256k1 -genkey -out private.pem
 ```
 
-The AES key is just be a 32 byte binary file - you can create one with
+The AES key is just a 32 byte binary file - you can create one with
```
kilograham, line 7: : at the end

bootloaders/encrypted/README.md
```diff
@@ -4,12 +4,19 @@ Replace private.pem and privateaes.bin with your own keys - your signing key mus
...
 
 ```bash
 dd if=/dev/urandom of=privateaes.bin bs=1 count=32
 ```
 
+You will need to program your OTP using the generated `otp.json` file in the build folder. Note that this will enable secure boot on your device, so only signed binaries can run, and will also lock down the OTP page the AES key is stored in. If you wish to test without enabling secure boot then you can load the `otp.json` file in the source folder, which will just program the AES key and lock down that OTP page.
```
kilograham, line 13: is it clear what "the source folder" and "the build folder" are?

lurch: I wonder if it's worth adding a note telling the user that it's important that they don't lose the `private.pem` and `privateaes.bin` files?

will-v-pi:
Superseded by #619

Issue #554, opened 2024-09-26
otp/hello_otp example problem


Octopus1633:
Hi, I am running the otp/hello_otp example, why does the output show that the first write fails for both ECC and RAW, is this normal, is there any problem?
<img width="263" alt="企业微信截图_17273176772738" src="https://github.com/user-attachments/assets/09b27f9a-8473-4098-ae9e-e64a68f3dcd7">

will-v-pi:
Is this your second time running the example? This example will only work correctly when run for the first time on a board as it modifies the OTP (One-Time-Programable Memory), so if you run it a second time on the same board it will throw these errors - this bootrom error is documented in the SDK

https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/src/rp2_common/pico_bootrom/include/pico/bootrom_constants.h#L58

If you want to run this example again on the same board, you can modify the OTP rows used by the example - again, this will only work the first time for each set of OTP rows used

https://github.com/raspberrypi/pico-examples/blob/7e77a0c381863be0c49086567e7f1934d78ac591/otp/hello_otp/hello_otp.c#L18C1-L21C30

Octopus1633:
Thank you for your patience. After changing the address, I successfully ran the program again.
<img width="240" alt="企业微信截图_17273477136502" src="https://github.com/user-attachments/assets/c57e2522-0f45-411a-943b-0c3d3c8ac6c9">

lurch:
@will-v-pi I wonder if it'd be useful if that example was modified to print out the existing OTP row contents _before_ atempting to write to them, and/or check that the rows are indeed empty before trying to modify them? :thinking:  (he blindly suggests, knowing very little about OTP!)

will-v-pi:
Yeah, that sounds sensible - if the rows are non-zero then it can just throw an error and stop

PR #555, opened 2024-09-27
Check rows before running hello_otp example

will-v-pi:
Check that the OTP rows being used for the hello_otp example are 0 before the example starts, else the example will throw other errors. This was prompted by #554, so it will now throw an error in that case explaining that the rows need to be changed.

### DIFF
--- a/otp/hello_otp/hello_otp.c
+++ b/otp/hello_otp/hello_otp.c
@@ -20,6 +20,25 @@ int main() {
     // Row to write raw data
     uint16_t raw_row = 0x410;
 
+    // Check rows are empty - else the rest of the tests won't behave as expected
+    unsigned char initial_data[32] = {0};
+    cmd.flags = ecc_row;
+    ret = rom_func_otp_access(initial_data, sizeof(initial_data)/2, cmd);
+    if (ret) {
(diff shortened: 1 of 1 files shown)


kilograham (approved):

otp/hello_otp/hello_otp.c
```diff
@@ -20,6 +20,25 @@ int main() {
...
+        printf("ERROR: Initial ECC Row Read failed with error %d\n", ret);
+    }
+    cmd.flags = raw_row;
+    ret = rom_func_otp_access(initial_data+(sizeof(initial_data)/2), sizeof(initial_data)/2, cmd);
+    if (ret) {
+        printf("ERROR: Initial Raw Row Read failed with error %d\n", ret);
+    }
+    for (int i=0; i < sizeof(initial_data); i++) {
+        if (initial_data[i] != 0) {
+            printf("ERROR: This example requires empty OTP rows to run - change the ecc_row and raw_row variables to an empty row and recompile\n");
```
lurch, line 37: Minor nitpick: I wonder if "to an empty row" should have been "to empty rows", in case people misinterpret "change the ecc_row and raw_row variables to an empty row" to mean "set them to the same empty row"?

PR #567, opened 2024-10-24
Add binary info examples of configurable binaries

will-v-pi:
Add two examples of binaries which are configurable after compilation, demonstrating use of `bi_ptr` variables and the `picotool config` command.

The first is a blink binary which blinks on any pin, and if compiled for a W board it can also be configured between blinking a pin on the CYW43 chip or a normal GPIO

The second is a hello_world binary, with configurable UART config, ability to disable USB/UART output, and configurable printed string

### DIFF
--- a/binary_info/blink_any/blink_any.c
+++ b/binary_info/blink_any/blink_any.c
@@ -15,15 +15,18 @@
 // Note that LED_TYPE == 1 is only supported when initially compiled for
 // a board with PICO_CYW43_SUPPORTED (eg pico_w), else the required
 // libraries won't be present
+bi_decl(bi_program_feature_group(0x1111, 0, "LED Configuration"));
 #if defined(PICO_DEFAULT_LED_PIN)
-    bi_decl(bi_ptr_int32(0x1234, 0x5678, LED_TYPE, 0));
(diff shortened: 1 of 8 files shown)


binary_info/blink_any/blink_any.c
```diff
@@ -0,0 +1,73 @@
...
+#ifdef CYW43_WL_GPIO_LED_PIN
+#include "pico/cyw43_arch.h"
+#endif
+
+// Set an LED_TYPE variable - 0 is default, 1 is connected to WIFI chip
+// Note that LED_TYPE == 1 is only supported when initially compiled for
+// a board with PICO_CYW43_SUPPORTED (eg pico_w), else the required
+// libraries won't be present
+#if defined(PICO_DEFAULT_LED_PIN)
+    bi_decl(bi_ptr_int32(0x1234, 0x5678, LED_TYPE, 0));
```
peterharperuk, line 19: I wonder if it's worth adding a note about what these numbers are for?

will-v-pi: I've changed the examples to actually use feature groups, which I think shows what those numbers are for? It still uses arbitrary ones (0x1111 and 0/1), but shows their use

kilograham (approved):

PR #571, opened 2024-11-12
Add simple UART bootloader example

will-v-pi:
Add a UART bootloader example, which reads a binary from the partition table on the main device, and uses the UART bootloader to boot that binary on a separate device. Requires a modified separate board to run, as you need access to the QSPI pins to use them as UARTs for the bootloader (see 5.8 in the datasheet) - the main board can be a normal RP2350 board.

This sample binary flashes the LED and prints "Hello, world" back over the UART interface, which demonstrates setting up QSPI GPIOs as UARTs.

### DIFF
--- a/bootloaders/CMakeLists.txt
+++ b/bootloaders/CMakeLists.txt
@@ -1,3 +1,5 @@
+add_subdirectory_exclude_platforms(uart host rp2040)
+
 if (TARGET pico_mbedtls)
     # older clang seem to have a segment overlap issue that confuses picotool
     if (PICO_C_COMPILER_IS_CLANG AND CMAKE_C_COMPILER_VERSION VERSION_LESS "17.0.0")
--- a/bootloaders/uart/CMakeLists.txt
+++ b/bootloaders/uart/CMakeLists.txt
@@ -5,7 +5,10 @@ add_executable(uart_boot
 # pull in common dependencies
 target_link_libraries(uart_boot pico_stdlib hardware_flash)
 
+# add partition table
 pico_embed_pt_in_binary(uart_boot ${CMAKE_CURRENT_LIST_DIR}/uart-pt.json)
+
+# create absolute UF2
 pico_set_uf2_family(uart_boot "absolute")
 
 # create map/bin/hex file etc.
@@ -15,6 +18,7 @@ pico_add_extra_outputs(uart_boot)
 example_auto_set_url(uart_boot)
 
 
+# Create separate binary to be loaded onto other device
 add_executable(uart_binary
     uart_binary.c
     )
@@ -23,6 +27,8 @@ add_executable(uart_binary
 target_link_libraries(uart_binary pico_stdlib)
 
 pico_set_binary_type(uart_binary no_flash)
+
+# package uf2 in flash
 pico_package_uf2_output(uart_binary 0x10000000)
 
 # create map/bin/hex/uf2 file etc.
--- a/bootloaders/uart/CMakeLists.txt
+++ b/bootloaders/uart/CMakeLists.txt
@@ -1,0 +1,32 @@
+add_executable(uart_boot
+    uart_boot.c
+    )
+
+# pull in common dependencies
+target_link_libraries(uart_boot pico_stdlib hardware_flash)
+
+pico_embed_pt_in_binary(uart_boot ${CMAKE_CURRENT_LIST_DIR}/uart-pt.json)
+pico_set_uf2_family(uart_boot "absolute")
+
+# create map/bin/hex file etc.
+pico_add_extra_outputs(uart_boot)
+
+# add url via pico_set_program_url
+example_auto_set_url(uart_boot)
+
+
+add_executable(uart_binary
+    uart_binary.c
+    )
+
+# pull in common dependencies
+target_link_libraries(uart_binary pico_stdlib)
+
+pico_set_binary_type(uart_binary no_flash)
+pico_package_uf2_output(uart_binary 0x10000000)
+
+# create map/bin/hex/uf2 file etc.
+pico_add_extra_outputs(uart_binary)
+
+# call pico_set_program_url to set path to example on github, so users can find the source for an example via picotool
+example_auto_set_url(uart_binary)
--- a/bootloaders/uart/CMakeLists.txt
+++ b/bootloaders/uart/CMakeLists.txt
@@ -8,7 +8,7 @@ target_link_libraries(uart_boot pico_stdlib hardware_flash)
 # add partition table
 pico_embed_pt_in_binary(uart_boot ${CMAKE_CURRENT_LIST_DIR}/uart-pt.json)
 
-# create absolute UF2
+# create absolute UF2, as it's a bootloader so shouldn't go in a partition
 pico_set_uf2_family(uart_boot "absolute")
 
 # create map/bin/hex file etc.
--- a/bootloaders/uart/uart-pt.json
+++ b/bootloaders/uart/uart-pt.json
@@ -1,0 +1,23 @@
+{
+  "version": [1, 0],
+  "unpartitioned": {
+    "families": ["absolute"],
+    "permissions": {
+      "secure": "rw",
+      "nonsecure": "rw",
+      "bootloader": "rw"
+    }
+  },
+  "partitions": [
+    {
+      "start": "128K",
+      "size": "32K",
+      "families": ["rp2350-arm-s"],
+      "permissions": {
+        "secure": "rw",
+        "nonsecure": "rw",
+        "bootloader": "rw"
+      }
+    }
+  ]
+}
--- a/bootloaders/uart/uart-pt.json
+++ b/bootloaders/uart/uart-pt.json
@@ -12,7 +12,7 @@
     {
       "start": "128K",
       "size": "32K",
-      "families": ["rp2350-arm-s"],
+      "families": ["rp2350-arm-s", "rp2350-riscv"],
       "permissions": {
         "secure": "rw",
         "nonsecure": "rw",
--- a/bootloaders/uart/uart_binary.c
+++ b/bootloaders/uart/uart_binary.c
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "pico/stdlib.h"
+#include "hardware/uart.h"
+#include "hardware/structs/pads_qspi.h"
+#include "hardware/structs/io_qspi.h"
+
+#ifndef LED_DELAY_MS
+#define LED_DELAY_MS 500
+#endif
+
+// Initialize the GPIO for the LED
+void pico_led_init(void) {
+#ifdef PICO_DEFAULT_LED_PIN
+    // A device like Pico that uses a GPIO for the LED will define PICO_DEFAULT_LED_PIN
+    // so we can use normal GPIO functionality to turn the led on and off
+    gpio_init(PICO_DEFAULT_LED_PIN);
+    gpio_set_dir(PICO_DEFAULT_LED_PIN, GPIO_OUT);
+#endif
+}
+
+// Turn the LED on or off
+void pico_set_led(bool led_on) {
+#if defined(PICO_DEFAULT_LED_PIN)
+    // Just set the GPIO on or off
+    gpio_put(PICO_DEFAULT_LED_PIN, led_on);
+#endif
+}
+
+// Set function for QSPI GPIO pin
+void qspi_gpio_set_function(uint gpio, gpio_function_t fn) {
+    // Set input enable on, output disable off
+    hw_write_masked(&pads_qspi_hw->io[gpio],
+                   PADS_QSPI_GPIO_QSPI_SD2_IE_BITS,
+                   PADS_QSPI_GPIO_QSPI_SD2_IE_BITS | PADS_QSPI_GPIO_QSPI_SD2_OD_BITS
+    );
+    // Zero all fields apart from fsel; we want this IO to do what the peripheral tells it.
+    // This doesn't affect e.g. pullup/pulldown, as these are in pad controls.
+    io_qspi_hw->io[gpio].ctrl = fn << IO_QSPI_GPIO_QSPI_SD2_CTRL_FUNCSEL_LSB;
+
+    // Remove pad isolation now that the correct peripheral is in control of the pad
+    hw_clear_bits(&pads_qspi_hw->io[gpio], PADS_QSPI_GPIO_QSPI_SD2_ISO_BITS);
+}
+
+int main() {
+    pico_led_init();
+
+    // SD2 is QSPI GPIO 3, SD3 is QSPI GPIO 4
+    qspi_gpio_set_function(3, GPIO_FUNC_UART_AUX);
+    qspi_gpio_set_function(4, GPIO_FUNC_UART_AUX);
+
+    uart_init(uart0, 1000000);
+
+    while (true) {
+        uart_puts(uart0, "Hello, world\n");
+        pico_set_led(true);
+        sleep_ms(LED_DELAY_MS);
+        pico_set_led(false);
+        sleep_ms(LED_DELAY_MS);
+    }
+}
--- a/bootloaders/uart/uart_boot.c
+++ b/bootloaders/uart/uart_boot.c
@@ -8,7 +8,6 @@
 
 // UART defines for uart boot
 #define UART_ID uart1
-#define BAUD_RATE 1000000
 
 // Use pins 4 and 5 for uart boot
 #define UART_TX_PIN 4
@@ -108,7 +107,7 @@ int main()
     stdio_init_all();
 
     // Set up our UART for booting the other device
-    uart_init(UART_ID, BAUD_RATE);
+    uart_init(UART_ID, 1000000);
     gpio_set_function(UART_TX_PIN, GPIO_FUNC_UART);
     gpio_set_function(UART_RX_PIN, GPIO_FUNC_UART);
 
--- a/bootloaders/uart/uart_boot.c
+++ b/bootloaders/uart/uart_boot.c
@@ -1,0 +1,152 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "pico/stdlib.h"
+#include "hardware/uart.h"
+#include "pico/bootrom.h"
+#include "boot/picobin.h"
+#include "hardware/flash.h"
+
+// UART defines for uart boot
+#define UART_ID uart1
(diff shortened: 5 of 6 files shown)


bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,152 @@
...
+        while (uart_is_readable(UART_ID) && i < sizeof(buf)) {
+            char in = uart_getc(UART_ID);
+            printf("%c", in);
+            buf[i] = in;
+            i++;
+        }
+        if (i > 0) {
+            printf(" ...Read done\n");
+        }
+        char *ptr = memchr(buf, 'R', sizeof(buf));
```
lurch, line 137: Is the `'R'` here actually `splash[0]` ? (which might make it more obvious what the code is doing?)

lurch:
I wonder if things like the "splash string", the "knock sequence", the "uart-bootloader baudrate", etc. ought to be defined in a header-file somewhere in `pico-sdk`? (As I assume they'll always be the same for **every** RP2350 chip?)

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,152 @@
...
+        // Send the knock sequence
+        uart_putc_raw(UART_ID, 0x56);
+        uart_putc_raw(UART_ID, 0xff);
+        uart_putc_raw(UART_ID, 0x8b);
+        uart_putc_raw(UART_ID, 0xe4);
+        uart_putc_raw(UART_ID, 'n');
+
+        if (uart_is_readable_within_us(UART_ID, 1000)) {
+            char in = uart_getc(UART_ID);
+            printf("%c\n", in);
```
lurch, line 41: Should this check that `in == 'n'` ?

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,152 @@
...
+    uint32_t location_and_permissions = buffer[1];
+    uint32_t saddr = XIP_BASE + ((location_and_permissions >> PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_LSB) & 0x1fffu) * FLASH_SECTOR_SIZE;
+    uint32_t eaddr = XIP_BASE + (((location_and_permissions >> PICOBIN_PARTITION_LOCATION_LAST_SECTOR_LSB) & 0x1fffu) + 1) * FLASH_SECTOR_SIZE;
+    printf("Start %08x, end %08x\n", saddr, eaddr);
+
+    free(buffer);
+
+    printf("Writing binary\n");
+    uint32_t tstart = time_us_32();
+    uint32_t caddr = saddr;
```
lurch, line 72: Using more readable variable names like `start_addr`, `end_addr`, `current_addr` might make this more readable and slightly less cryptic?

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,152 @@
...
+    // Get partition location in flash
+    const int buf_words = (16 * 4) + 1;   // maximum of 16 partitions, each with maximum of 4 words returned, plus 1
+    uint32_t* buffer = malloc(buf_words * 4);
+
+    int ret = rom_get_partition_table_info(buffer, buf_words, PT_INFO_PARTITION_LOCATION_AND_FLAGS | PT_INFO_SINGLE_PARTITION | (0 << 24));
+    assert(buffer[0] == (PT_INFO_PARTITION_LOCATION_AND_FLAGS | PT_INFO_SINGLE_PARTITION));
+    assert(ret == 3);
+
+    uint32_t location_and_permissions = buffer[1];
+    uint32_t saddr = XIP_BASE + ((location_and_permissions >> PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_LSB) & 0x1fffu) * FLASH_SECTOR_SIZE;
```
lurch, line 64: Where does this `0x1fffu` magic-number come from?

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,152 @@
...
+
+    printf("Boot starting\n");
+
+    // Get partition location in flash
+    const int buf_words = (16 * 4) + 1;   // maximum of 16 partitions, each with maximum of 4 words returned, plus 1
+    uint32_t* buffer = malloc(buf_words * 4);
+
+    int ret = rom_get_partition_table_info(buffer, buf_words, PT_INFO_PARTITION_LOCATION_AND_FLAGS | PT_INFO_SINGLE_PARTITION | (0 << 24));
+    assert(buffer[0] == (PT_INFO_PARTITION_LOCATION_AND_FLAGS | PT_INFO_SINGLE_PARTITION));
+    assert(ret == 3);
```
lurch, line 61: What does the magic value 3 indicate? :wink:

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    printf("Boot starting\n");
+
+    // Get partition location in flash
+    const int buf_words = (16 * 4) + 1;   // maximum of 16 partitions, each with maximum of 4 words returned, plus 1
+    uint32_t* buffer = malloc(buf_words * 4);
+
+    int ret = rom_get_partition_table_info(buffer, buf_words, PT_INFO_PARTITION_LOCATION_AND_FLAGS | PT_INFO_SINGLE_PARTITION | (0 << 24));
```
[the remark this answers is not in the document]
will-v-pi: I'm not sure what's gone wrong, but it looks like none of the RP2350 functions are included in that documentation - the doxygen is all there, but maybe the `#if !PICO_RP2040` halfway down the src/rp2_common/pico_bootrom/include/pico/bootrom.h file has something to do with it?

bootloaders/uart/uart_boot.c
will-v-pi: This one isn't actually a magic value - it's just the number of words written in the buffer (which should be 3 here - 1 word for the returned flags, and 2 words for the partition_location_and_flags)

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,152 @@
...
+#include <stdlib.h>
+#include "pico/stdlib.h"
+#include "hardware/uart.h"
+#include "pico/bootrom.h"
+#include "boot/picobin.h"
+#include "hardware/flash.h"
+
+// UART defines for uart boot
+#define UART_ID uart1
+#define BAUD_RATE 1000000
```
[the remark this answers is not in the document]
lurch: Yeah, I was assuming that in order to accommodate my suggestion, the code (in the main device) would need to dynamically reconfigure the baud rate at the start and end of the uart-boot sequence :slightly_smiling_face:

bootloaders/uart/uart_boot.c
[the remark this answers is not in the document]
lurch: I'll have a fiddle later... `rom_get_partition_table_info` isn't appearing in https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf either! :scream:

bootloaders/uart/uart_boot.c
lurch: I guess it might be useful if https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_bootrom/include/pico/bootrom.h#L628 documented the return value.

will-v-pi: It does mention it https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_bootrom/include/pico/bootrom.h#L634

bootloaders/uart/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+add_executable(uart_binary
+    uart_binary.c
+    )
+
+# pull in common dependencies
+target_link_libraries(uart_binary pico_stdlib)
+
+pico_set_binary_type(uart_binary no_flash)
+pico_package_uf2_output(uart_binary 0x10000000)
```
lurch, line 26: I assume that it's **this** line that puts the `uart_binary` into the partition? Seems quite opaque, and took me quite a while to figure that out :confused: 

Sorry to keep banging on about "magic numbers" but there's another one here - would be nice to at least have a comment indicating what this `0x10000000` means.

bootloaders/uart/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+# create map/bin/hex/uf2 file etc.
+pico_add_extra_outputs(uart_binary)
+
+# call pico_set_program_url to set path to example on github, so users can find the source for an example via picotool
+example_auto_set_url(uart_binary)
```
lurch, line 32: Will this URL also be visible in the final UF2, or is it only the `uart_boot` URL that'll be visible to `picotool` ?

bootloaders/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+add_subdirectory_exclude_platforms(uart host rp2040)
```
lurch, line 1: I guess it's the use of "partitions" (rather than anything else) that means that the `bootloaders/uart` example won't work on RP2040? But in theory at least (if you used some kind of other "enclosed binary" mechanism), there's no reason that RP2040 couldn't UART-boot an RP2350?

bootloaders/uart/CMakeLists.txt
will-v-pi: This doesn't put it in the partition - this just puts it in a packaged SRAM UF2 (datasheet 5.10.3), which when dragged & dropped onto a device will be written to flash rather than SRAM. If you've already dragged & dropped the `uart_boot.uf2` file which contains an embedded partition table, then this `uart_binary.uf2` will be written to the first partition in the Flash, rather than the start of Flash (which is where it would go if there's no partition table on the device).

Ie start with clean device, drag & drop `uart_boot.uf2`, put back in bootsel, then drag & drop `uart_binary.uf2`

bootloaders/CMakeLists.txt
will-v-pi: Yes, that's correct

bootloaders/uart/CMakeLists.txt
will-v-pi: It'll be visible in the `uart_binary.uf2` file, and will be visible to `picotool` once on the device

bootloaders/uart/uart_boot.c
lurch: Oh, I missed that :laughing:  My eyes just jumped straight to the bottom of the doxygen description looking for the `\return ...` line :stuck_out_tongue_winking_eye:

bootloaders/uart/CMakeLists.txt
lurch: > Ie start with clean device, drag & drop `uart_boot.uf2`, put back in bootsel, then drag & drop `uart_binary.uf2`

Ohhhh!! That wasn't clear at all - if this is the case then IMHO there definitely needs to be a `README.md` for this example explaining how to use it. My (naive) assumption was that when you built this example you'd just end up with a single UF2 image which contained both the `uart_boot` and `uart_binary` parts in different partitions. (I haven't yet got round to reading up about RP2350 partitions, etc.)

bootloaders/uart/CMakeLists.txt
lurch: (I asked this because I didn't realise that there'd actually be two separate UF2 files :man_facepalming: )

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,151 @@
...
+            return;
+        }
+        char in = uart_getc(UART_ID);
+        printf("%c\n", in);
+        caddr += 32;
+    }
+
+    uint32_t tend = time_us_32();
+    printf("Write took %dus\n", tend - tstart);
+    printf("Write complete - executing\n");
```
lurch, line 92: Before starting the execute, should we also do the read-back-and-verify as described in step 8 in section 5.8.3 of the RP2350 datasheet? (I guess it probably makes sense for our official examples to use the "best practice" approach?)

bootloaders/uart/uart-pt.json
```diff
@@ -0,0 +1,23 @@
...
+    "permissions": {
+      "secure": "rw",
+      "nonsecure": "rw",
+      "bootloader": "rw"
+    }
+  },
+  "partitions": [
+    {
+      "start": "128K",
+      "size": "32K",
```
lurch, line 14: If this is intended as a "generic UART bootloader", should we increase the size of this partition to allow for bigger UF2s?

bootloaders/uart/uart-pt.json
```diff
@@ -0,0 +1,23 @@
...
+      "families": ["rp2350-arm-s"],
```
lurch, line 15: Could this bootloader theoretically also boot the other RP2350 with a RISC-V UF2, or is this limited to Arm UF2s only?

bootloaders/uart/uart_boot.c
will-v-pi: Yep - fixed

bootloaders/uart/uart_boot.c
will-v-pi: Yep - fixed

bootloaders/uart/uart_boot.c
will-v-pi: It's `PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_BITS` - fixed to do the `&` and `>>` the other way round to use `PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_BITS` and `PICOBIN_PARTITION_LOCATION_LAST_SECTOR_BITS` in correct places

bootloaders/uart/uart-pt.json
will-v-pi: The issue is that this simple bootloader just loads the whole partition onto the other device, so the larger the partition the longer it takes. From a quick play it takes 0.35s for this 32kiB partition, vs 5.6s for a full 512kiB partition (which would be the maximum size, as that fills the RP2350 SRAM)

bootloaders/uart/uart-pt.json
will-v-pi: Yep, it all works in RISC-V too - I'll fix that

bootloaders/uart/uart_boot.c
will-v-pi: Added

bootloaders/uart/uart_boot.c
will-v-pi: Yep - I've added checks that will call `reset_chip()` if it's the incorrect character for all these read backs

bootloaders/uart/uart-pt.json
lurch: Ah, that makes sense. I guess in _theory_ a more complicated UART bootloader could look at the IMAGE_DEF and only copy the part of the partition that it needed to?

I've not played with partitions on the RP2350 yet, but what happens if you try drag'n'dropping a 128kiB image into a partition that's only 32kiB big?

bootloaders/uart/uart_boot.c
```diff
@@ -83,25 +88,84 @@ void uart_boot() {
...
+        if (i != 32) {
+            printf("Verify has incorrect data size - resetting\n");
+        }
+        in = uart_getc(UART_ID);
+        if (in != 'r') {
+            printf("Incorrect response - resetting\n");
+            reset_chip();
+            return;
+        }
+        current_addr += 32;
```
lurch, line 148: Or I guess you could do `+= i` :wink:

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,218 @@
...
+        }
+        if (i > 0) {
+            printf(" ...Read done\n");
+        }
+        char *ptr = memchr(buf, splash[0], sizeof(buf));
+        if (ptr && strncmp(ptr, splash, sizeof(splash) - 1) == 0) {
+            printf("Splash found\n");
+            uart_boot();
+        } else {
+            ptr = memchr(buf, hello[0], sizeof(buf));
```
lurch, line 208: Hah, I guess it's important that `hello[0]` doesn't appear anywhere in the `splash` string :wink: 

Is there any reason not to have this check for the full "Hello, world\n" that `uart_binary` transmits?

bootloaders/uart/uart_boot.c
```diff
@@ -0,0 +1,218 @@
...
+    gpio_init(RUN_PIN);
+    gpio_set_dir(RUN_PIN, GPIO_OUT);
+
+    // Reset chip
+    reset_chip();
+
+
+    while (true) {
+        char splash[] = "RP2350";
+        char hello[] = "Hello";
```
lurch, line 191: minor nitpick: I guess these two variables could be moved outside of the `while (true)` loop?

bootloaders/uart/uart_binary.c
```diff
@@ -0,0 +1,65 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "pico/stdlib.h"
+#include "hardware/uart.h"
+#include "hardware/structs/pads_qspi.h"
+#include "hardware/structs/io_qspi.h"
+
+#ifndef LED_DELAY_MS
+#define LED_DELAY_MS 500
```
lurch, line 13: From what I understand, I think the "parent" RP2350 will reset the "child" RP2350 if it doesn't receive the string "Hello" once per second; so I guess that means that `LED_DELAY_MS` **can't** be larger than 500?

peterharperuk:
This worked well for me once I'd figured out that the uart binary had to be loaded into the "master" device first. It would be more fun if the binary could be built into a header file to avoid that step

bootloaders/uart/uart_boot.c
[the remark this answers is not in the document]
kilograham: is this doc issue fixed?

kilograham (dismissed):

bootloaders/uart/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    uart_boot.c
+    )
+
+# pull in common dependencies
+target_link_libraries(uart_boot pico_stdlib hardware_flash)
+
+# add partition table
+pico_embed_pt_in_binary(uart_boot ${CMAKE_CURRENT_LIST_DIR}/uart-pt.json)
+
+# create absolute UF2
```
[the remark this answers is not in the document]
will-v-pi: Added - it's because it's a bootloader, so shouldn't go in a partition

bootloaders/uart/uart_boot.c
will-v-pi: I think it would still work fine if `hello[0]` was in the splash string, because it's only working with 1 line of output at a time - or if it has multiple lines, then the new attempts thing will make it still work. Only issue would be if the `splash` string appeared in `hello`, which would then cause problems.

kilograham (approved):

PR #574, opened 2024-11-19
Ws2812 example

peterharperuk:


kilograham (approved):

pio/ws2812/ws2812.c
```diff
@@ -105,12 +105,17 @@ const struct {
 int main() {
     //set_sys_clock_48();
     stdio_init_all();
-    printf("WS2812 Smoke Test, using pin %d", WS2812_PIN);
+    printf("WS2812 Smoke Test, using pin %d\n", WS2812_PIN);
 
     // todo get free sm
```
lurch, line 110: I guess this 'todo' comment can be removed now? :slightly_smiling_face:

pio/ws2812/ws2812.c
```diff
@@ -54,44 +54,44 @@ static inline uint32_t urgbw_u32(uint8_t r, uint8_t g, uint8_t b, uint8_t w) {
...
     t %= max;
     for (uint i = 0; i < len; ++i) {
-        put_pixel(t * 0x10101);
+        put_pixel(pio, sm, t * 0x10101);
         if (++t >= max) t = 0;
     }
 }
 
-typedef void (*pattern)(uint len, uint t);
+typedef void (*pattern)(PIO pio, uint sm, uint len, uint t);
```
lurch, line 94: I'm 99% sure that it doesn't matter, but just making an observation here that this now means that `ws2812_parallel.c` now has a different typedef
```c
typedef void (*pattern)(uint len, uint t);
```
to the one in `ws2812.c`.

peterharperuk: None of the functions reference the pio or sm in ws2812_parallel.c because they're using dma

PR #523, opened 2024-08-20
Update the auto-generated ws2812.pio.h to match changes made in #486

lurch:

PR #576, opened 2024-11-22
Update ws2812 generated file

j3270:
The file pico-examples/pio/ws2812/generated/ws2812.pio.h appears to be generated.

This PR stops tracking the file and adds it to .gitignore.


lurch:
I think I asked about this once, and I was told that it was checked-in deliberately? :thinking: 
So see also #523

peterharperuk:
@lurch What was the reason? I keep forgetting to fix this and my approach would be to move generated files to the build tree.

j3270:
Hello,

I only noticed this because I brought the sdk and examples repos into my project repo as submodules.  

I realized that wasn't the best approach and have now just created a 'workspace' folder where I have cloned the sdk, examples, picotool, and freeRTOS.  My project will be another dir in the workspace repo and the modified file in the examples doesn't bother me as much. 

I figured just adding the file to .gitignore was a little sledge hammerish and there was probably a better way using CMake.  I would be happy to work on that, since I have the PR open, but not sure if I understand enough about CMake to do so.  Any pointers to get me going in the right direction would be appreciated.

Thanks,
Justin

kilograham:
> @lurch What was the reason? I keep forgetting to fix this and my approach would be to move generated files to the build tree.

These files are referenced by datasheet and other documentation so must exist. Strictly they don't need to be generated there by the actual build, but never had any problems before (though i could see an issue with multiple concurrent builds)

They should generate as they were; if not, then we should make sure the checked in versions are correct

j3270:
Nope, building for pico2 generates the same output as pico, and pico_w.

It appears that what is checked in for that file is just different than what is generated for all three.

I can update this branch to track the file and check in the generated if you guys would like me to.

peterharperuk:
Yes, if you post the differences as a commit we can see what it is. For some reason it's not happening for me although I think I've seen this problem in the past.

j3270:
I haven't actually run the example with that update, I will do later, have to run.

kilograham:
thx - in parallel i updated the generated/ file as i had it locally and was pushing some github action stuff... this PR will probably be empty if you rebase

j3270:
@kilograham done.  I hope closing this PR was the appropriate thing to do.  

Thanks

PR #579, opened 2024-11-22
Update mpu6050_i2c.c

abdalla1912mohamed:
Explanation of Changes:
The first part of the function resets the MPU6050 by writing 0x80 to the 0x6B register, which is the reset command. After a short delay (100 ms) to allow the reset to take effect, the second part writes 0x00 to the 0x6B register to clear the sleep mode and wake up the device. A short 10 ms delay is added after clearing the sleep mode to allow the device to stabilize.


peterharperuk (approved):
Yes - I believe this example is currently broken and this change fixes it.

kilograham (approved):

PR #581, opened 2024-11-23
Fix pio headers that are used in SDK docs

peterharperuk:
The PIO version in no longer hard coded.


peterharperuk:
@kilograham I thought you'd fixed this but I can't see it - so here it is again.

kilograham (approved):

Issue #584, opened 2024-11-29
Encrypted bootloader example does not work with stdio over USB


eliasbakken:
The bootloader/encrypted example works as expected when compiling it as is, data is transferred via uart. But when stdio is enabled for USB, there is only output when the enc_bootloader is run, not the final program. 

Adding `pico_enable_stdio_usb(hello_serial_enc 1)` to the CMakeLists.txt does not render any output via USB.

lurch:
ping @will-v-pi

will-v-pi:
This should be fixed when using picotool 2.1.0 (which is used by SDK 2.1.0) - see #558 and https://github.com/raspberrypi/picotool/pull/150

Could you try this with SDK & picotool 2.1.0 and see if you still have this issue?

eliasbakken:
Thank you for the suggestion. I did try it, but I do not see any change on the output of the USB.

eliasbakken:
It looks as though interrupts are not working when running from SRAM. 
The following program will not fire when loaded with the decoding bootloader, but will fire if run from flash directly. 
```
#include <stdio.h>
#include "pico/stdlib.h"

void gpio_callback(uint gpio, uint32_t events) {
    printf("GPIO %d\n", gpio);
}

int main() {
    stdio_uart_init();
    gpio_init(2);
    gpio_set_dir(2, GPIO_IN);
    gpio_set_irq_enabled_with_callback(2, GPIO_IRQ_EDGE_RISE | GPIO_IRQ_EDGE_FALL, true, &gpio_callback);

    while(true);
}
```

will-v-pi:
It looks like the issue is the `save_and_disable_interrupts` call in `rom_chain_image` in bootrom.h which seems to disable interrupts for the encrypted binary too.

This can be fixed by calling `restore_interrupts_from_disabled(0);` (and including `#include "hardware/sync.h"`) before calling `stdio_init_all()` in the encrypted binary.

```c
/**
 * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
 *
 * SPDX-License-Identifier: BSD-3-Clause
 */

#include <stdio.h>
#include "pico/stdlib.h"
#include "hardware/sync.h"

int main() {
    restore_interrupts_from_disabled(0);
    stdio_init_all();
    while (true) {
        printf("Hello, world!\n");
        printf("I'm an encrypted binary\n");
        sleep_ms(1000);
    }
}
```

[...]

PR #588, opened 2024-12-17
Fix using stdio_usb and other interrupts from encrypted binaries

will-v-pi:
This is a fix for #584, to enable the interrupts in the encrypted binary, which then allows use of `stdio_usb` if desired

### DIFF
--- a/bootloaders/encrypted/hello_serial.c
+++ b/bootloaders/encrypted/hello_serial.c
@@ -9,7 +9,7 @@
 #include "hardware/sync.h"
 
 int main() {
-    restore_interrupts_from_disabled(0);
+    enable_interrupts();
     stdio_init_all();
     while (true) {
         printf("Hello, world!\n");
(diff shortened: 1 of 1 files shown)


bootloaders/encrypted/hello_serial.c
```diff
@@ -6,8 +6,10 @@
 
 #include <stdio.h>
 #include "pico/stdlib.h"
+#include "hardware/sync.h"
 
 int main() {
+    restore_interrupts_from_disabled(0);
```
kilograham, line 12: i was gonna say we should explicitly call enable IRQs as the parameter here is platform specific and thus confusing; I will add an issue to make an `enable_interrupts()` method in the SDK

kilograham: https://github.com/raspberrypi/pico-sdk/issues/2262

kilograham:
ah yes, @will-v-pi we should probably do this in the regular binary startup as you commented elsewhere (`runtime_init_per_core_irq_priorities` for simplificity)

PR #595, opened 2025-01-08
Added include guard to onewire_library.h

gumpump:
Hello :)

I'm using the onewire library of this example in one of my own projects. There I noticed the missing include guard, which would prevent the multiple definition of the functions while compiling.

I think this is code users would like to use in their own projects, so the added include guard would prevent preventable problems.

This is my first pull request, please be kind :)


lurch:
ping @mjcross who added this file in #334

mjcross:
@lurch no objections
@gumpump thanks for taking the time to contribute :-)

peterharperuk (approved):

PR #603, opened 2025-01-25
feat(cherryusb): add cherryusb device & host demo

sakumisu:
https://github.com/raspberrypi/pico-sdk/pull/2212


usb/cherryusb/host/host_cdc_msc_hid/FreeRTOSConfig.h
```diff
@@ -0,0 +1,7 @@
+#ifndef FREERTOS_CONFIG_H
+#define FREERTOS_CONFIG_H
+
+// This example uses a common include to avoid repetition
+#include "FreeRTOSConfig_examples_common.h"
```
lurch, line 5: This should probably just include https://github.com/raspberrypi/pico-examples/blob/develop/freertos/FreeRTOSConfig_examples_common.h rather than adding a separate copy of the file to the repo?

usb/cherryusb/host/host_cdc_msc_hid/main.c
```diff
@@ -0,0 +1,23 @@
...
+#include <string.h>
+#include "FreeRTOS.h"
+#include "task.h"
+
+#include "usb_host.c"
+
+int main(void)
+{
+    stdio_init_all();
+    printf("CherryUSB host cdc msc example\n");
```
lurch, line 14: Apologies for the stupid question, but what does this example actually do?

usb/cherryusb/host/host_cdc_msc_hid/usb_config.h
```diff
@@ -0,0 +1,295 @@
+/*
```
lurch, line 1: It seems that this `usb_config.h` is identical to the other `usb_config.h` - perhaps it makes sense to add just one copy of the file to the repo (in `usb/cherryusb/` ?), and have both examples share it? :thinking: 
(i.e. like what is done with https://github.com/raspberrypi/pico-examples/blob/develop/freertos/FreeRTOSConfig_examples_common.h and https://github.com/raspberrypi/pico-examples/blob/develop/pico_w/wifi/lwipopts_examples_common.h and https://github.com/raspberrypi/pico-examples/blob/develop/pico_w/wifi/mbedtls_config_examples_common.h )

sakumisu: This is an optional config for users to modify.

usb/cherryusb/host/host_cdc_msc_hid/FreeRTOSConfig.h
sakumisu: https://github.com/raspberrypi/pico-examples/tree/master/freertos i use this template, freertos uses FreeRTOSConfig.h not FreeRTOSConfig_examples_common.h, so i don't know why add FreeRTOSConfig_examples_common.h.

usb/cherryusb/host/host_cdc_msc_hid/main.c
sakumisu: Actually our class driver are all automatic loaded without call some apis, we include usb_host.c which has cdc + msc + hid transfer demo.

usb/cherryusb/host/host_cdc_msc_hid/usb_config.h
sakumisu: Like enable log, enable bufsize for performance, enable more class instances and so on.

usb/cherryusb/host/host_cdc_msc_hid/FreeRTOSConfig.h
lurch: I think maybe you could just do `#include ../../../freertos/FreeRTOSConfig_examples_common.h` in your `FreeRTOSConfig.h`? (although perhaps there's a "nicer" way of doing that without all the `../` ?) And then you could remove `FreeRTOSConfig_examples_common.h` from this PR. :shrug:

usb/cherryusb/host/host_cdc_msc_hid/main.c
sakumisu: In usb_host.c, when class enums done and go to class callback, then we will create a thread to do transaction, so in main.c we only use usbh_init api. For other class drivers are the same.

usb/cherryusb/host/host_cdc_msc_hid/FreeRTOSConfig.h
sakumisu: How about using FreeRTOSConfig.h only? If configs in FreeRTOSConfig_examples_common.h have conflict with other demo like less memory, no feature support like sem & mutex disabled, others will be confused.

usb/cherryusb/host/host_cdc_msc_hid/usb_config.h
lurch: Still, _perhaps_ it makes sense to have a per-example `usb_config.h` (where the users can add their optional customisation) which then does `#include "usb_config_examples_common.h"` ? (just to reduce the amount of duplication)

sakumisu: > Still, _perhaps_ it makes sense to have a per-example `usb_config.h` (where the users can add their optional customisation) which then does `#include "usb_config_examples_common.h"` ? (just to reduce the amount of duplication)

I know what you mean, yeah, reduce the duplication, also freertos. I will modify.

usb/cherryusb/host/host_cdc_msc_hid/FreeRTOSConfig.h
lurch: Hmmm I just checked and I hadn't realised that `FreeRTOSConfig_examples_common.h` doesn't use the normal
```c
#ifndef SOME_SETTING
#define SOME_SETTING some_value
#endif
```
pattern. But @peterharperuk understands all of this much better than I do! :laughing:

usb/cherryusb/host/host_cdc_msc_hid/usb_config.h
sakumisu: Have done. And i see pico_w demo has `FreeRTOSConfig_examples_common.h`, so i move this into config dir.

lurch: > Have done.

Thanks :+1:  But I'm still not sure if it makes sense to have both `freertos/FreeRTOSConfig_examples_common.h` and  `usb/cherryusb/config/FreeRTOSConfig_examples_common.h`, which appear to be identical copies of each other?

usb/cherryusb/host/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+if (NOT FREERTOS_KERNEL_PATH AND NOT DEFINED ENV{FREERTOS_KERNEL_PATH})
+    message("Skipping FreeRTOS examples as FREERTOS_KERNEL_PATH not defined")
```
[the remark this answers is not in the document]
sakumisu: OK

PR #608, opened 2025-02-16
Added Dockerfile and Readme instructions to build in a container

GeoffreyPlitt:
This PR adds a Dockerfile, and README instructions, to do builds inside a Docker container.

This makes it easier to build on almost any machine, without installing complex build toolchains. On Mac Mini M1 Sonoma 14.4, I found that the Homebrew ARM toolchain wasn't working. But this PR will benefit almost anybody, on any machine, trying to build this.


lurch:
Rather than using Homebrew, does https://github.com/raspberrypi/pico-vscode work for you? That's now intended to be our "low friction" approach for new users. Ping @will-v-pi in case he wants to add any additional context.

README.md
```diff
@@ -18,6 +18,33 @@ For RISC-V compilation, you should take a compiler from here: https://www.embeco
...
+
+# Build Docker image (default is pico if --build-arg not specified)
+docker build -t pico-examples --build-arg PICO_BOARD=pico .
+
+# Run container for file copying
+docker run -d --name pico-build pico-examples sleep infinity
+
+# Copy UF2 files from container to host
+docker container exec pico-build mkdir -p /workspace/pico-examples/build/uf2files
+docker container exec pico-build find /workspace/pico-examples/build -name "*.uf2" -exec cp {} /workspace/pico-examples/build/uf2files \;
```
lurch, line 40: Presumably this wouldn't work as expected if any UF2 files in different directories have the same filename? :wink:

will-v-pi:
Yes, I don’t think we’d want to merge and support this Dockerfile - the pico-vscode extension is our recommended low friction way to compile which should work on all 64-bit platforms.

GeoffreyPlitt:
Darn, that's a shame. IDEs come and go, but this solution works even for engineers who don't/can't use an IDE, and works in automation (CI/CD), etc.

lurch:
Of course there's nothing stopping you (or any other engineers) using this Dockerfile in your own projects; we just won't be including it in the official upstream pico-examples repo. (because the more different options we offer people, the harder it becomes for new users to determine which is the "right" option for them)

PR #610, opened 2025-02-18
ws2812 fixes

balika011:


pio/ws2812/ws2812.c
```diff
@@ -123,7 +144,11 @@ int main() {
     bool success = pio_claim_free_sm_and_add_program_for_gpio_range(&ws2812_program, &pio, &sm, &offset, WS2812_PIN, 1, true);
     hard_assert(success);
 
+#if IS_800KHz
     ws2812_program_init(pio, sm, offset, WS2812_PIN, 800000, IS_RGBW);
```
lurch, line 148: Perhaps it makes more sense to add e.g. `#define WS2812_FREQUENCY 800000` and then pass that to `ws2812_program_init`, rather than adding a `#if IS_800KHz` section?

pio/ws2812/ws2812.c
```diff
@@ -17,15 +17,19 @@
  *  Take into consideration if your WS2812 is a RGB or RGBW variant.
  *
  *  If it is RGBW, you need to set IS_RGBW to true and provide 4 bytes per 
- *  pixel (Red, Green, Blue, White) and use urgbw_u32().
+ *  pixel (Red, Green, Blue, White) and use urgb_u32() or urgbw_u32().
```
lurch, line 20: I can't tell why you've effectively removed support for `urgbw_u32()` ?

PR #616, opened 2025-02-20
Add COAP example

peterharperuk:
Adds a COAP client and server example.
Requires libcoap and tinydtls.


peterharperuk:
I think the coap server is a bad idea. Needs better host server example? Would prefer to use mbedtls but libcoap isn't working mbedtls with keys.

PR #619, opened 2025-02-24
Encrypted improvements

### LINKED ISSUE #613: otp.json in encrypted example breaks clean build
The file is generated over itself, and clean removes it... the build fails without it there; it would be better to generate it to elsewhere (also, and I didn't look, this could indicate a stale input style problem)

will-v-pi:
This adds a self-decrypting binary to demonstrate https://github.com/raspberrypi/pico-sdk/pull/2315 and https://github.com/raspberrypi/picotool/pull/207

Also fixes #613 in passing, as that is now handled by picotool.


bootloaders/encrypted/README.md
```diff
@@ -15,6 +15,16 @@ or in Powershell 7
...
 
+The IV salt is just a 16 byte binary file - you can create it the same way, replacing `128` with `16` and `privateaes.bin` with `ivsalt.bin` in the commands above.
+
+You will need to program your OTP using the `otp.json` file generated by the build in your build folder
+NOTE: This will enable secure boot on your device, so only correctly signed binaries can then run, and will also lock down the OTP pages the AES key and IV salt are stored in.
+```bash
+picotool otp load otp.json
+```
+
+> For more information on security see chapter 10 of the [RP2350 datasheet](https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf), and for information on how to sign other binaries to run on a secure chip see chapter 5.10
```
lurch, line 26: "chapter 5.10" -> "section 5.10"?

bootloaders/encrypted/README.md
```diff
@@ -1,4 +1,4 @@
-Replace private.pem and privateaes.bin with your own keys - your signing key must be for the _secp256k1_ curve, in PEM format. You can create a .PEM file with:
+For security you **must** replace private.pem and privateaes.bin with your own keys, and ivsalt.bin with your own per-device salt. Your signing key must be for the _secp256k1_ curve, in PEM format. You can create a .PEM file with:
```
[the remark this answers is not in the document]
will-v-pi: All 3 need to be kept if you want to compile new firmware for the device in the future using the same OTP pages for the AES key.

Also, I might add that the AES key should also be per-device for added security.

bootloaders/encrypted/README.md
```diff
@@ -1,4 +1,6 @@
-For security you **must** replace private.pem and privateaes.bin with your own keys, and ivsalt.bin with your own per-device salt. Your signing key must be for the _secp256k1_ curve, in PEM format. You can create a .PEM file with:
+For security you **must** replace private.pem and privateaes.bin with your own keys, and ivsalt.bin with your own per-device salt. The AES key should also be different for each device. Make sure you **don't lose your keys and salts**, else you may not be able to boot code on your device.
```
lurch, line 1: "boot code on your device" -> "update the code on your device" ?
(since I guess losing the keys won't stop the device booting from the code that's already in Flash!)

encrypted/hello_encrypted/hello_encrypted.c
```diff
@@ -15,10 +15,29 @@ int main() {
...
+    int ret = rom_get_boot_info(&boot_info);
+    if (ret) {
+        // BOOT_TBYB_AND_UPDATE_FLAG_BUY_PENDING will always be set, but check anyway
+        if (boot_info.tbyb_and_update_info & BOOT_TBYB_AND_UPDATE_FLAG_BUY_PENDING) {
+            // Need to check flash_update_base is set to see if this is a TBYB update
+            uint32_t flash_update_base = boot_info.reboot_params[0];
+            if (flash_update_base) {
+                printf("Perform self-check... ");
+                if (1 == 1) {
+                    printf("passed\n");
```
lurch, line 28: ???

lurch: Okay, perhaps it's worth adding a comment clarifying that then?

README.md
```diff
@@ -86,6 +86,12 @@ App|Description
 [channel_irq](dma/channel_irq) | Use an IRQ handler to reconfigure a DMA channel, in order to continuously drive data through a PIO state machine.
 [sniff_crc](dma/sniff_crc) | Use the DMA engine's 'sniff' capability to calculate a CRC32 on a data buffer.
 
+### Encrypted
```
lurch, line 89: I guess this ought to be suffixed with "(RP235x Only)" ?

bootloaders/encrypted/README.md
```diff
@@ -1,15 +1,32 @@
...
 ```bash
 openssl ecparam -name secp256k1 -genkey -out private.pem
 ```
 
-The AES key is just be a 32 byte binary file - you can create one with
+The AES key is stored as a 4-way share in a 128 byte binary file - you can create one with
+
+```bash
+dd if=/dev/urandom of=privateaes.bin bs=1 count=128
+```
```
lurch, line 13: I wonder if it might be worth changing the `privateaes.bin` filename to make it clearer that it's a key-share and not just a raw AES key? :thinking:

bootloaders/encrypted/aes.S
```diff
@@ -1,45 +1,31 @@
+/* MEMORY LAYOUT ASSUMPTIONS
```
lurch, line 1: Is this `aes.S` the same as the one in https://github.com/raspberrypi/picotool/pull/207 ?
(And if so, do we have any mechanism for ensuring that they stay in sync?)

encrypted/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+if (TARGET pico_mbedtls)
+    add_subdirectory_exclude_platforms(hello_encrypted host rp2040 rp2350-riscv)
+else()
+    # Assume picotool has no signing support, if no pico_mbedtls available
```
lurch, line 4: Although I guess that picotool _might_ have signing support if the user is using the prebuilt binary from pico-sdk-tools? :wink: 

Also, if mbedtls is the "faster but less secure" version, does that mean that this `encrypted` example might still work **without** pico_mbedtls if the user chooses to use the "slower but more secure" non-mbedtls version? (But that might be too much of an edge case to be worth considering!?)

encrypted/hello_encrypted/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+# Example encrypted binard
```
lurch, line 1: binard -> binary ?

encrypted/CMakeLists.txt
will-v-pi: > Also, if mbedtls is the "faster but less secure" version, does that mean that this `encrypted` example might still work **without** pico_mbedtls if the user chooses to use the "slower but more secure" non-mbedtls version? (But that might be too much of an edge case to be worth considering!?)

This is separate - picotool always requires mbedtls for signing/encryptiong - the new encrypted bootloader has the option to use pico_mbedtls for **decryption** instead of the `aes.S` code, but still requires picotool compiled with mbedtls for the encryption.

I think it's simplest to just require the user to checkout the mbedtls SDK submodule if they want to compile this example, rather than attempting to interrogate the picotool being used for encryption support (which may not even be compiled at this stage)

encrypted/hello_encrypted/hello_encrypted.c
lurch: Also, would it be safer / more idiomatic to move the `rom_explicit_buy` part to be _inside_ this "fake self-check" if-block?

will-v-pi: > Also, would it be safer / more idiomatic to move the `rom_explicit_buy` part to be _inside_ this "fake self-check" if-block?

No, you need to call `rom_explicit_buy` on every boot due to TBYB intricacies with self-decrypting binaries (which is why I put this sample code in to show what to do), but perform the self check beforehand if there's also a `flash_update_base` - see the picotool PR for an explanation of the intricacies, I put it in a comment somewhere

encrypted/hello_encrypted/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+add_executable(hello_encrypted
+        hello_encrypted.c
+        secret.S
+        )
+
+# include secret.txt
+target_include_directories(hello_encrypted PRIVATE ${CMAKE_CURRENT_LIST_DIR})
```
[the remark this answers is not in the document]
will-v-pi: You need to include the directory containing secret.txt, which is `CMAKE_CURRENT_LIST_DIR` - this is similar to the pico_w examples which need lwipopts eg
```cmake
target_include_directories(picow_tcpip_client_poll PRIVATE
        ${CMAKE_CURRENT_LIST_DIR}
        ${CMAKE_CURRENT_LIST_DIR}/.. # for our common lwipopts
        )
```

encrypted/CMakeLists.txt
lurch: Apologies for my confusion.

bootloaders/encrypted/aes.S
lurch: If the picotool CI is now also building pico-examples, perhaps the picotool CI could also check that these files in the two repos match? (or perhaps it's very unlikely that this file will change in future, so perhaps that'd be overkill)

will-v-pi: That sounds like a good idea, I will add that

encrypted/hello_encrypted/CMakeLists.txt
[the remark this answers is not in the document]
will-v-pi: I'll make it `# include directory containing secret.txt` - makes it clear what it's doing, for users to put secrets in different directories

PR #627, opened 2025-03-17
Add a sample to read the partition table

oyama:
Add a sample that reads the partition table from runtime.

I would appreciate comments on the partition table operation procedure itself, as I did not know the correct way to do it and coded it by trial and error.


lurch:
@oyama You might find some useful partition-related code in https://github.com/raspberrypi/picotool ? :man_shrugging:

oyama:
@lurch Thank you for the pointer! It looks like the relevant code is around here: https://github.com/raspberrypi/picotool/blob/de8ae5ac334e1126993f72a5c67949712fd1e1a4/bintool/metadata.h#L227 – I'll take a look.
However, my goal is to implement a file system that respects both partitions and permissions, so it seems I'll also need to create a separate read/write example using `rom_flash_op()`.

flash/partition_info/pt.json
```diff
@@ -0,0 +1,46 @@
...
+      "secure": "rw",
+      "nonsecure": "rw",
+      "bootloader": "rw"
+    }
+  },
+  "partitions": [
+    {
+      "name": "Firmware",
+      "id": 1,
+      "size": "512K",
```
will-v-pi, line 15: Probably a good idea to start this partition from 0, rather than the default location, as the PT is being embedded into the binary

flash/partition_info/partition_info.c
will-v-pi, line 1: I think it would be better to follow the approach used by picotool (see the `partition_info_command::execute` function), where it reads the locations and flags for all partitions first, and then individually reads out the variable length names for each partition. Picotool also reads the IDs with the locations and flags, but this could read the IDs separately along with the names.

This would also help reduce the magic numbers, eg `workarea[816]` could become `workarea[PARTITION_TABLE_MAX_PARTITIONS*2 + 4]`, and could even be included in the `pico_partition_t` struct to remove the need for it being static.

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,160 @@
...
+    int flags;
+    uint32_t *table;
+    size_t table_size;
+    int partition_count;
+    uint32_t unpartitioned_space_first_sector;
+    uint32_t unpartitioned_space_last_sector;
+    uint32_t permission;
+    int current_partition;
+    size_t idx;
+} pico_partition_t;
```
will-v-pi, line 31: Maybe rename this to `pico_partition_table_t` and the other one to `pico_partition_t`?

flash/partition_info/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_executable(partition_info partition_info.c)
+target_link_libraries(partition_info PRIVATE
+  pico_stdlib
+  pico_bootrom
+)
+
+pico_embed_pt_in_binary(partition_info ${CMAKE_CURRENT_LIST_DIR}/pt.json)
+pico_enable_stdio_usb(partition_info 1)
```
will-v-pi, line 8: Very minor, but `pico_enable_stdio_usb` should probably come before `pico_embed_pt_in_binary, and they should have simple comments decribing what they do (see runtime_flash_permissions example)

will-v-pi (commented):
A few comments from a first pass reviewing this

flash/partition_info/pt.json
```diff
@@ -0,0 +1,46 @@
...
+        "secure": "rw",
+        "nonsecure": "rw",
+        "bootloader": "rw"
+      }
+    },
+    {
+      "name": "Data",
+      "id": 2,
+      "size": "512K",
+      "families": ["rp2350-arm-s", "rp2350-riscv"],
```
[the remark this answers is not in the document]
oyama: Thanks, Changed to the "data" family.

flash/partition_info/pt.json
```diff
@@ -0,0 +1,46 @@
...
+        "secure": "rw",
+        "nonsecure": "rw",
+        "bootloader": "rw"
+      }
+    },
+    {
+      "name": "Read only Data",
+      "id": 3,
+      "size": "512K",
+      "families": ["rp2350-arm-s", "rp2350-riscv"],
```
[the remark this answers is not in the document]
oyama: Changed to the "data" family. 
It might be beneficial to have a separate explanation for partition operations. I’m planning to use the procedure that changes a partition from `rw` to `r` using the UF2 created with `picotool partition create`.

flash/partition_info/pt.json
oyama: Okay, that's good. I added `“start”: 0`.

flash/partition_info/partition_info.c
oyama: Thanks, that is more appropriate.
rename `pico_partition_t` to `pico_partition_table_t`, and `pico_partition_entry_t` to `pico_partition_t`.

flash/partition_info/CMakeLists.txt
oyama: Added comment following existing example.
Also removed `pico_enable_stdio_usb`.

flash/partition_info/partition_info.c
oyama: I see, so the procedure is to read each partition individually with PT_INFO_SINGLE_PARTITION instead of reading the whole partition table at once.

I'll give it a try.

oyama:
Based on the comments, the read process of the partition table was divided and organized into a fixed-length field portion and a variable-length field portion.

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,164 @@
...
+        uint32_t id_high = pt->table[idx++];
+        p->partition_id = ((uint64_t)id_high << 32) | id_low;
+    } else {
+        p->partition_id = 0;
+    }
+    pt->idx = idx;
+
+    if (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS) {
+        // Read variable length fields
+        uint32_t name_buf[(PARTITION_NAME_MAX + 1) / WORD_SIZE] = {0};
```
will-v-pi, line 104: I don't think `WORD_SIZE` needs to be defined, you could use `sizeof(uint32_t)` instead and remove the `WORD_SIZE` define at the top

```suggestion
        uint32_t name_buf[(PARTITION_NAME_MAX + 1) / sizeof(uint32_t)] = {0};
```

flash/partition_info/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+add_executable(partition_info partition_info.c)
+
+target_link_libraries(partition_info PRIVATE
+        pico_stdlib
+        pico_bootrom
+        )
+
+# not necessary for the example functionality, but shows you how you would add a partition table
```
will-v-pi, line 8: I think for this example the partition table is necessary

```suggestion
# add a partition table
```

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,164 @@
...
+ */
+#include <stdio.h>
+#include <pico/stdlib.h>
+#include <pico/bootrom.h>
+#include "boot/picobin.h"
+
+
+#define PART_LOC_FIRST(x) ( ((x) & PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_BITS) >> PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_LSB )
+#define PART_LOC_LAST(x)  ( ((x) & PICOBIN_PARTITION_LOCATION_LAST_SECTOR_BITS)  >> PICOBIN_PARTITION_LOCATION_LAST_SECTOR_LSB )
+#define WORD_PADDED_BYTES(x)  ((x + 3) & ~3)
```
will-v-pi, line 14: This isn't used, so should be removed

```suggestion
```

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,164 @@
...
+
+#define WORD_SIZE  4
```
will-v-pi, line 16: This can be removed too

```suggestion
```

will-v-pi (commented):
LGTM, just a few minor things

oyama:
Thanks suession. Very helpful.

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,161 @@
...
+#include <pico/bootrom.h>
+#include "boot/picobin.h"
+
+
+#define PART_LOC_FIRST(x) ( ((x) & PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_BITS) >> PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_LSB )
+#define PART_LOC_LAST(x)  ( ((x) & PICOBIN_PARTITION_LOCATION_LAST_SECTOR_BITS)  >> PICOBIN_PARTITION_LOCATION_LAST_SECTOR_LSB )
+
+#define PARTITION_LOCATION_AND_FLAGS_SIZE  2
+#define PARTITION_ID_SIZE                  2
+#define PARTITION_NAME_MAX                 127
```
lurch, line 17: @will-v-pi Is it worth hoisting any of these defines into pico-sdk, or is it fine for each project to define them itself?

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,161 @@
+/**
+ * Copyright (c) 2023 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#include <stdio.h>
+#include <pico/stdlib.h>
+#include <pico/bootrom.h>
+#include "boot/picobin.h"
```
lurch, line 9: Small nitpick, but any reason for the mix of `<>` and `""` ?

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,161 @@
...
+} pico_partition_t;
+
+
+/*
+ * Read the partition table information.
+ *
+ * See the RP2350 datasheet 5.1.2, 5.4.8.16 for flags and structures that can be specified.
+ */
+int pico_partitions_open(pico_partition_table_t *pt) {
+    // Reads a fixed size fields
```
lurch, line 53: nitpick: "Reads a fixed size fields" -> "Reads fixed size fields"

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,161 @@
...
+
+    pt->current_partition++;
+    return true;
+}
+
+int main() {
+    stdio_init_all();
+
+    pico_partition_table_t pt;
+    pico_partitions_open(&pt);
```
lurch, line 128: This is just a gut-feeling thing, but I think that functions named `pico_...` feel like they ought to be part of some standalone library, rather than just locally-defined utility functions? :thinking: 
You're entirely free to ignore this suggestion, but I wonder if it might be worth renaming your functions to `read_partition_table` and `read_next_partition` ?

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,161 @@
...
+        pt->status = rc;
+        return rc;
+    }
+
+    size_t pos = 0;
+    pt->fields = pt->table[pos++];
+    pt->partition_count = pt->table[pos++] & 0x000000FF;
+    uint32_t location = pt->table[pos++];
+    pt->unpartitioned_space_first_sector = PART_LOC_FIRST(location);
+    pt->unpartitioned_space_last_sector = PART_LOC_LAST(location);
```
lurch, line 68: I dunno if it's worth printing out these values or not in your `main{}` ? :man_shrugging:

flash/partition_info/pt.json
```diff
@@ -0,0 +1,47 @@
...
+      "bootloader": "rw"
+    }
+  },
+  "partitions": [
+    {
+      "name": "Firmware",
+      "id": 1,
+      "start": 0,
+      "size": "512K",
+      "families": ["rp2350-arm-ns", "rp2350-arm-s", "rp2350-riscv"],
```
lurch, line 17: @will-v-pi Do these "family" identifiers also get stored in the partition table? Is it worth the example-code also printing them out?

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,161 @@
...
+ * See the RP2350 datasheet 5.1.2, 5.4.8.16 for flags and structures that can be specified.
+ */
+int pico_partitions_open(pico_partition_table_t *pt) {
+    // Reads a fixed size fields
+    int rc = rom_get_partition_table_info(pt->table, sizeof(pt->table),
+                                          (PT_INFO_PT_INFO |
+                                           PT_INFO_PARTITION_LOCATION_AND_FLAGS |
+                                           PT_INFO_PARTITION_ID));
+    if (rc < 0) {
+        pt->status = rc;
```
lurch, line 59: Perhaps it's also worth setting `pt->partition_count = 0` here, so that the code in `pico_partitions_next` doesn't end up reading an uninitialised value? Alternatively, perhaps the `pico_partitions_open(&pt);` line in `main()` ought to actually check the return code? :wink:

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,161 @@
...
+ * Read the partition table information.
+ *
+ * See the RP2350 datasheet 5.1.2, 5.4.8.16 for flags and structures that can be specified.
+ */
+int pico_partitions_open(pico_partition_table_t *pt) {
+    // Reads a fixed size fields
+    int rc = rom_get_partition_table_info(pt->table, sizeof(pt->table),
+                                          (PT_INFO_PT_INFO |
+                                           PT_INFO_PARTITION_LOCATION_AND_FLAGS |
+                                           PT_INFO_PARTITION_ID));
```
lurch, line 57: Attempting to match this up against 5.4.8.16 in the RP2350 datasheet - do you need to explicitly request `PARTITION_NAME` too? The description there also warns "the first word in the returned buffer, is the (sub)set of those flags that the API supports. You should always check this value before interpreting the buffer." but you don't seem to be checking the value of your `pt->fields` ?

flash/partition_info/partition_info.c
oyama: No reason, just a typo on my part. I will unify it.

flash/partition_info/partition_info.c
oyama: Thanks, I'll fix it.

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,161 @@
...
+
+/*
+ * Extract each partition information
+ */
+bool pico_partitions_next(pico_partition_table_t *pt, pico_partition_t *p) {
+    if (pt->current_partition >= pt->partition_count) {
+        return false;
+    }
+
+    size_t idx = pt->idx;
```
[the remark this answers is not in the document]
oyama: Agreed. I also felt uncomfortable with it and will correct it.

flash/partition_info/partition_info.c
oyama: Agreed.
I started out writing this as a function to suggest to pico-sdk, so this is a remnant of that. But since this is a sample program, I think you are right that `read_*` is appropriate.

flash/partition_info/partition_info.c
oyama: I see, good point.
I will make sure to explicitly initialize the `pt->partition_count = 0` in open.

flash/partition_info/partition_info.c
oyama: Add a check for `PT->FIELDS`.
The reason for not requesting `PT_INFO_PARTITION_NAME` at this location, as in the picotool implementation, is to simplify the parsing process by splitting the steps as follows

1. request only fixed-length fields for the entire partition table
2. if a variable length field (`PT_INFO_PARTITION_NAME`) is included in the partition, request it separately.

However, assuming that the fixed-length fields requested in step 1 are always included, my code still seems to need to assert the `pt->fields` match.

lurch: Ah! I'd missed that you were requesting `PT_INFO_PARTITION_NAME` in a separate function call, sorry.

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,162 @@
...
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NSBOOT_R_BITS ? "r" : ""),
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NSBOOT_W_BITS ? "w" : ""),
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NS_R_BITS ? "r" : ""),
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NS_W_BITS ? "w" : ""));
+    printf("partitions:\n");
+    pico_partition_t p;
+    while (read_next_partition(&pt, &p)) {
+        printf("%3d:", pt.current_partition - 1);
+        printf("    %08x->%08x S(%s%s) NSBOOT(%s%s) NS(%s%s)",
+               p.first_sector * 4096, (p.last_sector + 1) * 4096,
```
lurch, line 142: It _might_ be worth doing `#define SECTOR_SIZE 4096` and then using that here?

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,162 @@
...
+        uint32_t id_high = pt->table[pos++];
+        p->partition_id = ((uint64_t)id_high << 32) | id_low;
+    } else {
+        p->partition_id = 0;
+    }
+    pt->pos = pos;
+
+    if (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS) {
+        // Read variable length fields
+        uint32_t name_buf[(PARTITION_NAME_MAX + 1) / sizeof(uint32_t)] = {0};
```
lurch, line 103: Does there need to be an extra ` + 1` in here, to account for the first word (`name_buf[0]`) that stores the fields? :thinking: i.e.
```c
uint32_t name_buf[((PARTITION_NAME_MAX + 1) / sizeof(uint32_t)) + 1] = {0};
```

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,162 @@
...
+        int rc = rom_get_partition_table_info(name_buf, sizeof(name_buf),
+                                              (pt->current_partition << 24 | flags));
+        if (rc < 0) {
+            pt->status = rc;
+            return false;
+        }
+        uint32_t __attribute__((unused)) fields = name_buf[0];
+        assert(fields == flags);
+        uint8_t *name_buf_u8 = (uint8_t *)&name_buf[1];
+        uint8_t name_length = *name_buf_u8++;
```
lurch, line 114: Datasheet says "Byte 0 : 7 bit length of the name (LEN); top bit reserved" so perhaps there ought to be a ` & 0x7F` here?

flash/partition_info/partition_info.c
```diff
@@ -49,7 +49,7 @@ typedef struct {
  *
  * See the RP2350 datasheet 5.1.2, 5.4.8.16 for flags and structures that can be specified.
  */
-int pico_partitions_open(pico_partition_table_t *pt) {
+int open_partition_table(pico_partition_table_t *pt) {
```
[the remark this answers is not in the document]
oyama: I agree with that sense. However, I find that the combination of `read_partition_table` and `read_next_partition` makes the distinction in their roles somewhat unclear. What do you think about renaming them to `load_partition_table` and `each_partition_entry` instead?

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,162 @@
...
+
+    pt->current_partition++;
+    return true;
+}
+
+int main() {
+    stdio_init_all();
+
+    pico_partition_table_t pt;
+    open_partition_table(&pt);
```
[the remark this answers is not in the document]
oyama: Oh, this is not safe. It could fail, or the partition could not be included.
This will be corrected.

flash/partition_info/partition_info.c
oyama: I agree.
There are two options. Either use `#define SECTOR_SIZE 4096` in the sample code or `#include “hardware/flash.h”` and use `FLASH_SECTOR_SIZE`, which do you think is more appropriate?

flash/partition_info/partition_info.c
oyama: Eagle eyes! That's right.
I stand corrected.

flash/partition_info/partition_info.c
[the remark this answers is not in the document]
lurch: Given that this is just a standalone example (and not a library or part of some bigger project), I guess it doesn't really matter what the functions are called; so I'll leave the final decision up to you.

flash/partition_info/partition_info.c
lurch: Use whichever you feel most comfortable with :slightly_smiling_face:  (but personally, I'd always prefer to re-use existing constants, rather than creating new ones)

flash/partition_info/partition_info.c
```diff
@@ -25,6 +25,7 @@
 typedef struct {
     uint32_t table[PARTITION_TABLE_FIXED_INFO_SIZE];
     uint32_t fields;
+    bool has_partition;
```
lurch, line 28: At first I thought this was a typo for `has_partitions`, but after cross-checking with the datasheet I guess it ought to be called `has_partition_table`?

flash/partition_info/partition_info.c
```diff
@@ -132,14 +135,19 @@ int main() {
     if (rc != 0) {
         panic("rom_get_partition_table_info returned %d", pt.status);
     }
-
+    if (!pt.has_partition) {
```
lurch, line 138: If there's no partition table at all, then I guess that implies that any other information you read might be invalid? (i.e. perhaps your program shouldn't try printing out any other information?)

flash/partition_info/partition_info.c
```diff
@@ -132,14 +135,19 @@ int main() {
...
+    }
     printf("un-partitioned_space: S(%s%s) NSBOOT(%s%s) NS(%s%s)\n",
            (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_S_R_BITS ? "r" : ""),
            (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_S_W_BITS ? "w" : ""),
            (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NSBOOT_R_BITS ? "r" : ""),
            (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NSBOOT_W_BITS ? "w" : ""),
            (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NS_R_BITS ? "r" : ""),
            (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NS_W_BITS ? "w" : ""));
+
     printf("partitions:\n");
```
lurch, line 151: I guess there's no point printing `partitions:` if `(pt.partition_count == 0)` ?

flash/partition_info/partition_info.c
oyama: Thanks. Change to `has_partition_table`.

flash/partition_info/partition_info.c
oyama: Even if `pt.has_partition == false`, which does not set the partition table, the `unpartitioned_space_first_sector` and `unpartitioned_space_last_sector` are set from `0x 0000000000` to `0x02000000` as usual.
Thus, it appears that valid information is set.

flash/partition_info/partition_info.c
oyama: Yes, modify to block subsequent codes based on `pt.partition_count`.

flash/partition_info/partition_info.c
will-v-pi: It'll probably be sensible to add them to `picobin.h`, along with more partition table handling in `bootrom.c` - @kilograham might have opinions too?

flash/partition_info/partition_info.c
will-v-pi: Yeah, the unpartitioned space information will always be valid, as there is a default partition table that the bootrom will use when no partition table is specified.

flash/partition_info/uf2_family_ids.c
```diff
@@ -0,0 +1,75 @@
...
+
+void uf2_family_ids_free(uf2_family_ids_t *ids) {
+    for (size_t i = 0; i < ids->count; i++) {
+        free(ids->items[i]);
+    }
+    free(ids->items);
+    free(ids);
+}
+
+void uf2_family_ids_add_default_families(uf2_family_ids_t *ids, uint32_t flags) {
```
lurch, line 56: Rather than having a separate `uf2_family_ids_add_default_families` function, would it make more sense to just pass the `default_families` directly to your `uf2_family_ids_new` function?

flash/partition_info/partition_info.c
```diff
@@ -168,6 +189,18 @@ int main() {
...
+        // print UF2 family ID
+        family_ids = uf2_family_ids_new();
+        uf2_family_ids_add_default_families(family_ids, p.flags_and_permissions);
+        for (size_t i = 0; i < p.extra_family_id_count; i++) {
+            uf2_family_ids_add_extra_family_id(family_ids, p.extra_family_ids[i]);
+        }
+        str_family_ids = uf2_family_ids_join(family_ids, ", ");
+        printf(", uf2 { %s }", str_family_ids);
+        uf2_family_ids_free(family_ids);
+        free(str_family_ids);
```
lurch, line 202: Super minor nitpick, but perhaps it makes sense to swap around `uf2_family_ids_free` and `free`, so that they appear in the same order as they do on lines 167 and 168 ?

flash/partition_info/uf2_family_ids.c
```diff
@@ -0,0 +1,75 @@
...
+
+uf2_family_ids_t *uf2_family_ids_new(void) {
+    uf2_family_ids_t *ids = malloc(sizeof(uf2_family_ids_t));
+    ids->count = 0;
+    ids->items = NULL;
+    return ids;
+}
+
+void uf2_family_ids_add(uf2_family_ids_t *ids, const char *str) {
+    ids->items = realloc(ids->items, (ids->count + 1) * sizeof(char *));
```
lurch, line 13: I guess calculating the required size of `items` up-front just once and `malloc`ing the memory just once would be more efficient than all the `realloc`ing? But it probably doesn't matter for this simple standalone example.

flash/partition_info/partition_info.c
```diff
@@ -9,13 +9,15 @@
...
 #include "hardware/flash.h"
+#include "uf2_family_ids.h"
 
 #define PART_LOC_FIRST(x) ( ((x) & PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_BITS) >> PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_LSB )
 #define PART_LOC_LAST(x)  ( ((x) & PICOBIN_PARTITION_LOCATION_LAST_SECTOR_BITS)  >> PICOBIN_PARTITION_LOCATION_LAST_SECTOR_LSB )
 
 #define PARTITION_LOCATION_AND_FLAGS_SIZE  2
 #define PARTITION_ID_SIZE                  2
 #define PARTITION_NAME_MAX                 127
+#define PARTITION_EXTRA_FAMILY_ID_MAX      3
```
lurch, line 20: Would `uf2_family_ids.h` be a sensible place for this? :man_shrugging:

flash/partition_info/partition_info.c
```diff
@@ -101,25 +105,34 @@ bool read_next_partition(pico_partition_table_t *pt, pico_partition_t *p) {
...
+        uint32_t extra_family_id_and_name[PARTITION_EXTRA_FAMILY_ID_MAX + (((PARTITION_NAME_MAX + 1) / sizeof(uint32_t)) + 1)];
+        uint32_t flags = PT_INFO_SINGLE_PARTITION | PT_INFO_PARTITION_FAMILY_IDS | PT_INFO_PARTITION_NAME;
+        int rc = rom_get_partition_table_info(extra_family_id_and_name, sizeof(extra_family_id_and_name),
                                               (pt->current_partition << 24 | flags));
         if (rc < 0) {
             pt->status = rc;
             return false;
         }
-        uint32_t __attribute__((unused)) fields = name_buf[0];
+        size_t pos = 0;
```
lurch, line 120: Will this conflict with the `size_t pos` on line 93?

flash/partition_info/partition_info.c
```diff
@@ -101,25 +105,34 @@ bool read_next_partition(pico_partition_table_t *pt, pico_partition_t *p) {
...
 
-    if (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS) {
+    p->extra_family_id_count = (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_ACCEPTS_NUM_EXTRA_FAMILIES_BITS)
+                                   >> PICOBIN_PARTITION_FLAGS_ACCEPTS_NUM_EXTRA_FAMILIES_LSB;
+    if (p->extra_family_id_count | (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS)) {
         // Read variable length fields
-        uint32_t name_buf[((PARTITION_NAME_MAX + 1) / sizeof(uint32_t)) + 1] = {0};
-        uint32_t flags = PT_INFO_SINGLE_PARTITION | PT_INFO_PARTITION_NAME;
-        int rc = rom_get_partition_table_info(name_buf, sizeof(name_buf),
+        uint32_t extra_family_id_and_name[PARTITION_EXTRA_FAMILY_ID_MAX + (((PARTITION_NAME_MAX + 1) / sizeof(uint32_t)) + 1)];
```
lurch, line 112: (another) minor nitpick, but should this variable be called `extra_family_ids_and_name`?

flash/partition_info/uf2_family_ids.c
lurch: Ahhh, but having read the code a bit deeper I guess any `realloc` cost is going to be swamped by the `strdup`s anyway!

lurch:
@oyama I'd just like to say thank you very much for putting up with all my niggly comments. I sometimes get a bit carried away when reviewing code :laughing:

flash/partition_info/partition_info.c
```diff
@@ -101,25 +105,34 @@ bool read_next_partition(pico_partition_table_t *pt, pico_partition_t *p) {
     }
     pt->pos = pos;
 
-    if (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS) {
+    p->extra_family_id_count = (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_ACCEPTS_NUM_EXTRA_FAMILIES_BITS)
+                                   >> PICOBIN_PARTITION_FLAGS_ACCEPTS_NUM_EXTRA_FAMILIES_LSB;
+    if (p->extra_family_id_count | (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS)) {
```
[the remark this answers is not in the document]
oyama: That's good. I'll fix it.

flash/partition_info/uf2_family_ids.c
oyama: Indeed, that method looks simple and good. I'll fix it.

flash/partition_info/partition_info.c
oyama: I see, I understand the feeling well. I'll fix it.

flash/partition_info/uf2_family_ids.c
oyama: I could not think of a way to write this section that would keep it simple, so I would like to go with the current implementation.

flash/partition_info/partition_info.c
oyama: Indeed it is,  I'll fix it.

flash/partition_info/partition_info.c
oyama: Uh, I can't remember if there are conflicts in C11. I'll change the name to `pos_` just in case.

flash/partition_info/partition_info.c
oyama: That's right, I'll fix it.

oyama:
@lurch Thank you so much for your detailed feedback, it was incredibly helpful and inspiring. I truly appreciate your attention to every little detail, and your comments have motivated me to improve further. Thanks again for your support!

flash/partition_info/partition_info.c
```diff
@@ -124,14 +126,14 @@ bool read_next_partition(pico_partition_table_t *pt, pico_partition_t *p) {
...
-        if (p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS) {
+        if (p->has_name) {
             uint8_t *name_buf = (uint8_t *)&extra_family_id_and_name[pos];
             uint8_t name_length = *name_buf++ & 0x7F;
             memcpy(p->name, name_buf, name_length);
             p->name[name_length] = '\0';
         }
     }
-    if (!(p->flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS))
+    if (!(p->has_name))
```
lurch, line 135: I think you could get rid of the inner brackets now :wink:

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,210 @@
...
+/*
+ * Stores information on each partition
+ */
+typedef struct {
+    uint32_t first_sector;
+    uint32_t last_sector;
+    uint32_t flags_and_permissions;
+    uint64_t partition_id;
+    bool has_name;
+    char name[PARTITION_NAME_MAX + 1];  // name length is indicated by 7 bits
```
lurch, line 47: The comment here would probably be better on line 19 ?

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,210 @@
...
+               (p.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_S_R_BITS ? "r" : ""),
+               (p.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_S_W_BITS ? "w" : ""),
+               (p.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NSBOOT_R_BITS ? "r" : ""),
+               (p.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NSBOOT_W_BITS ? "w" : ""),
+               (p.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NS_R_BITS ? "r" : ""),
+               (p.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NS_W_BITS ? "w" : ""));
+        if (p.flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_ID_BITS) {
+            printf(", id=%016llx", p.partition_id);
+        }
+        if (p.flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_NAME_BITS) {
```
lurch, line 189: Could use `p.has_name` ? :wink:

lurch:
Now that you've added all the code to display extra family IDs, perhaps it's worth exercising that code by putting an extra family ID in your `pt.json` ? :thinking:

flash/partition_info/partition_info.c
oyama: Agreed. I will fix.

flash/partition_info/partition_info.c
oyama: Ahhh, it was leaking. I will fix.
Also, add `p.has_id`.

flash/partition_info/partition_info.c
```diff
@@ -0,0 +1,212 @@
...
+
+    uf2_family_ids_t *family_ids = uf2_family_ids_new(pt.flags_and_permissions);
+    char *str_family_ids = uf2_family_ids_join(family_ids, ", ");
+    printf("un-partitioned_space: S(%s%s) NSBOOT(%s%s) NS(%s%s) uf2 { %s }\n",
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_S_R_BITS ? "r" : ""),
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_S_W_BITS ? "w" : ""),
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NSBOOT_R_BITS ? "r" : ""),
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NSBOOT_W_BITS ? "w" : ""),
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NS_R_BITS ? "r" : ""),
+           (pt.flags_and_permissions & PICOBIN_PARTITION_PERMISSION_NS_W_BITS ? "w" : ""),
```
lurch, line 167: Similarly to how you have a function (`uf2_family_ids_join`) for creating the formatted family-ids string, I wonder if it might be worth pulling out a separate function for creating the `S(%s%s) NSBOOT(%s%s) NS(%s%s)` part of this string? (which I believe you could then also use in the per-partition loop?) :thinking:

lurch: Haha, I was about to suggest that you could use `%c` instead of `%s` in your `printf` formatting, but then I realised that you can't do "empty string" when using `char` :joy:

flash/partition_info/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+add_executable(partition_info partition_info.c uf2_family_ids.c)
+
+target_link_libraries(partition_info PRIVATE
+        pico_stdlib
+        pico_bootrom
+        hardware_flash
```
will-v-pi, line 6:
```suggestion
        hardware_flash
        boot_uf2_headers
```

flash/partition_info/uf2_family_ids.c
```diff
@@ -0,0 +1,76 @@
...
+        free(ids->items[i]);
+    }
+    free(ids->items);
+    free(ids);
+}
+
+void uf2_family_ids_add_extra_family_id(uf2_family_ids_t *ids, uint32_t family_id) {
+    char hex_id[UF2_FAMILY_ID_HEX_SIZE];
+    sprintf(hex_id, "0x%08x", family_id);
+    _add(ids, hex_id);
```
will-v-pi, line 75:
```suggestion
    switch (family_id) {
        case CYW43_FIRMWARE_FAMILY_ID:
            _add(ids, "cyw43-firmware");
            break;
        default:
            sprintf(hex_id, "0x%08x", family_id);
            _add(ids, hex_id);
            break;
    }
```

flash/partition_info/uf2_family_ids.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "pico/stdlib.h"
+#include "boot/picobin.h"
```
will-v-pi, line 8:
```suggestion
#include "boot/picobin.h"
#include "boot/uf2.h"
```

will-v-pi (commented):
LGTM, but should probably add support for the new cyw43-firmware family - this can be done with the following changes

Also pinging @kilograham again about whether any of the defines from this example should be moved into the SDK? (I think specifically `PARTITION_LOCATION_AND_FLAGS_SIZE`, `PARTITION_ID_SIZE`, `PARTITION_NAME_MAX`, and possibly add `#define`s for family names)

will-v-pi (approved):

PR #632, opened 2025-04-10
Change TLS example URL

peterharperuk:
Change the TLS example to use our download server. worldtimeapi.org seems to be unreliable and it's a bit impolite to hit their server in our examples.


mudge:
I believe fw-download-alias1.raspberrypi.com uses a self-signed certificate that won't be accepted by most HTTP clients so you might be better off using downloads.raspberrypi.com instead.

peterharperuk:
> I believe fw-download-alias1.raspberrypi.com uses a self-signed certificate

That's exactly why I'm using this url.

PR #635, opened 2025-04-14
Fix workflows for new mbedtls

peterharperuk:
Anything that builds picotool now has to pass "recursive" flag to the "git submodule init" command

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -31,7 +31,7 @@ jobs:
 
     - name: Checkout pico-sdk submodules
       working-directory: ${{github.workspace}}/pico-sdk
-      run: git submodule update --init
+      run: git submodule update --init --recursive
 
     - name: Create Build Environment
       # Some projects don't allow in-source building, so create a separate build directory
--- a/.github/workflows/macOS.yml
+++ b/.github/workflows/macOS.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Checkout pico-sdk submodules
         working-directory: ${{github.workspace}}/pico-sdk
-        run: git submodule update --init
+        run: git submodule update --init --recursive
       - name: Install dependencies
         run: |
           brew install cmake
--- a/.github/workflows/multi-gcc.yml
+++ b/.github/workflows/multi-gcc.yml
@@ -32,7 +32,7 @@ jobs:
 
     - name: Checkout pico-sdk submodules
       working-directory: ${{github.workspace}}/pico-sdk
-      run: git submodule update --init
+      run: git submodule update --init --recursive
 
     - name: Create Build Environment
       # Some projects don't allow in-source building, so create a separate build directory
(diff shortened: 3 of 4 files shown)


peterharperuk:
I was wondering if we should go the other way and make more use of the mbedtls cmake files, although I haven't investigated if this is possible.

will-v-pi:
One benefit of moving to the SDK style is it reduces the number of files picotool compiles, as it doesn't need the x509 or TLS files so can skip those lists,  whereas I'm not sure if there's a way to do that when using the mbedtls cmake files?

But the disadvantage is obviously maintenance, as we need to update the source lists if the sources change

peterharperuk:
> reduces the number of files picotool compiles

Yes - that's a good reason!

will-v-pi (approved):
Probably good to merge this anyway

lurch:
> pass "recursive" flag to the "git submodule init" command

I'm slightly wary of doing that (and yes I'm aware that you've already done that in some of the recent pico-sdk PRs) because I'm worried about a situation like we previously had with tinyUSB cropping up again (where a 3rd-party repo had a lot of unnecessary submodules that we didn't need, which made the recursive git clone **really** slow). It's obviously not directly relevant to pico-sdk / pico-examples, but for an example of a repo with a LOT of submodules, see e.g. https://github.com/micropython/micropython/tree/master/lib

IMHO it's better to explicitly only `init` the submodules that we _actually_ need, rather than recursively init-ing ALL submodules. However I'm happy to adopt the "let's leave it for now, and cross that bridge if we get to it" approach :slightly_smiling_face: :crossed_fingers:

kilograham:
> > pass "recursive" flag to the "git submodule init" command
> 
> I'm slightly wary of doing that (and yes I'm aware that you've already done that in some of the recent pico-sdk PRs, e.g. [raspberrypi/pico-sdk#2416](https://github.com/raspberrypi/pico-sdk/pull/2416) ) because I'm worried about a situation like we previously had with tinyUSB cropping up again (where a 3rd-party repo had a lot of unnecessary submodules that we didn't need, which made the recursive git clone **really** slow). It's obviously not directly relevant to pico-sdk / pico-examples, but for an example of a repo with a LOT of submodules, see e.g. https://github.com/micropython/micropython/tree/master/lib
> 
> IMHO it's better to explicitly only `init` the submodules that we _actually_ need, rather than recursively init-ing ALL submodules. However I'm happy to adopt the "let's leave it for now, and cross that bridge if we get to it" approach 🙂 🤞

yes, recursive init is bad anywhere users might copy, which workflows aren't, but there they can be bad because it takes extra time - better to explicitly checkout

will-v-pi:
I think this can be closed, now https://github.com/raspberrypi/picotool/pull/226 has been merged?

peterharperuk:
Great - thanks. I added the --recursive flag to pico-sdk. I guess I could remove it now.

PR #638, opened 2025-04-26
Add separate builds for Pico and Pico 2 W to GiHub CI

lurch:
Many of the pico-examples are RP2350 only, so add a Pico 2 W build job so that these (and the wireless-only examples) get test-built too

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -33,28 +33,43 @@ jobs:
       working-directory: ${{github.workspace}}/pico-sdk
       run: git submodule update --init
 
-    - name: Create Build Environment
+    - name: Create Build Environment (Pico)
       # Some projects don't allow in-source building, so create a separate build directory
       # We'll use this as our working directory for all subsequent commands
       working-directory: ${{github.workspace}}/pico-examples
-      run:  cmake -E make_directory ${{github.workspace}}/pico-examples/build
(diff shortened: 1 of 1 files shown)


will-v-pi (commented):
Would it be worth using a matrix instead (ie run separate jobs for each board)? This has the advantage of making it easier to add other boards in future and reducing duplication, but the disadvantage that it'll require cloning the SDK separately for each job (so ~30s extra total execution time, plus a bit of overhead as both jobs run sequentially on the same runner).

lurch:
Cool, now looks much neater than my hacky attempt :star_struck:

will-v-pi:
I'll just push a couple more commits to fix up the action, while I'm here

lurch:
> I'll just push a couple more commits to fix up the action, while I'm here

Let me know when you want the green-approval tick :grinning:

lurch:
Is it worth doing this `matrix` thing for the Mac and Windows builds too, or would that just be overkill?
Is it worth trying to do a RISC-V build of the examples also? :shrug:

lurch:
... or @lloydalexporter ?

lurch:
With a sideways glance at #643 perhaps it is worth expanding this to build for each of Pico, Pico W, Pico 2 & Pico 2 W ?

kilograham (approved):

PR #643, opened 2025-04-30
BT examples don't build for pico2_w freertos

peterharperuk:


kilograham (approved):

pico_w/bt/CMakeLists.txt
```diff
@@ -178,7 +178,6 @@ if (TARGET FreeRTOS-Kernel)
                 )
         target_include_directories(picow_bt_example_common_freertos INTERFACE
```
kilograham, line 179: why do we need a target_include_directories at all, does it not come via freertos LIBRARY?

PR #644, opened 2025-05-01
Add to standalone/sensor example to start/stop BT

### LINKED ISSUE #502: Example of starting and stopping btstack needed
See https://github.com/bluekitchen/btstack/issues/550#event-13213206797

peterharperuk:
Fixes #502

### DIFF
--- a/pico_w/bt/standalone/server.c
+++ b/pico_w/bt/standalone/server.c
@@ -40,14 +40,27 @@ static void heartbeat_handler(struct btstack_timer_source *ts) {
     btstack_run_loop_add_timer(ts);
 }
 
+static volatile bool key_pressed;
+void key_pressed_func(void *param) {
+    int key = getchar_timeout_us(0); // get any pending key press but don't wait
+    if (key == 's' || key == 'S') {
+        key_pressed = true;
+    }
+}
+
 int main() {
     stdio_init_all();
 
+restart:
     // initialize CYW43 driver architecture (will enable BT if/because CYW43_ENABLE_BLUETOOTH == 1)
     if (cyw43_arch_init()) {
         printf("failed to initialise cyw43_arch\n");
         return -1;
     }
+
+    // Get notified if the user presses a key
+    printf("Press the \"S\" key to Stop bluetooth\n");
+    stdio_set_chars_available_callback(key_pressed_func, NULL);
 
     // Initialise adc for the temp sensor
     adc_init();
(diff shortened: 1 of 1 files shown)


pico_w/bt/standalone/server.c
```diff
@@ -74,20 +87,19 @@ int main() {
...
+
+    cyw43_arch_deinit();
+
+    printf("Press the \"S\" key to Start bluetooth\n");
+    key_pressed = false;
+    while(!key_pressed) {
         sleep_ms(1000);
     }
-#endif
+goto restart;
```
kilograham, line 103: weird indentation

pico_w/bt/standalone/server.c
```diff
@@ -40,15 +40,28 @@ static void heartbeat_handler(struct btstack_timer_source *ts) {
     btstack_run_loop_add_timer(ts);
 }
 
+static bool key_pressed;
```
kilograham, line 43: seems like this should be volatile

pico_w/bt/standalone/server.c
```diff
@@ -74,20 +87,19 @@ int main() {
...
-#if 0 // btstack_run_loop_execute() is not required, so lets not use it
-    btstack_run_loop_execute();
-#else
-    // this core is free to do it's own stuff except when using 'polling' method (in which case you should use 
-    // btstacK_run_loop_ methods to add work to the run loop.
-    
-    // this is a forever loop in place of where user code would go.
-    while(true) {      
+    key_pressed = false;
+    while(!key_pressed) {
```
lurch, line 91: tiny nitpick: the lack-of-space after `while` and before `(` (and also on line 100) is inconsistent with the `if (` on line 56

pico_w/bt/standalone/server.c
```diff
@@ -74,20 +87,19 @@ int main() {
...
+
+    cyw43_arch_deinit();
+
+    printf("Press the \"S\" key to Start bluetooth\n");
+    key_pressed = false;
+    while(!key_pressed) {
         sleep_ms(1000);
     }
-#endif
+    goto restart;
```
lurch, line 103: Any benefit to using `goto` rather than just putting this all inside a `while (true) {` loop?

pico_w/bt/standalone/server.c
peterharperuk: I literally never put a space after while in my whole programming career.

peterharperuk:
I think I need to do the same to the "reader" as well as the "sensor" as I'm not sure btstack is working when that's restarted

kilograham:
> I think I need to do the same to the "reader" as well as the "sensor" as I'm not sure btstack is working when that's restarted

what does this mean? is this ready to merge or not?

kilograham (approved):

PR #653, opened 2025-06-04
Update nuke.c to auto-detect flash size

### LINKED ISSUE #642: Official flash nuke doesn't work, but https://github.com/Gadgetoid/pico-universal-flash-nuke does work
Got an Adafruit RP2040 Adalogger Feather, and it previously had the Adafruit Wippersnapper offline firmware on it, which uses/creates a small FAT user-data partition to share with the OS. 
I used the official flash nuke, then installed a new version of the firmware, but the file system partition showed up on the windows host with the old files still present (config-old.json etc). I repeated the flash nuke downloading it from the pi site instead of adafruit learn guide copy, with the same result.
Then knowing about Gadgetoid's alternative I used that (from https://github.com/Gadgetoid/pico-universal-flash-nuke ), and then installed the new firmware and it then showed the user file system had been correctly erased.

Why does the official one fail to nuke the board entirely?

tyeth:
Fixes https://github.com/raspberrypi/pico-examples/issues/642

Original work by @Gadgetoid 🥇

### DIFF
--- a/flash/nuke/nuke.c
+++ b/flash/nuke/nuke.c
@@ -27,12 +27,11 @@
 
 int main() {
     uint flash_size_bytes;
-#ifndef PICO_FLASH_SIZE_BYTES
-#warning PICO_FLASH_SIZE_BYTES not set, assuming 16M
-    flash_size_bytes = 16 * 1024 * 1024;
-#else
-    flash_size_bytes = PICO_FLASH_SIZE_BYTES;
-#endif
+    uint8_t txbuf[4];
+    uint8_t rxbuf[4];
+    txbuf[0] = 0x9f;
+    flash_do_cmd(txbuf, rxbuf, 4);
+    flash_size_bytes = 1u << rxbuf[3];
     flash_range_erase(0, flash_size_bytes);
     // Leave an eyecatcher pattern in the first page of flash so picotool can
     // more easily check the size:
--- a/flash/nuke/nuke.c
+++ b/flash/nuke/nuke.c
@@ -25,13 +25,18 @@
 #include "hardware/flash.h"
 #include "pico/bootrom.h"
 
+#define MIN_FLASH_SIZE (1024 * 1024 * 2)
+
 int main() {
-    uint flash_size_bytes;
+    uint flash_size_bytes = 0;
     uint8_t txbuf[4];
     uint8_t rxbuf[4];
     txbuf[0] = 0x9f;
     flash_do_cmd(txbuf, rxbuf, 4);
     flash_size_bytes = 1u << rxbuf[3];
+    if (flash_size_bytes < MIN_FLASH_SIZE || flash_size_bytes > PICO_FLASH_SIZE_BYTES) {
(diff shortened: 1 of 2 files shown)


flash/nuke/nuke.c
```diff
@@ -27,12 +27,11 @@
...
-#ifndef PICO_FLASH_SIZE_BYTES
-#warning PICO_FLASH_SIZE_BYTES not set, assuming 16M
-    flash_size_bytes = 16 * 1024 * 1024;
-#else
-    flash_size_bytes = PICO_FLASH_SIZE_BYTES;
-#endif
+    uint8_t txbuf[4];
+    uint8_t rxbuf[4];
+    txbuf[0] = 0x9f;
+    flash_do_cmd(txbuf, rxbuf, 4);
```
lurch, line 33: Is this guaranteed to work with **every** variety of SPI Flash chip that might be used with RP2040 or RP2350?

lurch: > I'm not an expert on this matter.

Me neither! The only reason I asked, was because I assumed that if it _was_ universally supported, then there'd already be a wrapper-function for it in pico-sdk? :thinking: 
But that's merely a hunch, I could be **entirely** wrong! :speak_no_evil:

tyeth: This suggests for an FPGA situation with NOR flash that two possible command sets exist, but I've no idea if that first option is supported by RP2s.
https://electronics.stackexchange.com/a/673366

tyeth: Okay, so digging deeper, yes the Serial Flash Discovery Protocol spec also mandates the backwards compatible support for 0x9f command, or at least as far as my limited research indicates.

kilograham: cc @Wren6991 who i believe has said in the past that such a command is not universally supported

lurch: > That said it’s kinda up to the vendor to verify flash_nuke works on their board and - if not - supply a customised alternative.

:joy: :rofl: :joy: :rofl: :joy: :rofl: 

If every vendor provided even _half_ of the support that Pimoroni did, the world would be a much better place :grinning:

tyeth: Maybe there is a known case of returning wrong info, so some default would be wiser...
https://github.com/adafruit/circuitpython/blob/ffc42b9f0f4aa5e812de5302bc217e1dd5a0f8b6/ports/raspberrypi/supervisor/internal_flash.c#L81C35-L89C6

peterharperuk (changes requested):
-1 as this needs a fix to the sdk

peterharperuk (approved):

will-v-pi (approved):
LGTM, and it still builds universal UF2s (https://github.com/will-v-pi/pico-sdk-uf2s/actions/runs/27211253008)

One thought from testing is it might be nice to turn on the LED when the erase starts, as it can take a long time on some chips? Picos are very fast, but my Pimoroni Pico Plus 2 takes forever.

kilograham:
please add LED as separate PR

PR #659, opened 2025-06-27
Update onboard_temperature.c

ashleey30:
i just mentioned 2^12 = 4096

### DIFF
--- a/adc/onboard_temperature/onboard_temperature.c
+++ b/adc/onboard_temperature/onboard_temperature.c
@@ -17,7 +17,7 @@
  * pico-examples/adc/adc_console/adc_console.c */
 float read_onboard_temperature(const char unit) {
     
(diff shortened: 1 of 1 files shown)


adc/onboard_temperature/onboard_temperature.c
```diff
@@ -17,7 +17,7 @@
  * pico-examples/adc/adc_console/adc_console.c */
 float read_onboard_temperature(const char unit) {
     
-    /* 12-bit conversion, assume max value == ADC_VREF == 3.3 V */
+    /* 12-bit conversion, assume max value == ADC_VREF == 3.3 V */4096
```
peterharperuk, line 20: That looks like a compile error

lurch: Yeah, did you mean to put the `4096` **inside** the comment @ashleey30 ?

PR #675, opened 2025-07-19
Improve I2C LCD example

### LINKED ISSUE #597: Fixes for several lcd_1602_i2c.c issues
Re the following comment in lcd_toggle_enable(), accompanying the value
of 600 us for `DELAY_US`:

    "We cannot do this too quickly or things don't work"

Various atasheets I've come across for these 16x2 (and 20x4) modules from
3-4 different manufacturers show the Enable minimum active pulse dwell
(e.g. **t_w,min**, on p. 14 of [1]) as 250 ns or less.  So that 600 us value is 
excessive by more than three orders of magnitude.

The reason that the example code needs that huge value in order to "work"
properly is because the required delays (e.g. per [1], p. 17) between the
first three initialization commands are missing from` lcd_init().` But the time
expended in `lcd_toggle_enable()`, due that artificially huge 600 us 
`DELAY_US` is fortuitously compensating (partially) for the missing 
 delays in `lcd_init()` (since `lcd_init()` invokes `lcd_toggle_enable()` via
`lcd_send_byte()` on every write operation.)

Behavior observed with the present code, due to the above issues:

   * Unreliable initialization: Malformed screen characters occasionally
     appear immediately after init. (Observed on approximately 5 % of
     inits of most 16x2 units that I have, depending on LCD vendor, and
     considerably more often (perhaps 10% - 20%) on the 20x4 modules, 
     which have near-identical interface specs).

   * Write times much slower than the LCD module is capable of.

   * Write times scale very poorly with I2C clock rate. (For example
      increase I2C clock by factor of 10 leads to only 20% reduction

To address these issues: If you insert explicit delays between the first
two writes in lcd_init() in order to satisfy the requirements in [1], e.g.

### LINKED ISSUE #654: Hint for i2c Lcd 1602 example
Dear @kilograham,

I looked into the example of lcd1602 and one point came up.

This would be the perfect example making a lib kind of lcd class for the lcd and put only the relevant part in the main file.
Putting all stuff in the main file makes the example unclear from my point of view.

Also would beva good example here how to create a lib kind of an class an use it like in cpp.

Windwoes:
This PR improve the I2C LCD example in the following ways:
 - Extracts driver code to a separate C file for code cleanliness
 - Supports creating multiple instances of the driver so that multiple displays can be driven
 - Improves performance by using better chosen delay values
 - Has functionality to simplify printed formatted strings to the display

The following article was used as a reference for timing: https://web.alfredstate.edu/faculty/weimandn/lcd/lcd_initialization/lcd_initialization_index.html

This code has been tested on real hardware with a Pi Pico.

Fixes #654 
Fixes #597

### DIFF
--- a/i2c/lcd_1602_i2c/i2c_lcd.c
+++ b/i2c/lcd_1602_i2c/i2c_lcd.c
@@ -65,7 +65,7 @@ struct i2c_lcd
     uint8_t address;
 };
 
-static void writeExpanderData(i2c_lcd_handle inst, bool clockDisplayController)
+static void write_expander_data(i2c_lcd_handle inst, bool clockDisplayController)
 {
     if (clockDisplayController)
     {
@@ -89,7 +89,7 @@ static void writeExpanderData(i2c_lcd_handle inst, bool clockDisplayController)
     }
 }
 
-static void write4bits(i2c_lcd_handle inst, uint8_t bits)
+static void write_4_bits(i2c_lcd_handle inst, uint8_t bits)
 {
     // Load in the bits
     inst->portExpanderDat.D4 = (bits & (1 << 0)) != 0;
@@ -97,23 +97,23 @@ static void write4bits(i2c_lcd_handle inst, uint8_t bits)
     inst->portExpanderDat.D6 = (bits & (1 << 2)) != 0;
     inst->portExpanderDat.D7 = (bits & (1 << 3)) != 0;
 
-    writeExpanderData(inst, true);
+    write_expander_data(inst, true);
 }
 
-static void writeByte(i2c_lcd_handle inst, uint8_t data, bool dstControlReg)
+static void write_byte(i2c_lcd_handle inst, uint8_t data, bool dstControlReg)
 {
     inst->portExpanderDat.RS = !dstControlReg;
 
     // Write more significant nibble first
-    write4bits(inst, data >> 4);
+    write_4_bits(inst, data >> 4);
 
     // Then write less significant nibble
-    write4bits(inst, data);
+    write_4_bits(inst, data);
 }
 
-static void executeInstruction(i2c_lcd_handle inst, uint8_t instruction, uint8_t flags)
+static void send_instruction(i2c_lcd_handle inst, uint8_t instruction, uint8_t flags)
 {
-    writeByte(inst, instruction | flags, true);
+    write_byte(inst, instruction | flags, true);
 
     switch (instruction)
     {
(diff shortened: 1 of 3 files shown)


i2c/lcd_1602_i2c/i2c_lcd.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef I2C_LCD_H
+#define I2C_LCD_H
+#include <stdbool.h>
+#include <hardware/i2c.h>
+
+typedef struct i2c_lcd* i2c_lcd_handle;
+
+i2c_lcd_handle i2c_lcd_init(i2c_inst_t* handle, uint8_t address);
+void i2c_lcd_setCursorLocation(i2c_lcd_handle inst, uint8_t x, uint8_t y);
```
lurch, line 15: I wonder if it'd be better to stick to snake_case (i.e. `i2c_lcd_set_cursor_location`) rather than using a weird mix of snake_case and camelCase?

i2c/lcd_1602_i2c/i2c_lcd.h
```diff
@@ -0,0 +1,26 @@
...
+void i2c_lcd_setCursorLine(i2c_lcd_handle inst, uint8_t line);
+void i2c_lcd_writeString(i2c_lcd_handle inst, char* string);
+void i2c_lcd_writeStringf(i2c_lcd_handle inst, const char* __restrict format, ...) _ATTRIBUTE ((__format__ (__printf__, 2, 3)));
+void i2c_lcd_writeChar(i2c_lcd_handle inst, char c);
+void i2c_lcd_writeLines(i2c_lcd_handle inst, char* line1, char* line2);
```
lurch, line 20: This function-signature seems to be assuming that you're using a 16x2 LCD, and precludes the usage of a 20x4 LCD?

i2c/lcd_1602_i2c/main.c
```diff
@@ -0,0 +1,64 @@
...
+int main()
+{
+	// Set up USB virtual COM port
+	stdio_init_all();
+
+	// Set up the I2C peripheral at 100k
+	i2c_init(i2c0, 100 * 1000);
+
+	// Bind pins to I2C peripheral
+	gpio_set_function(0, GPIO_FUNC_I2C);
```
lurch, line 40: Rather than hard-coding this to GPIO 0 and GPIO 1, could you use `PICO_DEFAULT_I2C_SDA_PIN` and `PICO_DEFAULT_I2C_SCL_PIN` please, as the original example did? That way this example will still build and run on boards that make different I2C pins accessible to the user (see https://github.com/raspberrypi/pico-sdk/tree/develop/src/boards/include/boards for examples of this being used).

i2c/lcd_1602_i2c/i2c_lcd.h
Windwoes: I'm happy to revise the case if desired. This mix is my own convention that I've come to use to help me distinguish function names coming from different "classes"/modules, since C doesn't support namespaces.

i2c/lcd_1602_i2c/i2c_lcd.h
Windwoes: That's correct. The example name specifically says it's for 16x2. I also don't have 20x4 hardware to validate against.

i2c/lcd_1602_i2c/i2c_lcd.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#ifndef I2C_LCD_H
+#define I2C_LCD_H
+#include <stdbool.h>
+#include <hardware/i2c.h>
+
+typedef struct i2c_lcd* i2c_lcd_handle;
+
+i2c_lcd_handle i2c_lcd_init(i2c_inst_t* handle, uint8_t address);
```
lurch, line 14: If you're intending this to be a "library-like example", then it's probably worth adding some comments to this header-file, with a  brief explanation of what each function does, and what arguments it expects?

i2c/lcd_1602_i2c/i2c_lcd.h
Windwoes: Fixed

i2c/lcd_1602_i2c/main.c
Windwoes: Fixed

i2c/lcd_1602_i2c/i2c_lcd.c
```diff
@@ -0,0 +1,255 @@
...
+    inst->portExpanderDat.RS = !dstControlReg;
+
+    // Write more significant nibble first
+    write4bits(inst, data >> 4);
+
+    // Then write less significant nibble
+    write4bits(inst, data);
+}
+
+static void executeInstruction(i2c_lcd_handle inst, uint8_t instruction, uint8_t flags)
```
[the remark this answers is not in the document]
Windwoes: Fixed

i2c/lcd_1602_i2c/i2c_lcd.h
Windwoes: Fixed

PR #680, opened 2025-07-24
Fix warning in uart_boot

will-v-pi:
assert -> hard_assert to fix unused variable compile warning in release builds


andygpz11 (approved):
LGTM.

kilograham (approved):

PR #689, opened 2025-07-29
Update multi gcc

kilograham:


.github/workflows/multi-gcc.yml
```diff
@@ -41,141 +41,331 @@ jobs:
...
       shell: bash
       run: cd ${{github.workspace}}/pico-examples; mkdir -p build; rm -rf build/*; cd build; PICO_SDK_PATH=../../pico-sdk cmake ../ -DCMAKE_BUILD_TYPE=Debug -DPICO_TOOLCHAIN_PATH=/opt/arm/arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi -DPICO_BOARD=pico_w; make --output-sync=target --no-builtin-rules --no-builtin-variables -j$(nproc)
 
     - name: GCC 11.3.1 Release
-      if: always()
+      if: ${{ !cancelled() }}
       shell: bash
       run: cd ${{github.workspace}}/pico-examples; mkdir -p build; rm -rf build/*; cd build; PICO_SDK_PATH=../../pico-sdk cmake ../ -DCMAKE_BUILD_TYPE=Release -DPICO_TOOLCHAIN_PATH=/opt/arm/arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi -DPICO_BOARD=pico_w; make --output-sync=target --no-builtin-rules --no-builtin-variables -j$(nproc)
 
     - name: GCC 12.2.1 Debug
```
lurch, line 163: For consistency, it might be nice if all these older compiler-builds were also labelled with "... Pico W", just like GCC 13 and above are? e.g.
```
    - name: GCC 12.2.1 Debug Pico W
```
(because at first I thought "Why are _only_ the new compilers building for Pico W?", before I spotted that the old compilers are also doing `-DPICO_BOARD=pico_w` :wink: )

PR #690, opened 2025-07-29
Improve constants for picow_ntp_client.c

hasenradball:
Improve Constants.

Resend has nothing to do with re send a request, it is the time when the failed message occurs.

Request time is more clear whats behind.


hasenradball:
bot

PR #692, opened 2025-08-05
Add script (and CI check) to check / update the autogenerated pioasm outputs

lurch:


lurch:
I did this as two separate commits, to show the CI failing on the first commit, but passing on the second commit :grinning: 

**Huge** thanks to @will-v-pi for all his help with the GitHubCI part of this PR.

(My assumption here is that people building the `develop` branch of pico-examples, will probably also be building against the `develop` branch of pico-sdk.)

PR #697, opened 2025-08-18
Properly map IO and PAD register offsets for QSPI

lifeisafractal:
In the UART bootloader example the QSPI pins are used for normal IO and a `qspi_gpio_set_function()` function sets the pins' IO and PAD registers for use. This function didn't take into account that the IO and PAD register banks for the QSPI IOs are not in the same order. This PR corrects this by adding a lookup table to map between the two and takes an enum as the gpio parameter to signal to the caller the expected order for the function call.

It is worth noting that this example did happen to work for UART TX only due to happenstance. It configured offsets 3 and 4 (which were correct offsets for the PAD bank for SD2 and SD3), but these offsets map to SD1 and SD2 in the IO bank. This results in SD2 getting fully configured which provides UART0 TX, UART0 RX is left mis-configured and non-functional (only the PAD is configured), and SD1 gets its IO register erroneously configured. This example never made use of RX which is likely why this got missed since TX worked.

Even though this example happens to work, it's wroth correcting this function for posterity. Since the pico-sdk doesn't have functions for configuring the QSPI IOs, I found this example and assumed I could safely borrow this function, but the fact that it is incorrect lead to me burning several hours before I figured out what was going on. It's also likely worth getting the pin config for QSPI into the pico-sdk itself as a first class citizen, but for now it's worth fixing this example.

Also, if anyone has info on why the QSPI PAD and IO register offsets don't match, I'd love to know just out of curiosity. It's also curious that the QSPI IO bank has entries for the USB pins (DP and DM) at the start, but they don't have corresponding entries in PAD. You'd think those would have been put at the end of the bank. This is a non-issue code wise as the DP/DM are hard coded in the QSPI IO register map struct, and the `gpio[]` array for the actual QSPI pins does start at the base of the QSPI pins. It's all just odd...

Unrelated, this PR also adds `static` to local functions used in the UART example as a best practice.

### DIFF
--- a/bootloaders/uart/uart_binary.c
+++ b/bootloaders/uart/uart_binary.c
@@ -71,8 +71,8 @@ int main() {
     pico_led_init();
 
     // SD2 is UART0 TX, SD3 is UART0 RX (same as the ROM UART bootloader)
-    qspi_gpio_set_function(QSPI_GPIO_SD2, GPIO_FUNC_UART_AUX);
-    qspi_gpio_set_function(QSPI_GPIO_SD3, GPIO_FUNC_UART_AUX);
+    qspi_gpio_set_function(QSPI_GPIO_SD2, GPIO_FUNC1_UART_AUX);
+    qspi_gpio_set_function(QSPI_GPIO_SD3, GPIO_FUNC1_UART_AUX);
 
     uart_init(uart0, 1000000);
 
--- a/bootloaders/uart/uart_binary.c
+++ b/bootloaders/uart/uart_binary.c
@@ -14,7 +14,7 @@
 #endif
 
 // Initialize the GPIO for the LED
-static void pico_led_init(void) {
+void pico_led_init(void) {
 #ifdef PICO_DEFAULT_LED_PIN
     // A device like Pico that uses a GPIO for the LED will define PICO_DEFAULT_LED_PIN
     // so we can use normal GPIO functionality to turn the led on and off
@@ -24,7 +24,7 @@ static void pico_led_init(void) {
 }
 
 // Turn the LED on or off
-static void pico_set_led(bool led_on) {
+void pico_set_led(bool led_on) {
 #if defined(PICO_DEFAULT_LED_PIN)
     // Just set the GPIO on or off
     gpio_put(PICO_DEFAULT_LED_PIN, led_on);
@@ -43,7 +43,7 @@ enum qspi_gpio {
 
 // curiously the IO and PAD register banks for the QSPI GPIOs are not in the same order
 // This look up table will map the PAD offset to the IO offset for the same pin
-static const uint QSPI_GPIO_PAD_TO_IO_OFFSET[] = {
+const uint QSPI_GPIO_PAD_TO_IO_OFFSET[] = {
     0, // SCLK
     2, // SD0
     3, // SD1
@@ -53,7 +53,7 @@ static const uint QSPI_GPIO_PAD_TO_IO_OFFSET[] = {
 };
 
 // Set function for QSPI GPIO pin
-static void qspi_gpio_set_function(enum qspi_gpio gpio, gpio_function_t fn) {
+void qspi_gpio_set_function(enum qspi_gpio gpio, gpio_function_t fn) {
     // Set input enable on, output disable off
     hw_write_masked(&pads_qspi_hw->io[gpio],
                    PADS_QSPI_GPIO_QSPI_SD2_IE_BITS,
--- a/bootloaders/uart/uart_binary.c
+++ b/bootloaders/uart/uart_binary.c
@@ -14,7 +14,7 @@
 #endif
 
 // Initialize the GPIO for the LED
-void pico_led_init(void) {
+static void pico_led_init(void) {
 #ifdef PICO_DEFAULT_LED_PIN
     // A device like Pico that uses a GPIO for the LED will define PICO_DEFAULT_LED_PIN
     // so we can use normal GPIO functionality to turn the led on and off
@@ -24,15 +24,36 @@ void pico_led_init(void) {
 }
 
 // Turn the LED on or off
-void pico_set_led(bool led_on) {
+static void pico_set_led(bool led_on) {
 #if defined(PICO_DEFAULT_LED_PIN)
     // Just set the GPIO on or off
     gpio_put(PICO_DEFAULT_LED_PIN, led_on);
 #endif
 }
 
+enum qspi_gpio {
+    // arbitrarily choose PAD register bank to set the order
+    QSPI_GPIO_SCLK = 0,
+    QSPI_GPIO_SD0 = 1,
+    QSPI_GPIO_SD1 = 2,
+    QSPI_GPIO_SD2 = 3,
+    QSPI_GPIO_SD3 = 4,
+    QSPI_GPIO_SS = 5,
+};
+
+// curiously the IO and PAD register banks for the QSPI GPIOs are not in the same order
+// This look up table will map the PAD offset to the IO offset for the same pin
+static const uint QSPI_GPIO_PAD_TO_IO_OFFSET[] = {
+    0, // SCLK
+    2, // SD0
+    3, // SD1
+    4, // SD2
+    5, // SD3
+    1, // SS
+};
+
 // Set function for QSPI GPIO pin
-void qspi_gpio_set_function(uint gpio, gpio_function_t fn) {
+static void qspi_gpio_set_function(enum qspi_gpio gpio, gpio_function_t fn) {
     // Set input enable on, output disable off
     hw_write_masked(&pads_qspi_hw->io[gpio],
                    PADS_QSPI_GPIO_QSPI_SD2_IE_BITS,
@@ -40,7 +61,7 @@ void qspi_gpio_set_function(uint gpio, gpio_function_t fn) {
     );
     // Zero all fields apart from fsel; we want this IO to do what the peripheral tells it.
     // This doesn't affect e.g. pullup/pulldown, as these are in pad controls.
-    io_qspi_hw->io[gpio].ctrl = fn << IO_QSPI_GPIO_QSPI_SD2_CTRL_FUNCSEL_LSB;
+    io_qspi_hw->io[QSPI_GPIO_PAD_TO_IO_OFFSET[gpio]].ctrl = fn << IO_QSPI_GPIO_QSPI_SD2_CTRL_FUNCSEL_LSB;
 
     // Remove pad isolation now that the correct peripheral is in control of the pad
     hw_clear_bits(&pads_qspi_hw->io[gpio], PADS_QSPI_GPIO_QSPI_SD2_ISO_BITS);
@@ -49,9 +70,9 @@ void qspi_gpio_set_function(uint gpio, gpio_function_t fn) {
 int main() {
     pico_led_init();
 
-    // SD2 is QSPI GPIO 3, SD3 is QSPI GPIO 4
-    qspi_gpio_set_function(3, GPIO_FUNC_UART_AUX);
-    qspi_gpio_set_function(4, GPIO_FUNC_UART_AUX);
+    // SD2 is UART0 TX, SD3 is UART0 RX (same as the ROM UART bootloader)
+    qspi_gpio_set_function(QSPI_GPIO_SD2, GPIO_FUNC_UART_AUX);
+    qspi_gpio_set_function(QSPI_GPIO_SD3, GPIO_FUNC_UART_AUX);
 
     uart_init(uart0, 1000000);
 


lurch:
Are there any useful defines from https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/hardware_regs/include/hardware/regs/io_qspi.h or https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/hardware_regs/include/hardware/regs/pads_qspi.h that can be used here?

Also, there might be useful info in https://github.com/raspberrypi/pico-sdk/issues/1799 and/or some of the other issues that it links to.

lifeisafractal:
I looked at `io_qspi.h` and `pads_qspi.h` and they do have `IO_QSPI_GPIO_QSPI_*_CTRL_OFFSET` and `PADS_QSPI_GPIO_QSPI_*_OFFSET`, but these are relative to the base of the peripheral where as the `pads_qspi_hw` and `io_qspi_hw` register map structs have the `gpio` array within the that are each relative to the base of the GPIOs within that bank (also, the IO bank has 2x32-bit registers per GPIO). My opinion is it'd end up being more confusing to try utilize any defines from these files but I'm open to a suggestion. FWIW, I'm referencing [`gpio_set_function(...)`](https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/hardware_gpio/gpio.c#L38-L53) which uses the similar IO/PAD reg map structs for the normal GPIO bank to setup the pins.

I had come across raspberrypi/pico-sdk#1799 before as I have a product in the works that use a QSPI-less RP2350 as a co-processor using the the UART bootloader. I went through the linked issues and PRs and I don't think there is too much new in any of those beyond some reinforcing discussion that this would be a nice-to-have in the SDK. I think the short of the long is that someone (maybe me?) just needs to go take a stab at adding first-class support for the QSPI io bank to the pico-sdk.

Independent of doing that (which is a bigger lift), I think it's worth getting this fix merged here still. This example seems to be where people trying to do QSPI IO on the RP2350 end up after some googling, and as it stands now this code is broken for general use cases which doesn't feel great.

lurch:
> Independent of doing that (which is a bigger lift), I think it's worth getting this fix merged here still.

I totally agree! Sorry if I gave the opposite impression. (I was just wondering if there was an "easy" way to avoid having two "lookup tables" hard-coded inside this example, but you've obviously already looked at this in more detail than me :slightly_smiling_face: )

ping @will-v-pi (who wrote this uart-bootloader example).

EDIT: Also ping @matiasilva in case there's any "fiddling" that needs to be done in the auto-generation of the `io_qspi_hw` and `pads_qspi_hw` structs that would make this any "easier" to solve in the long-term. EDIT2: At the very least, perhaps [hardware/structs/io_qspi.h#L196](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/hardware_structs/include/hardware/structs/io_qspi.h#L196) and [hardware/structs/pads_qspi.h#L42](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/hardware_structs/include/hardware/structs/pads_qspi.h#L42) should have a comment listing what order the GPIOs are in within each `io[6]` array?

bootloaders/uart/uart_binary.c
```diff
@@ -24,23 +24,44 @@ void pico_led_init(void) {
...
-void pico_set_led(bool led_on) {
+static void pico_set_led(bool led_on) {
 #if defined(PICO_DEFAULT_LED_PIN)
     // Just set the GPIO on or off
     gpio_put(PICO_DEFAULT_LED_PIN, led_on);
 #endif
 }
 
+enum qspi_gpio {
+    // arbitrarily choose PAD register bank to set the order
```
lurch, line 35: Not that it really matters, but I wonder if it might be better to "arbitrarily choose" the IO register bank here (and then change `QSPI_GPIO_PAD_TO_IO_OFFSET` to `QSPI_GPIO_IO_TO_PAD_OFFSET`), as it seems _in general_ you're more likely to want to read / write the IO pin than you are to modify the PAD controls?

bootloaders/uart/uart_binary.c
```diff
@@ -49,9 +70,9 @@ void qspi_gpio_set_function(uint gpio, gpio_function_t fn) {
 int main() {
     pico_led_init();
 
-    // SD2 is QSPI GPIO 3, SD3 is QSPI GPIO 4
-    qspi_gpio_set_function(3, GPIO_FUNC_UART_AUX);
-    qspi_gpio_set_function(4, GPIO_FUNC_UART_AUX);
+    // SD2 is UART0 TX, SD3 is UART0 RX (same as the ROM UART bootloader)
+    qspi_gpio_set_function(QSPI_GPIO_SD2, GPIO_FUNC_UART_AUX);
+    qspi_gpio_set_function(QSPI_GPIO_SD3, GPIO_FUNC_UART_AUX);
```
will-v-pi: While here, these should be `GPIO_FUNC1_UART_AUX` (it's the same number, but this is the define from the io_qspi.h header)

```suggestion
    qspi_gpio_set_function(QSPI_GPIO_SD2, GPIO_FUNC1_UART_AUX);
    qspi_gpio_set_function(QSPI_GPIO_SD3, GPIO_FUNC1_UART_AUX);
```

bootloaders/uart/uart_binary.c
will-v-pi: I don't think it matters - if this is moved into the sdk then both should probably be defined, but for this example the PAD is used more than the IO so it is currently the better way round for this example

bootloaders/uart/uart_binary.c
```diff
@@ -14,7 +14,7 @@
 #endif
 
 // Initialize the GPIO for the LED
-void pico_led_init(void) {
+static void pico_led_init(void) {
```
will-v-pi: What's the reasoning for making all these functions static?

will-v-pi:
> It's also curious that the QSPI IO bank has entries for the USB pins (DP and DM) at the start, but they don't have corresponding entries in PAD

This is because the USB PHY is directly connected to the pins, there is no PADs block in the hardware

bootloaders/uart/uart_binary.c
lifeisafractal: agreed on all. I pushed a fix.

bootloaders/uart/uart_binary.c
lifeisafractal: Just general best practice to prevent polluting the global symbol namespace with functions intended to be internal. In practice here since this is a single file binary it doesn't matter, but again it's a good best practice. My general rule is that if a function isn't declared in a header, it should be static.

I made the change as I added a new internal function that I would make static as a best practice, but I can see how this is sorta out of scope for this PR. I'm happy to revert everything to non-static to keep this PR focused if you prefer.

lifeisafractal:
> > It's also curious that the QSPI IO bank has entries for the USB pins (DP and DM) at the start, but they don't have corresponding entries in PAD
> 
> This is because the USB PHY is directly connected to the pins, there is no PADs block in the hardware

That part makes sense to me, although I still don't get why they didn't tack them on the end rather than the beginning, but I guess that's not a huge deal. The fact that the actual QSPI pins are in different orders for the PAD and IO banks is the part that really confuses me :thinking: .

lifeisafractal:
> > Independent of doing that (which is a bigger lift), I think it's worth getting this fix merged here still.
> 
> I totally agree! Sorry if I gave the opposite impression. (I was just wondering if there was an "easy" way to avoid having two "lookup tables" hard-coded inside this example, but you've obviously already looked at this in more detail than me 🙂 )
> 
> ping @will-v-pi (who wrote this uart-bootloader example).
> 
> EDIT: Also ping @matiasilva in case there's any "fiddling" that needs to be done in the auto-generation of the `io_qspi_hw` and `pads_qspi_hw` structs that would make this any "easier" to solve in the long-term. EDIT2: At the very least, perhaps [hardware/structs/io_qspi.h#L196](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/hardware_structs/include/hardware/structs/io_qspi.h#L196) and [hardware/structs/pads_qspi.h#L42](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/hardware_structs/include/hardware/structs/pads_qspi.h#L42) should have a comment listing what order the GPIOs are in within each `io[6]` array?

No prob, and all good! I agree I'm not 100% sure the best way to tackle this on the SDK side, but I think the current solution here is solid enough and won't lead people astray if they steal it for QSPI IO usage which isn't in the SDK currently.

I'm going to drop a comment on raspberrypi/pico-sdk#1799 referencing this PR and throw in some thoughts on what that API should look like given the confusing orders of the QSPI IO and PAD register banks.

lurch:
So to summarise, the issue here is that the QSPI IO registers are in this order:
```console
$ grep "IO_QSPI_GPIO_.*_CTRL_OFFSET" src/rp2350/hardware_regs/include/hardware/regs/io_qspi.h
#define IO_QSPI_GPIO_QSPI_SCLK_CTRL_OFFSET _u(0x00000014)
#define IO_QSPI_GPIO_QSPI_SS_CTRL_OFFSET _u(0x0000001c)
#define IO_QSPI_GPIO_QSPI_SD0_CTRL_OFFSET _u(0x00000024)
#define IO_QSPI_GPIO_QSPI_SD1_CTRL_OFFSET _u(0x0000002c)
#define IO_QSPI_GPIO_QSPI_SD2_CTRL_OFFSET _u(0x00000034)
#define IO_QSPI_GPIO_QSPI_SD3_CTRL_OFFSET _u(0x0000003c)
```
(i.e. SCLK, SS, SD0, SD1, SD2, SD3), whereas the QSPI PADS registers are in this order:
```console
$ grep "PADS_QSPI_GPIO_.*_OFFSET" src/rp2350/hardware_regs/include/hardware/regs/pads_qspi.h
#define PADS_QSPI_GPIO_QSPI_SCLK_OFFSET _u(0x00000004)
#define PADS_QSPI_GPIO_QSPI_SD0_OFFSET _u(0x00000008)
#define PADS_QSPI_GPIO_QSPI_SD1_OFFSET _u(0x0000000c)
#define PADS_QSPI_GPIO_QSPI_SD2_OFFSET _u(0x00000010)
#define PADS_QSPI_GPIO_QSPI_SD3_OFFSET _u(0x00000014)
#define PADS_QSPI_GPIO_QSPI_SS_OFFSET _u(0x00000018)
```
(i.e. SCLK, SD0, SD1, SD2, SD3, SS).

I asked our hardware engineers about this, and they pointed me at our internal ticket discussing it (i.e. this is a known issue). I'm not sure how much of that ticket I can discuss publicly, but something that it mentions is that the QSPI ACCESSCTRL bits:
```console
$ grep "ACCESSCTRL_GPIO_NSMASK1_QSPI_.*SB" src/rp2350/hardware_regs/include/hardware/regs/accessctrl.h 
#define ACCESSCTRL_GPIO_NSMASK1_QSPI_SD_MSB    _u(31)
#define ACCESSCTRL_GPIO_NSMASK1_QSPI_SD_LSB    _u(28)
#define ACCESSCTRL_GPIO_NSMASK1_QSPI_CSN_MSB    _u(27)
#define ACCESSCTRL_GPIO_NSMASK1_QSPI_CSN_LSB    _u(27)
#define ACCESSCTRL_GPIO_NSMASK1_QSPI_SCK_MSB    _u(26)
#define ACCESSCTRL_GPIO_NSMASK1_QSPI_SCK_LSB    _u(26)
```
(i.e. SCK, CSN, SD0, SD1, SD2, SD3), and the QSPI SIO bits:
```console
$ grep "SIO_GPIO_HI_IN_QSPI_.*SB" src/rp2350/hardware_regs/include/hardware/regs/sio.h 
#define SIO_GPIO_HI_IN_QSPI_SD_MSB    _u(31)
#define SIO_GPIO_HI_IN_QSPI_SD_LSB    _u(28)
#define SIO_GPIO_HI_IN_QSPI_CSN_MSB    _u(27)
#define SIO_GPIO_HI_IN_QSPI_CSN_LSB    _u(27)
#define SIO_GPIO_HI_IN_QSPI_SCK_MSB    _u(26)
#define SIO_GPIO_HI_IN_QSPI_SCK_LSB    _u(26)
```
(i.e. SCK, CSN, SD0, SD1, SD2, SD3) are effectively in the same order as the QSPI IO registers.

lurch:
To expand on my previous comment even further: I also noticed that the example-code in https://github.com/raspberrypi/pico-sdk/issues/2417#issuecomment-2802154287 defines:
```c
// The SDK sources do not currently enumerate these very high order GPIOs
#define GPIO58_QSPI_SCLK                58
#define GPIO59_QSPI_SS                  59
#define GPIO60_QSPI_SD0                 60
#define GPIO61_QSPI_SD1                 61
#define GPIO62_QSPI_SD2                 62
#define GPIO63_QSPI_SD3                 63
```
So out of curiosity I asked @andygpz11 where these "GPIO numbers" came from, and after a bit of rummaging he found them:
```console
$ grep "IO_QSPI_.*_CTRL_FUNCSEL_VALUE_SIOB_PROC" src/rp2350/hardware_regs/include/hardware/regs/io_qspi.h
#define IO_QSPI_USBPHY_DP_CTRL_FUNCSEL_VALUE_SIOB_PROC_56 _u(0x05)
#define IO_QSPI_USBPHY_DM_CTRL_FUNCSEL_VALUE_SIOB_PROC_57 _u(0x05)
#define IO_QSPI_GPIO_QSPI_SCLK_CTRL_FUNCSEL_VALUE_SIOB_PROC_58 _u(0x05)
#define IO_QSPI_GPIO_QSPI_SS_CTRL_FUNCSEL_VALUE_SIOB_PROC_59 _u(0x05)
#define IO_QSPI_GPIO_QSPI_SD0_CTRL_FUNCSEL_VALUE_SIOB_PROC_60 _u(0x05)
#define IO_QSPI_GPIO_QSPI_SD1_CTRL_FUNCSEL_VALUE_SIOB_PROC_61 _u(0x05)
#define IO_QSPI_GPIO_QSPI_SD2_CTRL_FUNCSEL_VALUE_SIOB_PROC_62 _u(0x05)
#define IO_QSPI_GPIO_QSPI_SD3_CTRL_FUNCSEL_VALUE_SIOB_PROC_63 _u(0x05)
```

bootloaders/uart/uart_binary.c
will-v-pi: I think it'd be better to remove the `static` changes, so the examples matches the rest of the examples - as you say, they're all single binary files and not designed to be used as libraries, so it doesn't really matter

lifeisafractal: makes sense, I just pushed the change.

will-v-pi (approved):

PR #699, opened 2025-08-20
Add support for pico 2 w

### LINKED ISSUE #695: How can I read bootsel button status in pico 2 w
I wonder if it is possible to use [this example ](https://github.com/raspberrypi/pico-examples/blob/master/picoboard/button/button.c) in pico 2 w and if it isn't possible then what should I do?

A404M:
Refactored code to use pico_status_led instead of the old way ( according to https://github.com/raspberrypi/pico-examples/issues/695#issuecomment-3189236040 )

### DIFF
--- a/picoboard/button/button.c
+++ b/picoboard/button/button.c
@@ -6,6 +6,7 @@
 
 #include <stdio.h>
 #include "pico/stdlib.h"
+#include "pico/status_led.h"
 #include "hardware/gpio.h"
 #include "hardware/sync.h"
 #include "hardware/structs/ioqspi.h"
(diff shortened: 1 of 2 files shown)


peterharperuk (approved):

picoboard/button/button.c
```diff
@@ -58,13 +59,16 @@ bool __no_inline_not_in_flash_func(get_bootsel_button)() {
...
 int main() {
-#ifndef PICO_DEFAULT_LED_PIN
-#warning picoboard/button example requires a board with a regular LED
+#ifndef PICO_STATUS_LED_AVAILABLE
+#warning picoboard/button example requires a board with status LED
 #else
-    gpio_init(PICO_DEFAULT_LED_PIN);
-    gpio_set_dir(PICO_DEFAULT_LED_PIN, GPIO_OUT);
+    int result = status_led_init();
+    hard_assert(result);
```
lurch, line 66: As this example code is primarily concerned with reading the bootsel button, rather than testing the status_led code, we probably don't need all these `hard_assert`s? (which means you can ignore the return-values from the `status_led_` methods too)

peterharperuk: The example won't work very well without status led support - so I'd keep it in.

lurch: I _wasn't_ suggesting removing the status led support; I was just suggesting that e.g.
```c
    int result = status_led_init();
    hard_assert(result);
```
could probably be changed to
```c
    status_led_init();
```
?

peterharperuk: I was suggesting that error handling is good when something fails.

lurch: Fair enough. It felt to me like it was a bit too much of a sledgehammer, but perhaps I was over-reacting. (And apologies for misunderstanding your comment)

PR #700, opened 2025-08-20
chore: Create .gitattributes

siberex:
With this Github will have proper syntax highlighting for `.pio` asm files

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.pio linguist-language=Assembly


kilograham:
can you include a before/after pic; i can't imagine it is "correct"; it might be "better looking"

siberex:
Sure.

Before:
Screenshot 2025-08-21 at 20 44 21
https://github.com/raspberrypi/pico-examples/blob/4c3a3dc0196dd426fddd709616d0da984e027bab/pio/addition/addition.pio

After:
Screenshot 2025-08-21 at 20 44 16
https://github.com/siberex/pico-examples/blob/04eb59b236f6c2e3a8f1075c49f2190fe4bb5f4a/pio/addition/addition.pio

Far from ideal, but better than nothing

peterharperuk (approved):

PR #701, opened 2025-08-22
modified picow access point example to use lwip http server

### LINKED ISSUE #521: Change access point example to use the lwip http server

Louis31423142:


pico_w/wifi/access_point/picow_access_point.c
```diff
@@ -10,292 +10,102 @@
...
-        DEBUG_printf("all done\n");
-        return tcp_close_client_connection(con_state, pcb, ERR_OK);
+    // Now flip the state
+    if (led_state) {
+        // Turn led off
+        status_led_set_state(0);
+    } else {
+        // Turn led on
+        status_led_set_state(1);
     }
```
lurch: Or you could just do:
```suggestion
    status_led_set_state(!led_state);
```
?

pico_w/wifi/access_point/picow_access_point.c
```diff
@@ -10,292 +10,102 @@
...
-        return false;
-    }
-
-    state->server_pcb = tcp_listen_with_backlog(pcb, 1);
-    if (!state->server_pcb) {
-        DEBUG_printf("failed to listen\n");
-        if (pcb) {
-            tcp_close(pcb);
+        default: { // unknown tag
+            printed = 0;
```
lurch: Don't _really_ need this line, given that `printed` is initialised to `0` on line 65 anyway.

pico_w/wifi/access_point/picow_access_point.c
```diff
@@ -10,292 +10,102 @@
...
+            break;
         }
-        return false;
     }
-
-    tcp_arg(state->server_pcb, state);
-    tcp_accept(state->server_pcb, tcp_server_accept);
-
-    printf("Try connecting to '%s' (press 'd' to disable access point)\n", ap_name);
```
lurch, line 267: Perhaps this message still ought to be displayed somewhere?

pico_w/wifi/access_point/content/index.shtml
```diff
@@ -0,0 +1,13 @@
+<html>
+    <head>
+        <title>Pico Access Point</title>
+    </head>
+    <body>
+        <h1>Hello from Pico.</h1>
```
lurch: Perhaps this ought to say "Hello from Pico W" ? :wink:

pico_w/wifi/access_point/content/index.shtml
```diff
@@ -0,0 +1,13 @@
...
+        <p>Led is <!--#led--></p>
+        <p></p>
+        <form action="/switch.cgi" method="get">
+            <button name="switch" value="switch">Turn led <!--#not_led--></button>
```
lurch: Hmmm, rather than having the `switch_cgi_handler` simply reading the LED's current state, and then inverting it, I wonder if it might be more idiomatic to do e.g.:
```html
<button name="switch" value="<!--#not_led-->">Turn led <!--#not_led--></button>
```
and then have the `switch_cgi_handler` read the value of the `switch` parameter? (which I guess would also serve as a useful example of how to deal with GET query-parameters :slightly_smiling_face: )

pico_w/wifi/access_point/picow_access_point.c
```diff
@@ -24,16 +24,14 @@
...
 
-    // Get led state
-    led_state = status_led_get_state();
-    printf("LED state: %d\n", led_state);
-
-    // Now flip the state
-    status_led_set_state(!led_state);
+    // Get desired LED state from GET request and set LED
+    desired_state = !strcmp("ON", pcValue[0]);
+    status_led_set_state(desired_state);
```
lurch: This is a good start, but if we want this to be a "good example" (that people can then extend further) you probably ought to enumerate through all the params, looking for "switch" (and also ignore any unexpected values).
Off the top of my head, _something_ roughly like:
```c
int iParam;
for (iParam = 0; i < iNumParams; i++) {
    if (strcmp("switch", pcParam[iParam]) == 0) {
        if (strcmp("ON", pcValue[iParam]) == 0) {
            status_led_set_state(true);
        } else if (strcmp("OFF", pcValue[iParam]) == 0) {
            status_led_set_state(false);
        }
        break;
    }
}
```

pico_w/wifi/access_point/picow_access_point.c
```diff
@@ -10,292 +10,94 @@
...
-        err_t err = tcp_close(client_pcb);
-        if (err != ERR_OK) {
-            DEBUG_printf("close failed %d, calling abort\n", err);
-            tcp_abort(client_pcb);
-            close_err = ERR_ABRT;
-        }
-        if (con_state) {
-            free(con_state);
+static absolute_time_t wifi_connected_time;
+bool led_state;
```
lurch, line 25: Does `led_state` need to be initialised somewhere? i.e. maybe you need `led_state = status_led_get_state();` somewhere in `main()` ?

pico_w/wifi/access_point/picow_access_point.c
```diff
@@ -10,292 +10,94 @@
...
+bool complete;
+
+static const char *switch_cgi_handler(int iIndex, int iNumParams, char *pcParam[], char *pcValue[]) {
+    // Get desired LED state from GET request and set LED
+    for (int iParam = 0; iParam < iNumParams; iParam++) {
+        if (strcmp("switch", pcParam[iParam]) == 0) {
+            if (strcmp("ON", pcValue[iParam]) == 0) {
+                status_led_set_state(true);
+                led_state = true;
```
lurch, line 34: @peterharperuk Is it ever possible for a call to `status_led_set_state(true)` to **not** turn on the LED?

pico_w/wifi/access_point/picow_access_point.c
```diff
@@ -96,6 +96,9 @@ int main() {
         return 1;
     }
 
+    // Initialise led state
+    led_state = status_led_get_state();
```
lurch, line 105: I think there also needs to be a call to either [`status_led_init`](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_pico_status_led_1gaf83057f40dddf57bbaa9845c2c56201c) or [`status_led_init_with_context`](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_pico_status_led_1ga1d39f6237f4c2eb0b7d89938f379bd41) ? (Sorry I forgot to think of this earlier :facepalm: )

lurch:
I've not tested this myself, but it looks pretty good to my eyes :+1:

pico_w/wifi/access_point/picow_access_point.c
peterharperuk: Not in this case. It will only not work if the status led is not supported on the board or you're using colored leds and there isn't enough pio space to drive the led,

peterharperuk:
This is good. But it's broken the captive portal functionality. I think this is an issue with the makefsdata.py tool. I'll assign an issue which you could have a go at fixing?

peterharperuk:
See https://github.com/raspberrypi/pico-sdk/issues/2646
Waiting for fix here https://github.com/raspberrypi/pico-sdk/issues/2646

peterharperuk (approved):

PR #704, opened 2025-08-29
Skip uart bootloader when no mbedtls

will-v-pi:
This requires the `picotool seal` command which is only available when compiled with mbedtls

Also improve message, in line with message in pico_w/wifi/CMakeLists.txt

### DIFF
--- a/bootloaders/CMakeLists.txt
+++ b/bootloaders/CMakeLists.txt
@@ -1,8 +1,8 @@
-add_subdirectory_exclude_platforms(uart host rp2040)
-
 if (TARGET pico_mbedtls)
+    # note uart bootloader doesn't use Mbed TLS, but it requires picotool with Mbed TLS, so the above check is a good proxy
+    add_subdirectory_exclude_platforms(uart host rp2040)
     add_subdirectory_exclude_platforms(encrypted host rp2040 rp2350-riscv)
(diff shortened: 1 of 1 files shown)


bootloaders/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
-add_subdirectory_exclude_platforms(uart host rp2040)
-
 if (TARGET pico_mbedtls)
+    # note uart bootloader doesn't use Mbed TLS, but it requires picotool with Mbed TLS, so the above check is a good proxy
```
lurch, line 2: ...unless the user is using the Pico VS Code extension, as presumably the picotool in pico-sdk-tools **was** compiled with Mbed TLS ? But that might be an edge-case not worth worrying about.

will-v-pi: The Pico VS Code extension also always clones the MbedTLS submodule when cloning the SDK, so that edge case wouldn't occur anyway

This check is used here and elsewhere, as the best approximation for checking whether picotool was compiled without MbedTLS, because we're assuming that you use the same pico-sdk to build picotool and pico-examples

peterharperuk (approved):

kilograham (approved):

PR #709, opened 2025-09-08
Added 24LC32 EEPROM I2C example

Louis31423142:


i2c/24lc32_i2c/24lc32_i2c_lib.c
```diff
@@ -0,0 +1,85 @@
...
+            return true;
+        }
+        sleep_ms(1);
+    }
+    printf("ack timeout\n");
+    return false;
+}
+
+// Read num bytes of memory starting at given address into result
+void read_eeprom(uint16_t address, uint8_t *result, uint16_t num, uint8_t i2c_addr) {
```
lurch: Perhaps this should check that `(address + num - 1) < 4096` ? (and similarly in the other read / write functions)
Might also be worth doing e.g. `#define EEPROM_SIZE 4096` ?

i2c/24lc32_i2c/24lc32_i2c_lib.c
```diff
@@ -0,0 +1,85 @@
...
+    i2c_write_blocking(i2c_default, i2c_addr, buf, 2 + num, false);
+
+    // ack_poll to wait for write to be complete
+    ack_poll(i2c_addr);
+    return true;
+}
+
+// Initialise i2c for given gpio at given frequency
+void eeprom_init(int sda_pin, int scl_pin, int freq, i2c_inst_t *i2c_instance) {
+    i2c_init(i2c_default, freq);
```
lurch: You pass in `i2c_instance`, but then this function (and all the others) just use `i2c_default` anyway! :facepalm:  (I guess you need to pass in an `i2c_inst_t` to _all_ the functions?)

i2c/24lc32_i2c/24lc32_i2c_lib.c
```diff
@@ -0,0 +1,85 @@
...
+void byte_write_eeprom(uint16_t address, uint8_t data, uint8_t i2c_addr) {
+    uint8_t byte1 = (uint8_t)(address >> 8);
+    uint8_t byte2 = (uint8_t)(address);
+
+    uint8_t buf[3] = {byte1, byte2, data};
+
+    i2c_write_blocking(i2c_default, i2c_addr, buf, 3, false);
+
+    // ack_poll to wait for write to be complete
+    ack_poll(i2c_addr);
```
lurch: Should this be checking the return-value of `ack_poll` ?

i2c/24lc32_i2c/24lc32_i2c_lib.h
```diff
@@ -0,0 +1,12 @@
+#ifndef PICO_I2C_EEPROM_LIB_H
+#define PICO_I2C_EEPROM_LIB_H
+
+#include "hardware/i2c.h"
+
+void read_eeprom(uint16_t address, uint8_t *result, uint16_t num, uint8_t i2c_addr);
+void byte_write_eeprom(uint16_t address, uint8_t data, uint8_t i2c_addr);
+bool page_write_eeprom(uint16_t address, uint8_t *data, uint8_t num, uint8_t i2c_addr);
```
lurch: Only a gut-feeling, but I wonder if `uint8_t i2c_addr` ought to be the first parameter? (Or the second parameter, if you also add `i2c_inst_t *i2c_instance`?)
Probably worth checking with @peterharperuk what style he prefers.

i2c/24lc32_i2c/READ_ME.adoc
```diff
@@ -0,0 +1,30 @@
+= Reading and writing 24LC32 EEPROM via I2C
+
+This example code demonstrates how to read and write data from 24LC32 EEPROM. It contains a library with functions for reading data, writing single bytes, and writing pages, and an example which uses these functions.
```
lurch:
```suggestion
This example code demonstrates how to read and write data from 24LC32 EEPROM. It contains a library with functions for reading data, writing single bytes, and writing pages; and an example which uses these functions.
```
(otherwise it looks a bit weird with two "and"s :slightly_smiling_face: )

i2c/24lc32_i2c/example.c
```diff
@@ -0,0 +1,38 @@
...
+
+    for (int i = 0; i < 32; i++) {
+        printf("read: %i, expected: %i\n", result[i], i);
+        if (result[i] != i) {
+            printf("Unexpected read!\n");
+            return 1;
+        }
+    }
+    
+    printf("Test passed!\n");
```
lurch: I guess this test would also pass if the EEPROM happened to have the correct contents (from a prior test-run), but your `byte_write_eeprom` and `page_write_eeprom` functions were totally broken :wink: 
Perhaps it's also worth doing a second test-pass, where you write locations 0..31 with values 31..0 ?

i2c/24lc32_i2c/24lc32_i2c_lib.c
```diff
@@ -0,0 +1,85 @@
...
+    return false;
+}
+
+// Read num bytes of memory starting at given address into result
+void read_eeprom(uint16_t address, uint8_t *result, uint16_t num, uint8_t i2c_addr) {
+    for (int i = 0; i < num; i++) {
+        uint8_t byte1 = (uint8_t)(address >> 8);
+        uint8_t byte2 = (uint8_t)(address);
+
+        uint8_t buf[2] = {byte1, byte2};
```
lurch, line 36: @peterharperuk Should these variables be declared outside of the `for` loop, or does that not really matter?

lurch:
Could you rename `READ_ME.aoc` to `README.adoc` please?

lurch:
Eeek, looks like you've overwritten `24lc32_i2c_lib.c` with the contents of `example.c` ?!

i2c/24lc32_i2c/24lc32_i2c_lib.c
```diff
@@ -0,0 +1,100 @@
...
+    return ack_poll(i2c_addr, i2c_instance);
+}
+
+// Write a block of data to eeprom 
+bool page_write_eeprom(uint16_t address, uint8_t *data, uint8_t num, uint8_t i2c_addr, i2c_inst_t *i2c_instance) {
+    if (num > MAX_PAGE_WRITE) {
+        printf("ERROR: Tried to write more than 32 bytes.\n");
+        return false;
+    } else if ((address + num - 1) > EEPROM_SIZE) {
+        printf("ERROR: Tried to write to memory outside EEPROM's range.\n");
```
lurch, line 71: Missing a `return false;` ?

i2c/24lc32_i2c/24lc32_i2c_lib.h
```diff
@@ -0,0 +1,12 @@
+#ifndef PICO_I2C_EEPROM_LIB_H
+#define PICO_I2C_EEPROM_LIB_H
+
+#include "hardware/i2c.h"
+
+bool ack_poll(uint8_t i2c_addr, i2c_inst_t *i2c_instance);
+bool read_eeprom(uint16_t address, uint8_t *result, uint16_t num, uint8_t i2c_addr, i2c_inst_t *i2c_instance);
+bool byte_write_eeprom(uint16_t address, uint8_t data, uint8_t i2c_addr, i2c_inst_t *i2c_instance);
+bool page_write_eeprom(uint16_t address, uint8_t *data, uint8_t num, uint8_t i2c_addr, i2c_inst_t *i2c_instance);
+void eeprom_init(int sda_pin, int scl_pin, int freq, i2c_inst_t *i2c_instance);
```
lurch: Might be worth copying your function-comments from `24lc32_i2c_lib.c` into here too? :slightly_smiling_face:

i2c/24lc32_i2c/24lc32_i2c_lib.c
```diff
@@ -0,0 +1,101 @@
...
+
+        address++;
+    }
+    
+    return true;
+}
+
+// Write a single byte of data at a specified address in memory
+bool byte_write_eeprom(uint16_t address, uint8_t data, uint8_t i2c_addr, i2c_inst_t *i2c_instance) {
+    if (address > EEPROM_SIZE) {
```
lurch: I guess this needs to be `if (address >= EEPROM_SIZE) {` ? Because I'm assuming the EEPROM addresses are 0..4095 ?

i2c/24lc32_i2c/24lc32_i2c_lib.c
```diff
@@ -0,0 +1,101 @@
...
+    i2c_write_blocking(i2c_instance, i2c_addr, buf, 3, false);
+
+    // ack_poll to wait for write to be complete
+    return ack_poll(i2c_addr, i2c_instance);
+}
+
+// Write a block of data to eeprom 
+bool page_write_eeprom(uint16_t address, uint8_t *data, uint8_t num, uint8_t i2c_addr, i2c_inst_t *i2c_instance) {
+    if (num > MAX_PAGE_WRITE) {
+        printf("ERROR: Tried to write more than 32 bytes.\n");
```
lurch: Might be nice to make this also use the `MAX_PAGE_WRITE` define?
According to https://stackoverflow.com/questions/47346133/how-to-use-a-define-inside-a-format-string you should be able to do this by adding
```c
// Stringification macros
#define STR_(X) #X
#define STR(X) STR_(X)
```
and then changing this line to:
```c
printf("ERROR: Tried to write more than " STR(MAX_PAGE_WRITE) " bytes.\n");
```

i2c/24lc32_i2c/example.c
```diff
@@ -0,0 +1,59 @@
...
+#include "hardware/i2c.h"
+#include "24lc32_i2c_lib.h"
+
+// Default I2C address of 24LC32
+#define I2C_ADDRESS 0x50
+
+int main() {
+    stdio_init_all();
+
+    eeprom_init(PICO_DEFAULT_I2C_SDA_PIN, PICO_DEFAULT_I2C_SCL_PIN, 400*1000, i2c_default);
```
lurch: If you're feeling up to it, it might be nice to also add lines like these https://github.com/raspberrypi/pico-examples/blob/master/i2c/bmp280_i2c/bmp280_i2c.c#L214-L221 so that a) you get a suitable compile-time warning if you try to build it for a [board](https://github.com/raspberrypi/pico-sdk/tree/master/src/boards/include/boards) that doesn't have  `PICO_DEFAULT_I2C_SDA_PIN` and `PICO_DEFAULT_I2C_SCL_PIN` defined and b) running `picotool info` against the built UF2 will tell you what pins it expects to use.

For a similar reason (i.e. `picotool info`), it might also be nice to add https://github.com/raspberrypi/pico-examples/blob/master/i2c/bmp280_i2c/CMakeLists.txt#L11-L12 to your `CMakeLists.txt` :slightly_smiling_face:

i2c/24lc32_i2c/24lc32_i2c_lib.c
```diff
@@ -0,0 +1,101 @@
+#include <stdio.h>
+#include "pico/stdlib.h"
+#include "hardware/i2c.h"
+
+// This is the max number of bytes which can be written at once during a page write
+#define MAX_PAGE_WRITE 32
+
+// This is the 24LC32's capacity in bytes
+#define EEPROM_SIZE 4096
```
lurch, line 9: @peterharperuk Would it make sense for these to be in the `.h` file rather than the `.c` file? :man_shrugging:

lurch:
Looks pretty good to me now :+1:

README.md
```diff
@@ -159,6 +159,7 @@ App|Description
 [ht16k33_i2c](i2c/ht16k33_i2c) | Drive a 4 digit 14 segment LED with an HT16K33.
 [slave_mem_i2c](i2c/slave_mem_i2c) | i2c slave example where the slave implements a 256 byte memory.
 [slave_mem_i2c_burst](i2c/slave_mem_i2c) | i2c slave example where the slave implements a 256 byte memory. This version inefficiently writes each byte in a separate call to demonstrate read and write burst mode.
+[24lc32_i2c](i2c/24lc32_i2c) | Read and write data to 24LC32 EEPROM via I2C.
```
kilograham, line 162: arguably missing an "a "

i2c/24lc32_i2c/24lc32_i2c_lib.c
peterharperuk: There may be a way to query the hardware for these values rather than hard code them. But maybe this is good enough for this example

i2c/24lc32_i2c/24lc32_i2c_lib.c
peterharperuk: It looks like it makes more sense as it is here?

PR #498, opened 2024-05-25
mqtt_client_test

ellebi2000:
This PR introduces functionality for temperature monitoring and LED control on a Raspberry Pi Pico W using MQTT. The key features include:

Temperature Monitoring:

    Reads onboard temperature using the ADC.
    Converts ADC readings to Celsius or Fahrenheit.
    Publishes temperature to the /temperature MQTT topic every 5 seconds.
LED Control:

    Controls the onboard LED via MQTT messages received on the /led topic.
    Turns the LED on/off based on "On" or "Off" messages.
    Publishes LED state to the /state MQTT topic.

WiFi and MQTT Configuration:

    Connects to WiFi using predefined SSID and password.
    Connects to the MQTT broker with the given IP address and port.
    Configures an MQTT client with a predefined client ID and keep-alive period.

MQTT Callbacks:

    mqtt_incoming_data_cb: Handles incoming data and controls the LED.
    mqtt_incoming_publish_cb: Handles incoming publish events.
    mqtt_connection_cb: Manages MQTT connection status and subscribes to the /led topic.

PR #712, opened 2025-10-02
Add .md to README and clarify what example does

mjcross:
1. For pico_w/wifi/mqtt, rename README to README.md to improve how the content renders in an IDE or when browsing the online repository.
2. Clarify that the example doesn't re-publish the temperature reading if it hasn't changed; so that users can understand why the updates can appear sporadic.
3. Document how the led status can be checked by subscribing to /led/status

### DIFF
--- a/pico_w/wifi/mqtt/mqtt_client.c
+++ b/pico_w/wifi/mqtt/mqtt_client.c
@@ -243,6 +243,8 @@ static void mqtt_connection_cb(mqtt_client_t *client, void *arg, mqtt_connection
         if (!state->connect_done) {
             panic("Failed to connect to mqtt server");
         }
+        // note that the main() loop will soon terminate because mqtt_client_is_connected()
+        // will return false.  
     }
     else {
         panic("Unexpected status");
@@ -368,7 +370,13 @@ int main(void) {
         panic("dns request failed");
     }
 
+    // We are not in a callback but we can get away with calling mqtt_client_is_connected()
+    // because it's a read-only operation
     while (!state.connect_done || mqtt_client_is_connected(state.mqtt_client_inst)) {
+        // As supplied the example configures cyw43_arch for thread_safe_background operation
+        // by linking `ico_cyw43_arch_lwip_threadsafe_background` in CMakeLists.txt, so the 
+        // following two lines are unnecessary (but do no harm). However you will need them 
+        // if you reconfigure the build to use cyw43_arch in polling mode.
         cyw43_arch_poll();
         cyw43_arch_wait_for_work_until(make_timeout_time_ms(10000));
     }
--- a/pico_w/wifi/mqtt/mqtt_client.c
+++ b/pico_w/wifi/mqtt/mqtt_client.c
@@ -259,8 +259,10 @@ static void start_client(MQTT_CLIENT_DATA_T *state) {
     const int port = MQTT_PORT;
     INFO_printf("Warning: Not using TLS\n");
 #endif
-
+    // mqtt_client_new() has LWIP_ASSERT_CORE_LOCKED(), so we should protect this call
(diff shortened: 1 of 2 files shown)


pico_w/wifi/mqtt/mqtt_client.c
```diff
@@ -368,7 +370,13 @@ int main(void) {
         panic("dns request failed");
     }
 
+    // We are not in a callback but we can get away with calling mqtt_client_is_connected()
+    // because it's a read-only operation
     while (!state.connect_done || mqtt_client_is_connected(state.mqtt_client_inst)) {
+        // As supplied the example configures cyw43_arch for thread_safe_background operation
+        // by linking `ico_cyw43_arch_lwip_threadsafe_background` in CMakeLists.txt, so the
```
lurch: Typo here? Did you mean `pico_cyw43_arch_lwip_threadsafe_background` ?

peterharperuk:
Looks good. Just the one typo pointed out by @lurch

lurch:
FYI #498 is where this example was first added.

mjcross:
Thanks @peterharperuk and well spotted @lurch - fixed in my [a23bca5](https://github.com/mjcross/pico-examples/commit/a23bca58ddf7b6de1f9e2d5e011b5e521c12d95b). One of these days I'll manage to get a PR through without a typo 8-)

mjcross:
If the dns query returns immediately then start_client() is called directly from main(); so shouldn't the call to mqtt_client_new() be protected? The lwIP implementation has LWIP_ASSERT_CORE_LOCKED

mjcross:
With the greatest respect, although this example is very thorough it is also quite hard to follow. 
Might I suggest that In the spirit of the other examples it would be clearer to use global variables and move some of the work outside the callbacks?

peterharperuk:
> If the dns query returns immediately then start_client() is called directly from main(); so shouldn't the call to mqtt_client_new() be protected? The lwIP implementation has LWIP_ASSERT_CORE_LOCKED

Yes

peterharperuk:
If you have suggestions for other improvements and fixes. I would suggest raising them as separate pull requests.

pico_w/wifi/mqtt/mqtt_client.c
```diff
@@ -368,7 +372,13 @@ int main(void) {
         panic("dns request failed");
     }
 
+    // We are not in a callback but we can get away with calling mqtt_client_is_connected()
+    // because it's a read-only operation
     while (!state.connect_done || mqtt_client_is_connected(state.mqtt_client_inst)) {
+        // As supplied the example configures cyw43_arch for thread_safe_background operation
+        // by linking `pico_cyw43_arch_lwip_threadsafe_background` in CMakeLists.txt, so the 
+        // following two lines are unnecessary (but do no harm). However you will need them 
+        // if you reconfigure the build to use cyw43_arch in polling mode.
```
lurch: If the lines are needed in one mode but not in a different mode, maybe wrapping them in an `#ifdef` would make that more obvious? (Which I think is a pattern I've seen used in other examples?)

mjcross: Agreed - see [12bbe5c](https://github.com/raspberrypi/pico-examples/pull/712/commits/12bbe5c154cefef55d05e52896ff1e2cf6d211ef)

mjcross:
> > If the dns query returns immediately then start_client() is called directly from main(); so shouldn't the call to mqtt_client_new() be protected? The lwIP implementation has LWIP_ASSERT_CORE_LOCKED
> 
> Yes

The code as stands also calls mqtt_client_is_connected(), which the lwIP code also has the CORE_LOCKED assertion. Not sure whether that's safe or not?

peterharperuk (approved):

PR #714, opened 2025-10-14
Add PIO 7-segment display example

mjcross:
Add an example showing how to use the PIO to control a multiplexed 7-segment display.
@lurch this re-creates #375 as discussed, due to loss of the original repo


pio/seven_segment/7_segment.c
```diff
@@ -0,0 +1,38 @@
+#include <stdio.h>
+#include "pico/stdlib.h"
+#include "7_segment_lib.h"
+
+
+const PIO pio = pio0;
+const uint first_segment_pin = 8;   // gpio 15-8 = segments E,D,B,G,A,C,F,dp
+const uint first_digit_pin = 16;    // gpio 19-16 = common anodes 4,3,2,1
+
+//                      EDBGACF.       bit ordering depends on your display and wiring
```
lurch: It might be worth including an ASCII-art diagram showing which of the segments these letters correspond to?

pio/seven_segment/README.adoc
```diff
@@ -0,0 +1,22 @@
+= Driving a 7 segment LED display
+This example demonstrates how a PIO state machine can be used to control a four digit multiplexed 7-segment display with the Raspberry Pi Pico (RP2040).
+
+Multiplexed displays work by rapidly displaying each digit in turn and the PIO provides an excellent way to relieve the CPU of this time-consuming task.
+
+The total current required is typically more than the GPIO pins can supply so you're likely to need an external drive circuit like the one below.
+
+The PIO code uses four `side-set` pins to control the digit multiplex lines and displays the segment patterns received on the FIFO. It uses a non-blocking PULL to keep showing the same segments until the CPU sends new data.
+
+The provided example uses spells out the word **Pico** and then counts from 0 to 9999.
```
lurch:
```suggestion
The provided example spells out the word **Pico** and then counts from 0 to 9999.
```

pio/seven_segment/pio_7_segment_library/7_segment_lib.c
```diff
@@ -0,0 +1,75 @@
...
+    0b11011110,     // 6
+    0b00101100,     // 7
+    0b11111110,     // 8
+    0b01111110      // 9
+};
+
+
+// converts an integer into a 32-bit word representing up to four 7-segment digits
+//
+uint32_t int_to_seven_segment (int num) {
```
lurch, line 37: I guess this ought to check that `num <= 9999`? And if you were feeling clever, you could also handle numbers down to `-999` ? :wink:

pio/seven_segment/pio_7_segment_library/7_segment_lib.c
```diff
@@ -0,0 +1,75 @@
...
+    uint32_t word = 0;
+    if (num == 0) {
+        word = segments[0];
+    } else {
+        for (int bitshift = 0; bitshift < 32 && num > 0; bitshift += 8) {
+            word |= segments[num % 10] << bitshift;
```
lurch: I think this might be slightly more readable if you rearranged it to:
```suggestion
        for (int digit = 0; digit < 4 && num > 0; digit++) {
            word |= segments[num % 10] << (digit * 8);
```
?

pio/seven_segment/7_segment.c
mjcross: good idea - included in [0c1355e](https://github.com/raspberrypi/pico-examples/pull/714/commits/0c1355eeed5d0bbd3e4b927f30bf2e2400e6c3a4)

pio/seven_segment/README.adoc
mjcross: Well spotted - fixed in [0c1355e](https://github.com/raspberrypi/pico-examples/pull/714/commits/0c1355eeed5d0bbd3e4b927f30bf2e2400e6c3a4)

pio/seven_segment/pio_7_segment_library/7_segment_lib.c
mjcross: 1) OK [0c1355e](https://github.com/raspberrypi/pico-examples/pull/714/commits/0c1355eeed5d0bbd3e4b927f30bf2e2400e6c3a4) now returns an 'E' symbol for numbers >9999. 
2) While it would indeed be nice to handle negative numbers... and floats... and maybe octal/hex I think I'll leave that as an exercise for the user LoL

pio/seven_segment/pio_7_segment_library/7_segment_lib.c
mjcross: Hmmm - sorry to disagree but I think mine's more readable (and I know it works)... personal taste I guess, unless you feel strongly?

pio/seven_segment/pio_7_segment_library/7_segment_lib.c
lurch: If it _doesn't_ handle negative numbers, perhaps you ought to also display an `E` if `num < 0`? Alternatively, you could change the function to only take `unsigned int`? :thinking:

pio/seven_segment/pio_7_segment_library/7_segment_lib.c
lurch: > personal taste I guess

Fair enough :+1:

pio/seven_segment/pio_7_segment_library/7_segment_lib.c
mjcross: Agreed. Fixed in [4542b5a](https://github.com/raspberrypi/pico-examples/pull/714/commits/4542b5a8e8238178cadde525c1b5d06e32e36e6f)

lurch:
@mjcross I've not tested this, as I don't have a bagful of transistors handy, but does this work?
```c
// Simple example of how to convert an integer between -999 and 9999
// into a 32-bit word representing up to four 7-segment digits.
//
uint32_t int_to_seven_segment (int num) {
    uint32_t word = 0;
    if (num < -999 || num > 9999) {
        // number out of range, display 'E' symbol
        //       EDBGACF.
        word = 0b11011010
    } else { 
        if (num == 0) {
            word = segments[0];
        } else {
            bool negative = num < 0;
            if (negative) {
                num *= -1;
            }
            int bitshift;
            for (bitshift = 0; bitshift < 32 && num > 0; bitshift += 8) {
                word |= segments[num % 10] << bitshift;
                num /= 10;
            }
            if (negative) {
                bitshift += 8;
                // display '-' symbol
                //        EDBGACF.
                word |= 0b00010000 << bitshift;
            }
        }
    }
    return word;
}
```

mjcross:
@lurch yes: tested a couple of years ago when I wrote it, with tiny 7-segment displays that didn't need driver transistors ;-) The project I wrote it for needed up using an OLED display instead so it never got used.
The only thing I changed was the ordering of the bit-patterns in the lookup tables, just to make the wiring look simpler in the circuit diagram.
I'll do a quick check on that function later today, just to be sure though.

mjcross:
TBH if there's any bit of the project likely to cause issues it's probably the CMakeLists.txt files: I see at some point several of the example builds were adjusted to use `add_subdirectory_exclude_platforms()` so I've copied that, but I must admit I have no idea what it does.

lurch:
> yes: tested a couple of years ago when I wrote it

Ahhh, sorry for the miscommunication - I wasn't asking "Does your function work?", I was asking "Does my suggested change to your function, to add support for displaying negative numbers, work?" :joy: 

> to use `add_subdirectory_exclude_platforms()` so I've copied that, but I must admit I have no idea what it does.

It's so that if an example uses a hardware-feature that a particular board / chip / platform doesn't support, we don't try building that example for that chip. See e.g. https://github.com/raspberrypi/pico-examples/blob/master/rtc/CMakeLists.txt where we prevent the RTC examples from building for the RP2350 (as it doesn't have an RTC) or https://github.com/raspberrypi/pico-examples/blob/master/sha/CMakeLists.txt where we prevent the SHA examples from building for the RP2040 (as it doesn't have SHA256 hardware).

P.S. I've just realised that you also need to edit the top-level `README.md` in your PR, to add a link to your new example to https://github.com/raspberrypi/pico-examples?tab=readme-ov-file#pio :slightly_smiling_face:

mjcross:
Ahh - sorry! Didn't read carefully enough. That code looks nice - I'll check it this afternoon: tks!

mjcross:
I like it! Turns out you don't need the `bitshift += 8;`  otherwise the minus sign falls off the end LoL

mjcross:
You can simulate it with this slightly hacky code...

```
#include <stdio.h>
#include <stdlib.h>
#include <stdbool.h>
#include <string.h>

// segment patterns for digits 0-9
//
// By convention the segments are labelled as follows:
//
//  AAAA
// F    B
// F    B
//  GGGG
// E    C
// E    C
//  DDDD  .
//
// the bit ordering should match the way you connect up the GPIOs. Here it is
//    EDBGACF.
static const uint8_t segments[] = {
    0b11101110,     // 0
    0b00100100,     // 1
    0b11111000,     // 2
    0b01111100,     // 3
    0b00110110,     // 4
    0b01011110,     // 5
    0b11011110,     // 6
    0b00101100,     // 7
    0b11111110,     // 8
    0b01111110      // 9
};

// for display
static char rows[7][50];

// Simple example of how to convert an integer between -999 and 9999
// into a 32-bit word representing up to four 7-segment digits.
//
uint32_t int_to_seven_segment (int num) {
    uint32_t word = 0;
    if (num < -999 || num > 9999) {
        // number out of range, display 'E' symbol
        //       EDBGACF.
        word = 0b11011010;
    } else { 
        if (num == 0) {
            word = segments[0];
        } else {
            bool negative = num < 0;
            if (negative) {
                num *= -1;
            }
            int bitshift;
            for (bitshift = 0; bitshift < 32 && num > 0; bitshift += 8) {
                word |= segments[num % 10] << bitshift;
                num /= 10;
            }
            if (negative) {
                //bitshift += 8;
                // display '-' symbol
                //        EDBGACF.
                word |= 0b00010000 << bitshift;
            }
        }
    }
    return word;
}



void append_digit(uint8_t word) {
    static char temp_row[7][10];

    char e = word & 0b10000000 ? '|': ' ';
    char d = word & 0b01000000 ? '-': ' ';
    char b = word & 0b00100000 ? '|': ' ';
    char g = word & 0b00010000 ? '-': ' ';
    char a = word & 0b00001000 ? '-': ' ';
    char c = word & 0b00000100 ? '|': ' ';
    char f = word & 0b00000010 ? '|': ' ';

    snprintf(temp_row[0], sizeof(temp_row[0]), " %c%c%c%c   ", a, a, a, a);
    snprintf(temp_row[1], sizeof(temp_row[1]), "%c    %c  ", f, b);
    snprintf(temp_row[2], sizeof(temp_row[2]), "%c    %c  ", f, b);
    snprintf(temp_row[3], sizeof(temp_row[3]), " %c%c%c%c   ", g, g, g, g);
    snprintf(temp_row[4], sizeof(temp_row[4]), "%c    %c  ", e, c);
    snprintf(temp_row[5], sizeof(temp_row[5]), "%c    %c  ", e, c);
    snprintf(temp_row[6], sizeof(temp_row[6]), " %c%c%c%c   ", d, d, d, d);

    for (int i=0; i<7; i+=1) {
        strlcat(rows[i], temp_row[i], sizeof(rows[i]) - 1);
    }
}

void display_all(uint32_t word) {
    for (int i=0; i<7; i+=1) {
        rows[i][0] = '\0';
    }

    append_digit(word >> (3 * 8));
    append_digit(word >> (2 * 8));
    append_digit(word >> (1 * 8));
    append_digit(word);

    for (int i=0; i<7; i+=1) {
        puts(rows[i]);
    }
}


int main(int argc, char **argv) {

    display_all(int_to_seven_segment(-999));
    return (0);
}
```

mjcross:
The output is actually rather cute
```
                 ----    ----   
             |       |       |  
             |       |       |  
 ----            ----    ----   
             |  |            |  
             |  |            |  
                 ----    ----  
```

mjcross:
Hope the complex-looking circuit doesn't put people off: it's a really good use-case for the PIO. 
You're right that using a driver chip would have been a better idea than discrete transistors...

magy00:
Maybe add a dimmable version by adding something like
```
    mov y, isr       side 0b0000
delay:
    jmp y--, delay   side 0b0000
```
after `mov` (might require reshuffle of delays and placing `.wrap`/`.wrap_target` so that they could be changed using `pio_sm_set_wrap()` to skip these two instructions for full brightness, etc.)?

mjcross:
That’s neat :-)

peterharperuk:
I think you need to add more detail about the display you have written this for. A part number perhaps? A quick search mostly finds common cathode displays

mjcross:
> I think you need to add more detail about the display you have written this for. A part number perhaps? A quick search mostly finds common cathode displays

I agree - the parts I used were recovered from a (very) old calculator and as you say, common cathode displays are much more normal these days. I think I'll kill two birds with one stone and re-work this with a darlington driver chip (as previously suggested by @lurch IIRC) at the same time.

lurch:
Just in case it's useful: https://raspi.tv/2015/how-to-drive-a-7-segment-display-directly-on-raspberry-pi-in-python

mjcross:
> Just in case it's useful: https://raspi.tv/2015/how-to-drive-a-7-segment-display-directly-on-raspberry-pi-in-python

Very helpful, thanks 🙂 ... what that guy may not have quite worked out mind you, is that with 100 Ohm resistors he's probably driving 15mA per segment, so if he illuminates all 8 he's going to clobber the GPIO he's using for 'common' with 120mA #facepalm

mjcross:
Waiting for some components to arrive for testing

PR #716, opened 2025-10-26
add example pico_w/wifi/ntp_system_time

mjcross:
A proposed new NTP example that differs from the existing `pico_w/wifi/ntp_client` in the following ways:

1. uses the SNTP app already provided by lwIP (in threadsafe background mode)
2. uses `pico_aon_timer` to create a time-of-day clock that can be read asynchronously from user code
3. illustrates the use of a POSIX timezone (TZ) to convert UTC to local time with daylight saving
4. includes a README.md to explain what the example does and how to use it

The program connects to Wi-Fi, initialises the lwIP SNTP app, syncs the `pico_aon_timer` to NTP and then enters a loop to display the local time in London (or UTC, or whichever timezone the user defines); resynchronising to ntp.pool.org in the background every hour.

I suggest it's quite an important use case for the Pico-W, and illustrates a couple of features users may find helpful (such as how to create and use a POSIX timezone).


pico_w/wifi/ntp_system_time/README.md
```diff
@@ -0,0 +1,92 @@
+# Overview
+
+Creates a time of day clock that periodically synchronises itself to Internet time servers using simple NTP (see [RFC 4330](https://datatracker.ietf.org/doc/html/rfc4330)).
+
+The example connects to Wi-Fi and displays the local time in the UK or another timezone that you specify, synchronised once an hour to one of the servers from [pool.ntp.org](https://www.ntppool.org/en/). 
+
+Uses the SNTP application provided by lwIP and the Pico 'always-on timer' _(RTC on Pico/rp2040, powman timer on Pico-2/rp2350)_.
```
lurch: Capital letters for RP2040 and RP2350 please :slightly_smiling_face:

README.md
```diff
@@ -200,6 +200,7 @@ App|Description
 [picow_blink_fast_clock](pico_w/wifi/blink) | Blinks the on-board LED (which is connected via the WiFi chip) with a faster system clock to show how to reconfigure communication with the WiFi chip at build time under those circumstances.
 [picow_iperf_server](pico_w/wifi/iperf) | Runs an "iperf" server for WiFi speed testing.
 [picow_ntp_client](pico_w/wifi/ntp_client) | Connects to an NTP server to fetch and display the current time.
+[picow_ntp_system_time](pico_w/wifi/ntp_system_time) | Creates a background time of day clock that periodically updates itself from a pool of NTP servers, and uses it to display local time.
```
lurch: "background time-of-day clock" might be easier to read?

pico_w/wifi/ntp_system_time/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Running the example
+
+Provide the SSID and password of your Wi-Fi network by editing `CmakeLists.txt` or on the command line; then build and run the example as usual.
```
lurch: `CmakeLists.txt` -> `CMakeLists.txt`

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -0,0 +1,119 @@
...
+    struct tm tm;
+
+    while (true) {
+
+        if(aon_timer_is_initialised) {
+
+            // read the current time as UTC seconds and ms since the epoch
+            get_time_utc(&ts);
+
+            // if you simply want to display the local time you could now do so with
```
lurch: "local time" -> "UTC time" ?

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -0,0 +1,119 @@
...
+    }
+    return retval;
+}
+
+int main() {
+    stdio_init_all();
+
+    // If you (optionally) define a POSIX TZ here then the example will display local time instead of UTC.
+    // For the format see https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_431.html
+    setenv("TZ", "BST0GMT,M3.5.0/1,M10.5.0/2", 1);  // <-- this is the timezone spec for Europe/London
```
lurch: I wonder if it might be useful to hoist this up to the `CMakeLists.txt`, rather than having it hardcoded in the `.c` file? And that way you could also conditionally call `pico_localtime_r` as necessary based on whether the timezone is defined or not?

pico_w/wifi/ntp_system_time/README.md
mjcross: Thanks - wasn't sure which you'd prefer :-)

pico_w/wifi/ntp_system_time/ntp_system_time.c
mjcross: Well, in fact `ctime()` respects the TZ if one has been defined; so it would display local time

lurch: Ah, perhaps the comments need clarifying then :slightly_smiling_face:

pico_w/wifi/ntp_system_time/README.md
lurch: For written-text (rather than in source code), https://www.raspberrypi.com/products/ generally has the definitive product-names and capitalisation. See also https://github.com/raspberrypi/style-guide if you're _really_ curious :wink:

pico_w/wifi/ntp_system_time/ntp_system_time.c
mjcross: > hoist [the TZ def] up to the CMakeLists.txt, rather than having it hardcoded in the .c file?

Personally in the context of an example I think it's better for it to be defined in the source, close where it's being used (in fact I will move it closer). The `CMakeLists.txt` seems like black magic for a lot of people and they probably wouldn't look in there.

lurch: Fair enough!

mjcross: I don't think you'd want to make the call to `pico_localtime_r` conditional: it's needed to convert linear time to broken-down time (year/month/day/hour/ etc) whether a TZ is defined or not :-)

pico_w/wifi/ntp_system_time/ntp_system_time.c
mjcross: Good idea - will do

pico_w/wifi/ntp_system_time/ntp_system_time.c
lurch: Apologies for my misunderstanding - I refer the honourable gentleman to my earlier request for clearer comments :wink:

mjcross: Really useful to have your feedback: it's surprisingly hard to explain (and widely misunderstood)
Please can you have a look at [815159f](https://github.com/raspberrypi/pico-examples/pull/716/commits/815159f02346d31a9a1a5a626503ab14e97d53c0) and see if you think that's any better? Thx

pico_w/wifi/ntp_system_time/ntp_system_time.c
mjcross: Hopefully improved by [815159f](https://github.com/raspberrypi/pico-examples/pull/716/commits/815159f02346d31a9a1a5a626503ab14e97d53c0)

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -0,0 +1,124 @@
...
+
+    while (true) {
+
+        if(aon_timer_is_initialised) {
+
+            // read the current time as UTC seconds and ms since the epoch
+            get_time_utc(&ts);
+
+            // if you just want a string representation of the current time and you're not interested
+            // in the individual date/time fields, you could simply call ctime(&(ts.tv_sec)) here
```
lurch: Thank you, the comments are *much* clearer now. However I think it might be nice to show a commented-out example of using `ctime` (e.g. do I need to store/parse its return value? Do I just pass its return value straight into `printf`?) along with an example of the output-string it would create.

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -0,0 +1,124 @@
...
+            // (if you have set 'TZ' then the result will be in local time, otherwise UTC)
+
+            // unpack the individual date/time fields (if you have set 'TZ' then the values will be
+            // in local time, otherwise UTC)
+            pico_localtime_r(&(ts.tv_sec), &tm);
+
+            // display individual date/time fields in human readable form
+            printf("%s: %s", tm.tm_isdst ? "BST": "GMT", asctime(&tm));
```
lurch: ...along with an example of the output-string this would generate, so that you can see how it's different.

lurch:
Hmmm, if #716 and #714 got together and had a baby... :thinking:  :joy:

pico_w/wifi/ntp_system_time/ntp_system_time.c
mjcross: Done

pico_w/wifi/ntp_system_time/ntp_system_time.c
mjcross: done in [8af2a1f](https://github.com/raspberrypi/pico-examples/pull/716/commits/8af2a1f4783f9d8eaf5a7cbf4dc6443cf4d31395)

mjcross:
> Hmmm, if #716 and #714 got together and had a baby... 🤔 😂

https://digitalcollections.smu.edu/digital/collection/tir/id/205/

lurch:
Thanks for the extra comments :+1:  
Not having done any time-based processing in C myself, I was misreading `asctime()` as `ascii_time()` whereas I guess it's actually meant to be read as `as_ctime()` :joy:

mjcross:
> I was misreading `asctime()` as `ascii_time()` whereas I guess it's actually meant to be read as `as_ctime()` 😂

Oooo never considered that... My "lengthy and extensive research" (ha ha) suggests it might be the former: https://retrocomputing.stackexchange.com/questions/25601/what-is-the-meaning-of-asctime
In fact for anything more complex than a timestamp it's probably better to use `strftime()` but that's a bit OTT for an embedded app :-)

mjcross:
Found a way to avoid hardcoding the timezone names in the printf() :-)

pico_w/wifi/ntp_system_time/README.md
```diff
@@ -0,0 +1,92 @@
...
+ To reduce the logging level change the `SNTP_DEBUG` option in **lwipopts.h** to `LWIP_DBG_OFF` and/or remove the informational messages from `sntp_set_system_time_us()` in **ntp_system_time.c**.
+
+
+# Further details
+
+The example uses:
+
+1. the [SNTP application](https://www.nongnu.org/lwip/2_0_x/group__sntp.html) provided by the lwIP network stack
+2. the Pico SDK high level "always on timer" abstraction: [pico_aon_timer](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_pico_aon_timer)
+3. a [POSIX timezone](https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_431.html) to convert UTC to local time
```
lurch:
```suggestion
3. an optional [POSIX timezone](https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_431.html) to convert UTC to local time
```
?

pico_w/wifi/ntp_system_time/lwipopts.h
```diff
@@ -0,0 +1,45 @@
...
+// configure SNTP to use our callback to read the system time
+#define SNTP_GET_SYSTEM_TIME(sec, us)  sntp_get_system_time_us(&(sec), &(us))
+
+#define SNTP_RETRY_TIMEOUT           SNTP_RECV_TIMEOUT
+#define SNTP_RETRY_TIMEOUT_MAX       (SNTP_RETRY_TIMEOUT * 10)
+#define	SNTP_RETRY_TIMEOUT_EXP       1
+#define SNTP_MONITOR_SERVER_REACHABILITY    1
+
+// configure SNTP to use our callback to set the system time
+#define SNTP_SET_SYSTEM_TIME_US(sec, us)   sntp_set_system_time_us(sec, us)
```
lurch, line 43: _Perhaps_ it makes sense to move the `#define SNTP_GET_SYSTEM_TIME` line to be next to this one?

pico_w/wifi/ntp_system_time/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+### lwIP SNTP
+
+The lwIP SNTP app provides a straightforward way to obtain and process timestamps from a pool of NTP servers without the complexity of a full NTP implementation. The lwIP documentation covers the [configuration options](https://www.nongnu.org/lwip/2_0_x/group__sntp.html) but the comments in the [source code](https://github.com/lwip-tcpip/lwip/blob/master/src/apps/sntp/sntp.c) are also very helpful.
+
+SNTP uses the **macros** `SNTP_GET_SYSTEM_TIME(sec, us)` and `SNTP_SET_SYSTEM_TIME(sec, us)` to call user-provided functions for accessing the system clock. The example defines the macros in `lwipopts.h` and the callbacks themselves are near the top of `ntp_system_time.c`.
```
lurch: Your `lwipopts.h` file declares `SNTP_GET_SYSTEM_TIME` and `SNTP_SET_SYSTEM_TIME_US` which don't quite match the macro-names you mention here?

mjcross: Good point! 
In fact you can set the clock with either, but the non-US function only uses 1sec resolution.
For reasons best known to themselves the lwIP/SNTP authors didn't follow the same naming convention for the 'get' version...

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -0,0 +1,139 @@
...
+    sntp_init();
+
+
+    // ----- simple demonstration of how to read and display the system time -----
+    //
+    struct timespec ts;
+    struct tm tm;
+
+    // OPTIONAL: set the 'TZ' env variable to the local POSIX timezone (in this case Europe/London,
+    // to create your own see https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_431.html)
```
lurch:
```suggestion
    // OPTIONAL: set the 'TZ' env variable to the local POSIX timezone
    // this example is for Europe/London, to create your own see 
    // https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_431.html
```
(IMHO the brackets being split over multiple
lines was a bit confusing)

Additionally, maybe it's worth pointing people at https://github.com/nayarsystems/posix_tz_db/blob/master/zones.csv ?

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -0,0 +1,139 @@
...
+
+            // display the name of the currently active local timeszone, if defined
+            if (getenv("TZ")) {
+                printf("%s: ", tm.tm_isdst ? tzname[0]: tzname[1]);
+                // <time.h> defines `extern char *tzname[2]` to hold the names of the POSIX timezones
+            } else {
+                printf("UTC: ");
+            }
+            
+            // display individual date/time fields in the form "Mon Oct 27 22:06:08 2025\n"
```
lurch:
```suggestion
            // formats the date/time fields like: "Mon Oct 27 22:06:08 2025\n"
```
?

pico_w/wifi/ntp_system_time/lwipopts.h
mjcross: I think that's a good shout: it makes sense to co-locate them

pico_w/wifi/ntp_system_time/ntp_system_time.c
mjcross: Oh excellent! I was looking for a resource like that but couldn't find one...
I'll mention it in a comment and the README

mjcross:
Many thanks for the excellent suggestions - should be all wrapped up in the latest commit (please let me know if I missed anything)

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -108,15 +111,14 @@ int main() {
 
             // if you just want a string representation of the current time and you're not interested
             // in the individual date/time fields, then here you can simply call:
```
lurch: There's now a dangling colon at the end of this line.

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -108,15 +111,14 @@ int main() {
...
-            // then the output will be in local time, otherwise UTC.
-
-            // you can unpack the raw UTC seconds count into individual date/time fields like this.
-            // Again, if you set 'TZ' then the values will be in local time, otherwise UTC. Note
-            // that by default `pico_localtime_r()` just calls `localtime_r()` from the standard 'C' 
-            // library, but the declaration is 'weak' so that a user can override it with their own
-            // implementation if desired.
+
+            // if you don't need the date/time fields, you can call `ctime()` or one of its variants
+            // here to convert the raw timer value into a string like "Mon Oct 27 22:06:08 2025\n".
```
lurch:
```suggestion
            // here to convert the raw time value into a string like "Mon Oct 27 22:06:08 2025\n". 
```

pico_w/wifi/ntp_system_time/ntp_system_time.c
```diff
@@ -108,15 +111,14 @@ int main() {
...
-            // that by default `pico_localtime_r()` just calls `localtime_r()` from the standard 'C' 
-            // library, but the declaration is 'weak' so that a user can override it with their own
-            // implementation if desired.
+
+            // if you don't need the date/time fields, you can call `ctime()` or one of its variants
+            // here to convert the raw timer value into a string like "Mon Oct 27 22:06:08 2025\n". 
+            // If you have defined a valid 'TZ' the string will be in local time, otherwise UTC.
+            //printf("%s", ctime(&(ts.tv_sec)));
+
+            // you can extract the date/time fields use `localtime()` or one of its variants. If you
```
lurch:
```suggestion
            // you can extract the date/time fields using `localtime()` or one of its variants. If you
```

mjcross:
Looking really good now - many thanks again for your time to review

mjcross:
Spent far too long today chasing down why some simple test code couldn't find the function definition for `setenv()`. Eventually realised it was because in the example I'd forgotten to `#include <stdlib.h>`, and got away with it because lwIP pulls it in.
Hopefully by importing it explicitly it'll save someone else from the same fate...

mjcross:
I discovered this was giving very incorrect times of day when run on RP2040. This is because the aon_timer on that platform has a resolution of only one second, which is incompatible with SNTP round-trip compensation. To fix this I have added a test in the example's `CMakeLists.txt` to disable SNTP_COMP_ROUNDTRIP in `lwipopts.h` for RP2040.
If you know of a cleaner/better way of handling it I'd be happy to oblige :-)

lurch:
Looking at the RP2040 datasheet it seems like you could set up the RTC peripheral to trigger an interrupt once a minute (e.g. when the "seconds" value rolls around to 0), and then you could use the callback for that interrupt to read the current Timer peripheral value ("The system timer peripheral on RP2040 provides a global microsecond timebase for the system"). And then when you need the time with a smaller granularity than a second, you could re-read the timer again, subtract the value you stored in the callback, and that would give you the number of microseconds since the start of the last minute? (I guess it's very unlikely that an SNTP round-trip would take longer than a minute!)

Of course it's entirely up to you to decide if it's worth adding that level of complexity to your example :wink:

mjcross:
> Looking at the RP2040 datasheet it seems like you could set up the RTC peripheral to trigger an interrupt once a minute 
[...] 

very inventive!

> Of course it's entirely up to you to decide if it's worth adding that level of complexity to your example 😉

To be honest if anyone needs round-trip compensation that much it'd be more cost-effective to just send them a free Pico2_W!

mjcross:
@lurch I was thinking more of the proposed method of adjusting the lwIP options depending on the platform (i.e. the condition in the CMakeLists.txt): I was wondering whether there was a more preferred way of doing that :-)

lurch:
> @lurch I was thinking more of the proposed method of adjusting the lwIP options depending on the platform (i.e. the condition in the CMakeLists.txt): I was wondering whether there was a more preferred way of doing that :-)

That's more of a question for @peterharperuk - he's much more familiar with lwIP than I am!

mjcross:
Thanks @lurch 👍
Having had a look at §2.6 of [the SDK pdf](https://pip-assets.raspberrypi.com/categories/609-microcontroller-boards/documents/RP-009085-KB-1-raspberry-pi-pico-c-sdk.pdf) I'm more comfortable that the conditional `target_compiler_definitions()` in CMake is a fairly normal thing to do.

peterharperuk (approved):

PR #719, opened 2025-11-13
fix potential buffer overflows

mjcross:
Replace two unsafe uses of strncpy() in the MQTT client example, which would return unterminated strings if an over-long topic string was truncated


peterharperuk:
@kilograham Do you think it's safe for us to use strlcpy - I've not seen us use it much?

kilograham:
yes you can use strncpy

kilograham:
ah you said `strlcpy` it seems fine on all c-libs ive tried, so i will merge for now

kilograham (approved):

PR #721, opened 2025-11-22
Add simple async_context example

mjcross:
A simple example showing how to add an _at-time_ worker to a background async context.

The intention is to supplement the High Level API documentation in the SDK which doesn't really seem to go into the practical details of how to create and add workers.

The example builds and runs for both Pico and Pico-W boards so I suggest it belongs as a top level example rather than in the Pico-W folder. I've put it in an async_context folder on the assumption there might in due course be other examples (e.g. _when-pending_ worker).


mjcross:
Any reviewer comments?

peterharperuk:
It should use the status led library. I will get around to pushing this before the next sdk release. I'm afraid not much will happen till then.

mjcross:
Thanks @peterharperuk - that's a useful library I'd overlooked!
Your efforts are as always very much appreciated

peterharperuk (approved):

will-v-pi:
Why does this PR delete the sample generated PIO files, and add them and the `.gitignore` to the `.gitignore`? I think those changes from this PR should be reverted, as those changes are unrelated to the new example.

peterharperuk:
> Why does this PR delete the sample generated PIO files, and add them and the `.gitignore` to the `.gitignore`? I think those changes from this PR should be reverted, as those changes are unrelated to the new example.

Sorry - didn't spot that - will revert

Issue #725, opened 2025-12-15
PIO SPI Example not working for GPIO>=32


viralpatel9:
I am trying to run PIO SPI example using GPIOs more than 31, but I am not able to initialise them correctly. Things I have configured in my SDK:
- I have changed the `#define PICO_RP2350A 0`
- I have added gpio base to 16 for my `pio1` using  `pio_set_gpio_base(pio1, 16);`

I am using the latest SDK v2.2.0

peterharperuk:
What gpios are you using? Are you using a board based on rp2350? What do you mean by "not able to initialise them correctly"?

viralpatel9:
Yes, I am using RP2350B variant and I am trying to configure my SPI using below GPIOs    
``` 
{43, 46, 45},  // SCK, MOSI, MISO
```
I am toggling the CS pin (`GPIO44`) using generic GPIO APIs, which seems to work ok, but I can't see any behaviour on any of the other GPIOs (on scope).

I am using https://github.com/raspberrypi/pico-examples/blob/master/pio/spi/spi.pio example where my `cpol=0` and `cpha=1`

peterharperuk:
> Yes, I am using RP2350B

Make sure you are using a PICO_BOARD that has PICO_PIO_VERSION > 1
It doesn't look like pio_spi_init will work for >32 so that will need to be fixed.

viralpatel9:
I believe this is not the SDK limitations, and I can try and fix the `pio_spi_init` function to work with RP2350B variant?

peterharperuk:
Yes, its the example that needs to be fixed.

viralpatel9:
Hi, I have created a minimal code here https://github.com/viralpatel9/test-pio?tab=readme-ov-file#issue-with-pins32-on-rp2350b to report this issue. I have tried different combinations (of GPIOs) with my SPI peripheral (INA229) that I have been using. Unfortunately, I haven't got it working. Would you mind reviewing the code and seeing if I have missed anything? Thanks.

peterharperuk:
Your code looks ok. Are you sure you have the pins connected correctly?
The only issue is with the line...

```
hw_set_bits(&pio->input_sync_bypass, 1u << pin_miso);
```

which should be changed to something like...

```
hw_set_bits(&pio->input_sync_bypass, (pin_miso >= 32) ? (1u << (pin_miso - 16)) : (1u << pin_miso));
```

But I wouldn't expect that to be an issue here.
[...]

PR #727, opened 2025-12-17
Fix: Update SPI PIO initialization to use 64-bit masks for RP2350B

viralpatel9:
This PR is a fix for Issue #725 to allow using the GPIOs>32 (typically for RP2350B variant). This pull request updates the SPI PIO initialisation functions to improve support for higher-numbered GPIO pins and 64-bit pin masks. 

The main changes involve switching to 64-bit versions of pin mask functions and correcting the logic for bypassing input synchronisation on the MISO pin.

### DIFF
--- a/pio/spi/spi.pio
+++ b/pio/spi/spi.pio
@@ -55,8 +55,8 @@ static inline void pio_spi_init(PIO pio, uint sm, uint prog_offs, uint n_bits,
     sm_config_set_clkdiv(&c, clkdiv);
 
     // MOSI, SCK output are low, MISO is input
-    pio_sm_set_pins_with_mask(pio, sm, 0, (1u << pin_sck) | (1u << pin_mosi));
-    pio_sm_set_pindirs_with_mask(pio, sm, (1u << pin_sck) | (1u << pin_mosi), (1u << pin_sck) | (1u << pin_mosi) | (1u << pin_miso));
+    pio_sm_set_pins_with_mask64(pio, sm, 0, (1ull << pin_sck) | (1ull << pin_mosi));
+    pio_sm_set_pindirs_with_mask64(pio, sm, (1ull << pin_sck) | (1ull << pin_mosi), (1ull << pin_sck) | (1ull << pin_mosi) | (1ull << pin_miso));
     pio_gpio_init(pio, pin_mosi);
     pio_gpio_init(pio, pin_miso);
     pio_gpio_init(pio, pin_sck);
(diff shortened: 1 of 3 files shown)


peterharperuk (approved):

pio/spi/spi.pio
```diff
@@ -65,7 +65,7 @@ static inline void pio_spi_init(PIO pio, uint sm, uint prog_offs, uint n_bits,
     // and this is a cheesy way to get CPOL=1
     gpio_set_outover(pin_sck, cpol ? GPIO_OVERRIDE_INVERT : GPIO_OVERRIDE_NORMAL);
     // SPI is synchronous, so bypass input synchroniser to reduce input delay.
-    hw_set_bits(&pio->input_sync_bypass, 1u << pin_miso);
+    hw_set_bits(&pio->input_sync_bypass, (pin_miso >= 32) ? (1u << (pin_miso - 16)) : (1u << pin_miso));
```
lurch, line 68: I'm not _entirely_ sure that this is correct. You'll need to do the `(pin_miso - 16)` thing if the PIO has been remapped to the "upper pin range", which will indeed happen if `(pin_miso >= 32)`; but there's a slight edge-case here where `pin_miso` might be less than 32, but one of the _other_ pins (e.g. `pin_sck`) might be above 32? (which would also cause the PIO to use the "upper pin range")
So I _think_ the correct fix here is actually:
```c
    hw_set_bits(&pio->input_sync_bypass, (1u << (pin_miso - pio_get_gpio_base(pio))));
```
?
(but I've also suggested https://github.com/raspberrypi/pico-sdk/issues/2986 to make this easier in future)

pio/spi/spi.pio
```diff
@@ -150,14 +150,14 @@ static inline void pio_spi_cs_init(PIO pio, uint sm, uint prog_offs, uint n_bits
...
-    pio_sm_set_pindirs_with_mask(pio, sm, (3u << pin_sck) | (1u << pin_mosi), (3u << pin_sck) | (1u << pin_mosi) | (1u << pin_miso));
+    pio_sm_set_pins_with_mask64(pio, sm, (2ull << pin_sck), (3ull << pin_sck) | (1ull << pin_mosi));
+    pio_sm_set_pindirs_with_mask64(pio, sm, (3ull << pin_sck) | (1ull << pin_mosi), (3ull << pin_sck) | (1ull << pin_mosi) | (1ull << pin_miso));
     pio_gpio_init(pio, pin_mosi);
     pio_gpio_init(pio, pin_miso);
     pio_gpio_init(pio, pin_sck);
     pio_gpio_init(pio, pin_sck + 1);
     gpio_set_outover(pin_sck, cpol ? GPIO_OVERRIDE_INVERT : GPIO_OVERRIDE_NORMAL);
-    hw_set_bits(&pio->input_sync_bypass, 1u << pin_miso);
+    hw_set_bits(&pio->input_sync_bypass, (pin_miso >= 32) ? (1u << (pin_miso - 16)) : (1u << pin_miso));
```
lurch, line 160: ...and obviously the same applies here too :slightly_smiling_face:

pio/spi/spi.pio
viralpatel9: Hi @lurch that seems to be correct rather than using magic number.

PR #731, opened 2025-12-29
Add example spi/ssd1309_spi

mjcross:
Add an example of how to attach one of the widely available small OLED displays based on the SSD1309 controller via SPI, complementing the existing SSD1306 example that uses i2c. In fact the two controllers have the same command set so the example should work on either (but I don't have an SSD1306 to test).
For many applications the SPI interface might be a better bet, as it runs a lot faster (10Mbit/sec).


mjcross:
Incidentally I've also created a similar version that uses a frame buffer updated via DMA direct to the SPI peripheral; but I'll submit that as a separate PR. This version is much easier to understand.

spi/ssd1309_spi/ssd1309_spi.c
```diff
@@ -0,0 +1,144 @@
...
+    uint col = text_col * 8;
+    send_command(col & 0x0f);                   // set the column start address (low nibble)
+    send_command(0x10 + ((col & 0xf0) >> 4));   // set the column start address (high nibble)
+}
+
+// display a single text character at the cursor position
+void display_char(char c) {
+    uint font_index = 0;             // default (space)
+    c = toupper(c);
+    if (isalpha(c)) {
```
lurch, line 70: Minor nitpick: I guess you could put the `c = toupper(c);` inside `if (isalpha(c)) {` :wink:

spi/ssd1309_spi/ssd1309_spi.c
```diff
@@ -0,0 +1,144 @@
...
+        font_index = c - 'A' + 1;    // a-z and A-Z
+    } else if (isdigit(c)) {
+        font_index = c - '0' + 27;   // 0-9
+    }
+    gpio_put(PIN_DC, DC_DATA_MODE);
+    spi_write_blocking(SPI_DEVICE, font + font_index * 8, 8);
```
lurch: It's been a while since I've written low-level C, but I _think_ this is the same as:
```c
spi_write_blocking(SPI_DEVICE, &font[font_index * 8], 8);
```
?
IMHO the array-access notation might be slightly easier to read than direct pointer arithmetic?

spi/ssd1309_spi/ssd1309_spi.c
```diff
@@ -0,0 +1,144 @@
...
+    spi_write_blocking(SPI_DEVICE, &cmd_byte, 1);
+}
+
+// send a data byte to the display
+void send_data(uint8_t data_byte) {
+    gpio_put(PIN_DC, DC_DATA_MODE);
+    spi_write_blocking(SPI_DEVICE, &data_byte, 1);
+}
+
+// set the text cursor position (row, col)
```
lurch, line 55: Is it worth clarifying what the maximum text row and column values are?

spi/ssd1309_spi/ssd1309_spi.c
```diff
@@ -0,0 +1,144 @@
...
+    if (isalpha(c)) {
+        font_index = c - 'A' + 1;    // a-z and A-Z
+    } else if (isdigit(c)) {
+        font_index = c - '0' + 27;   // 0-9
+    }
+    gpio_put(PIN_DC, DC_DATA_MODE);
+    spi_write_blocking(SPI_DEVICE, font + font_index * 8, 8);
+}
+
+// display a null-terminated string at the cursor position
```
lurch, line 79: What happens if the string is too long to fit on the display?

spi/ssd1309_spi/ssd1309_spi.c
```diff
@@ -0,0 +1,144 @@
...
+
+    // after a short pause, send the 'display on' command
+    sleep_ms(1);
+    send_command(0xaf);
+    display_clear();
+}
+
+
+int main() {
+    stdio_init_all();
```
lurch, line 126: This example doesn't seem to actually use stdio?

spi/ssd1309_spi/ssd1309_spi.c
lurch: Also, perhaps it's worth adding a `send_multiple_data` function which accepts a pointer and a length, and then you could use that here so that `display_char` is "isolated" from the lower-level GPIO manipulation? :thinking: 
(and maybe you could use that to make your `display_clear` function more efficient too)

spi/ssd1309_spi/README.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+CMakeLists.txt:: A file to configure the CMake build system for the example.
+ssd1309_spi.c:: The example code.
+ssd1309_font.h:: A basic 8x8 bit font table used by the example.
+
+
+== Bill of Materials
+
+.A list of materials required for the example
+[[BME280-bom-table]]
```
lurch: `SSD1309-bom-table` ? :wink:

spi/ssd1309_spi/ssd1309_spi.c
mjcross: With respect, I think it's clearer the way it is - especially as `c` is used elsewhere in the function :-)

spi/ssd1309_spi/ssd1309_spi.c
mjcross: Yes the two forms are entirely equivalent.
I think you're right that the `&(array[index])` form is probably better here though: I'll change it - thanks

mjcross: Ideally `font[]` should be an array of 8-byte elements, which would get rid of the 'magic' `index * 8`... but I just lifted it from the I2C example ;-)

spi/ssd1309_spi/ssd1309_spi.c
mjcross: Nothing bad: it just wraps around.
To be precise the behaviour depends on what addressing mode you put the display into, but in the default mode it just wraps on the same row.
I'll add a comment to the code to point that out and refer to the mfrs datasheet.

spi/ssd1309_spi/ssd1309_spi.c
mjcross: Correct - but I thought it might be wise to enable it in case of runtime error messages... What do you think?

spi/ssd1309_spi/ssd1309_spi.c
lurch: Fair enough.

spi/ssd1309_spi/ssd1309_spi.c
lurch: I'll leave this one up to @peterharperuk :slightly_smiling_face:

spi/ssd1309_spi/ssd1309_spi.c
mjcross: If you really want, but I'm not sure it's worth it. I initially did it that way myself for the same reasons, but as the `send multi()` function was only being used in `display_char()` I took it out. I agree that 'isolation' would be the way to go for maintainable production code, but it seemed to me that clarity was the most important thing for an example and this was the clearest way.

On the efficiency of `display_clear()` I agree it's rudimentary and there are more efficient ways. But as we are _firmly_ IO-bound here I decided, again, to err on the side of simplicity and keep the swanky stuff for a separate example that uses a DMA frame buffer (far better for many reasons, but a lot harder to follow for a basic use-case).

Happy to go either way, but what do you think?

spi/ssd1309_spi/README.adoc
mjcross: Heh - good catch! Now you know where I 'lifted' the .adoc template from haha

spi/ssd1309_spi/ssd1309_spi.c
mjcross: fixed in https://github.com/raspberrypi/pico-examples/pull/731/commits/622210dc2773cd64dfdb7911f4b0633e6fb69551

spi/ssd1309_spi/ssd1309_spi.c
mjcross: fixed in https://github.com/raspberrypi/pico-examples/pull/731/commits/622210dc2773cd64dfdb7911f4b0633e6fb69551

spi/ssd1309_spi/ssd1309_spi.c
mjcross: fixed in https://github.com/raspberrypi/pico-examples/pull/731/commits/622210dc2773cd64dfdb7911f4b0633e6fb69551

spi/ssd1309_spi/README.adoc
mjcross: fixed in https://github.com/raspberrypi/pico-examples/pull/731/commits/308de3eabf89a07f5ebed168ff991e7de071527c

spi/ssd1309_spi/ssd1309_spi.c
lurch: These were just minor suggestions; I'm happy for you to leave the code as-is.

spi/ssd1309_spi/ssd1309_font.h
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// Vertical bitmaps, A-Z, 0-9. Each is 8 pixels high and wide
+// These are defined vertically to make them quick to copy to FB
+
+static uint8_t font[] = {
+0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // Nothing
+0x78, 0x14, 0x12, 0x11, 0x12, 0x14, 0x78, 0x00, //A
```
lurch, line 19: Probably not worth changing, but I guess you _could_ add `#define FONT_INDEX_SPACE 0`, `#define FONT_INDEX_A 1` and `#define FONT_INDEX_0 27` here, and then you could get rid of the "magic numbers" in `display_char` :wink:

lurch:
Thank you for the high-quality example :+1:  
Sorry if I went a bit overboard on the number of comments.

spi/ssd1309_spi/ssd1309_font.h
mjcross: That's a really good idea

mjcross:
> Thank you for the high-quality example 👍 Sorry if I went a bit overboard on the number of comments.

You're more than welcome! I wanted to use one of these displays in a current project, and writing it up as an example encourages me to do a better job. Your comments are really helpful and always on point :-)

I'll be interested to see what you think of the DMA frame-buffer version, which should be ready in the next day or two. First time I've dipped my toe into using DMA and it's a really well designed API 👍

spi/ssd1309_spi/ssd1309_font.h
mjcross: Fixed in https://github.com/raspberrypi/pico-examples/pull/731/commits/7267b95f76259d9ba873186a39352fc0c883d901

spi/ssd1309_spi_dma/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+add_executable(ssd1309_spi_dma
```
lurch: Oops, I assume you accidentally added this file to the wrong PR?

peterharperuk (dismissed):

peterharperuk (approved):

PR #732, opened 2026-01-04
Add example to display stdout text and simple graphics on an SSD1309-based OLED panel via SPI.

mjcross:
Small (128x64) monochrome OLED displays based on the SSD1309 are widely available at low cost such as [https://www.aliexpress.com/item/1005003171884489.html](https://www.aliexpress.com/item/1005003171884489.html). They are an attractive option for small embedded projects.

This PR adds an SPI-based example for the SSD1309 that complement the existing I2C-based one for the SSD1306, and adds significant functionality such as a DMA based frame buffer and a simple stdout driver.


peterharperuk (commented):
Just works - I will push a rebase

peterharperuk (approved):

PR #737, opened 2026-02-14
Various tweaks to pio_blink example

lurch:
* Add extra comments and defines to make the code easier to understand
* Use pio_claim_unused_sm instead of hardcoding "last SM + 1"
* Should now work with LED3 on GPIO 31 and LED4 on GPIO 32
* Stops flashing after 60 seconds to demonstrate PIO cleanup
* Add binary_info pin information

### DIFF
--- a/pio/pio_blink/blink.c
+++ b/pio/pio_blink/blink.c
@@ -13,76 +13,92 @@
 
 void blink_pin_forever(PIO pio, uint sm, uint offset, uint pin, uint freq);
 
-// by default flash leds on gpios 3-4
+// By default flash LEDs on GPIOs 3 and 4
 #ifndef PIO_BLINK_LED1_GPIO
 #define PIO_BLINK_LED1_GPIO 3
+#define PIO_BLINK_LED2_GPIO (PIO_BLINK_LED1_GPIO + 1)
 #endif
 
-// and flash leds on gpios 5-6
-// or if the device supports more than 32 gpios, flash leds on 32-33
+// and also flash LEDs on GPIOs 5 and 6
+// Or if the device has more than 32 gpios, also flash LEDs on GPIOs 32 and 33
 #ifndef PIO_BLINK_LED3_GPIO
 #if NUM_BANK0_GPIOS <= 32
 #define PIO_BLINK_LED3_GPIO 5
 #else
 #define PIO_BLINK_LED3_GPIO 32
 #endif
+#define PIO_BLINK_LED4_GPIO (PIO_BLINK_LED3_GPIO + 1)
 #endif
+
+#define PIO_BLINK_LED1_FREQUENCY 4
+#define PIO_BLINK_LED2_FREQUENCY 3
+#define PIO_BLINK_LED3_FREQUENCY 2
+#define PIO_BLINK_LED4_FREQUENCY 1
 
 int main() {
     setup_default_uart();
 
-    assert(PIO_BLINK_LED1_GPIO < 31);
-    assert(PIO_BLINK_LED3_GPIO < 31 || PIO_BLINK_LED3_GPIO >= 32);
+    // LED1 and LED2 are both expected to be in the "lower" range of PIO-addressable GPIOs
+    assert((PIO_BLINK_LED1_GPIO < 32) && (PIO_BLINK_LED2_GPIO < 32));
+    // check LED3 and LED4 are both in the same range of PIO-addressable GPIOs
+    assert(((PIO_BLINK_LED3_GPIO < 32) && (PIO_BLINK_LED4_GPIO < 32)) || ((PIO_BLINK_LED3_GPIO >= 16) && (PIO_BLINK_LED3_GPIO < 48) && (PIO_BLINK_LED4_GPIO >= 16) && (PIO_BLINK_LED4_GPIO < 48)));
 
+    // LED1 and LED2 are both controlled by the program loaded into pio[0] at offset[0]
+    // LED3 and LED4 are both controlled by the program loaded into pio[1] at offset[1] (which might be the same as pio[0] and offset[0] if LED3 and LED4 are both on GPIOs < 32)
+    // LED1 is controlled by sm[0]
+    // LED2 is controlled by sm[1]
+    // LED3 is controlled by sm[2]
+    // LED4 is controlled by sm[3]
     PIO pio[2];
-    uint sm[2];
+    uint sm[4];
     uint offset[2];
 
-    // Find a free pio and state machine and add the program
+    // Find a free PIO and state machine and add the program
     bool rc = pio_claim_free_sm_and_add_program_for_gpio_range(&blink_program, &pio[0], &sm[0], &offset[0], PIO_BLINK_LED1_GPIO, 2, true);
     hard_assert(rc);
     printf("Loaded program at %u on pio %u\n", offset[0], PIO_NUM(pio[0]));
 
-    // Start led1 flashing
-    blink_pin_forever(pio[0], sm[0], offset[0], PIO_BLINK_LED1_GPIO, 4);
+    // Start LED1 flashing
+    blink_pin_forever(pio[0], sm[0], offset[0], PIO_BLINK_LED1_GPIO, PIO_BLINK_LED1_FREQUENCY);
 
-    // Claim the next state machine and start led2 flashing
-    pio_sm_claim(pio[0], sm[0] + 1);
-    blink_pin_forever(pio[0], sm[0] + 1, offset[0], PIO_BLINK_LED1_GPIO + 1, 3);
+    // Claim the next unused state machine and start LED2 flashing
+    sm[1] = pio_claim_unused_sm(pio[0], true);
+    blink_pin_forever(pio[0], sm[1], offset[0], PIO_BLINK_LED2_GPIO, PIO_BLINK_LED2_FREQUENCY);
 
-    if (PIO_BLINK_LED3_GPIO >= 32) {
-        // Find a free pio and state machine and add the program
-        rc = pio_claim_free_sm_and_add_program_for_gpio_range(&blink_program, &pio[1], &sm[1], &offset[1], PIO_BLINK_LED3_GPIO, 2, true);
+    if ((PIO_BLINK_LED3_GPIO >= 32) || (PIO_BLINK_LED4_GPIO >= 32)) {
+        // Find a free PIO and state machine and add the program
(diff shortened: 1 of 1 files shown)


pio/pio_blink/blink.c
```diff
@@ -13,76 +13,92 @@
...
+        rc = pio_claim_free_sm_and_add_program_for_gpio_range(&blink_program, &pio[1], &sm[2], &offset[1], PIO_BLINK_LED3_GPIO, 2, true);
         printf("Loaded program at %u on pio %u\n", offset[1], PIO_NUM(pio[1]));
     } else {
         // no need to load the program again
         rc = true;
         pio[1] = pio[0];
-        sm[1] = sm[0] + 2;
         offset[1] = offset[0];
-        pio_sm_claim(pio[1], sm[1]);
+        sm[2] = pio_claim_unused_sm(pio[1], true);
```
peterharperuk, line 107: Maybe we should have a pio_claim_free_sm_and_add_for_gpio_range function

lurch: Sounds like a @kilograham question to me! :laughing: 
(Although I guess that potential function might be useful for @poetaster in https://github.com/raspberrypi/pico-sdk/issues/2030 )

pio/pio_blink/blink.c
```diff
@@ -13,76 +13,92 @@
...
+#define PIO_BLINK_LED4_FREQUENCY 1
+
 int main() {
     setup_default_uart();
 
-    assert(PIO_BLINK_LED1_GPIO < 31);
-    assert(PIO_BLINK_LED3_GPIO < 31 || PIO_BLINK_LED3_GPIO >= 32);
-
+    // LED1 and LED2 are both expected to be in the "lower" range of PIO-addressable GPIOs
+    assert((PIO_BLINK_LED1_GPIO < 32) && (PIO_BLINK_LED2_GPIO < 32));
```
peterharperuk, line 72: arguably, these asserts should be unnecessary - the code should be able to work whatever value is used.

lurch: I guess it depends on the focus of the example - is it showing "here's how you flash two GPIOs in one PIO-range and two GPIOs in a different range" (which I think was the original intention?), or is it showing "here's totally generic (but more complicated) code for flashing any four GPIOs in any combination" ?
Perhaps I ought to add a comment highlighting the intention of the example? :thinking:

pio/pio_blink/blink.c
peterharperuk: Or we should make the existing function clever enough to know not to add the same program to the same PIO twice.

lurch: > Or we should make the existing function clever enough to know not to add the same program to the same PIO twice.

Don't know if we can do that with the current API, as I think the PIO instruction memory is write-only?

peterharperuk: You might be able to stash an array of pio_program_t structures. Worth seeing if it makes the API better.

pio/pio_blink/blink.c
lurch: > Perhaps I ought to add a comment highlighting the intention of the example?

Done.

kilograham: I certainly dont think we care about the +1, so remove those asserts

pio/pio_blink/blink.c
```diff
@@ -11,78 +11,103 @@
...
-    pio_sm_claim(pio[0], sm[0] + 1);
-    blink_pin_forever(pio[0], sm[0] + 1, offset[0], PIO_BLINK_LED1_GPIO + 1, 3);
+    // Claim the next unused state machine and start LED2 flashing
+    sm[1] = pio_claim_unused_sm(pio[0], true);
+    blink_pin_forever(pio[0], sm[1], offset[0], PIO_BLINK_LED2_GPIO, PIO_BLINK_LED2_FREQUENCY);
 
-    if (PIO_BLINK_LED3_GPIO >= 32) {
-        // Find a free pio and state machine and add the program
-        rc = pio_claim_free_sm_and_add_program_for_gpio_range(&blink_program, &pio[1], &sm[1], &offset[1], PIO_BLINK_LED3_GPIO, 2, true);
+    if ((PIO_BLINK_LED3_GPIO >= 32) || (PIO_BLINK_LED4_GPIO >= 32)) {
```
kilograham, line 98: arguably we already checked this by assert - perhaps use NUM_BANK0_GPIOS instead

pio/pio_blink/blink.c
```diff
@@ -11,78 +11,103 @@
...
-    // Claim the next state machine and start led4 flashing
-    pio_sm_claim(pio[1], sm[1] + 1);
-    blink_pin_forever(pio[1], sm[1] + 1, offset[1], PIO_BLINK_LED3_GPIO + 1, 1);
+    printf("All LEDs should be flashing\n");
 
-    // free up pio resources
-    pio_sm_unclaim(pio[1], sm[1] + 1);
-    if (PIO_BLINK_LED3_GPIO >= 32) {
-        pio_remove_program_and_unclaim_sm(&blink_program, pio[1], sm[1], offset[1]);
+    // free up PIO resources
```
kilograham: not sure why we are freeing stuff up if we are having it continue to blink (frankly i thought in debug builds that remove program erased the instruction memory, but maybe i'm mis-remembering)

pio/pio_blink/blink.c
kilograham: this discussion a bit out of place here since it isn't clear how it helps - feel free to open a discussion in SDK

kilograham (changes requested):

pio/pio_blink/blink.c
peterharperuk: I think the rationale is that it's example code which might be copied - so we should show how to cleanup properly? Perhaps we should allow the user to press "x" to exit

kilograham: yeah i mean the intent is either to say "hey look it goes on blinking without CPU" in which case we should not clean up as that is confusing, or as you say we should have a key to exit, but then clean up and probably stop the SM

pio/pio_blink/blink.c
lurch: Not really - all the asserts are checking is that `PIO_BLINK_LED3_GPIO` and `PIO_BLINK_LED4_GPIO` are either both `< 32` or both `(>= 16 && < 48)`.
(I.e. even on a RP2350B where NUM_BANK0_GPIOS > 32, you _might_ have LED3 and LED4 both < 32)

pio/pio_blink/blink.c
lurch: I've removed the cleanup code

lurch:
Rebased onto latest `develop` and force-pushed.

pio/pio_blink/blink.c
```diff
@@ -93,19 +93,10 @@ int main() {
...
-    // free up PIO resources
-    pio_sm_unclaim(pio[1], sm[3]);
-    if ((PIO_BLINK_LED3_GPIO >= 32) || (PIO_BLINK_LED4_GPIO >= 32)) {
-        pio_remove_program_and_unclaim_sm(&blink_program, pio[1], sm[2], offset[1]);
-    } else {
-        pio_sm_unclaim(pio[1], sm[2]);
-    }
-    pio_sm_unclaim(pio[0], sm[1]);
-    pio_remove_program_and_unclaim_sm(&blink_program, pio[0], sm[0], offset[0]);
+    sleep_ms(3000);
```
peterharperuk: I really would like the example to show how you cleanup. I think we should leave this in and just add a way of stopping the program rather than running forever.

lurch: I'm reluctant to add some kind of "user interaction" to this example, as I feel like it'd distract from the core purpose of the example. What about if I changed it to leave the LEDs flashing for e.g. 10 seconds, and then disabled the SMs and cleanly shut everything down?

peterharperuk: That might work. Maybe make it 60s

lurch:
@peterharperuk I _think_ my latest change is correct, and it compiles, but I've not tested it on real hardware.

pio/pio_blink/blink.c
```diff
@@ -11,78 +11,136 @@
...
 #if NUM_BANK0_GPIOS <= 32
 #define PIO_BLINK_LED3_GPIO 5
 #else
 #define PIO_BLINK_LED3_GPIO 32
 #endif
+#define PIO_BLINK_LED4_GPIO (PIO_BLINK_LED3_GPIO + 1)
+#endif
+
+#ifndef PIO_BLINK_LED1_FREQUENCY
+#define PIO_BLINK_LED1_FREQUENCY 4
```
peterharperuk, line 47: It would be slightly more obvious if the led frequency doubled each time rather than just increased by 1, i.e. make this 8

pio/pio_blink/blink.c
```diff
@@ -11,78 +11,136 @@
...
+#endif
+#ifndef PIO_BLINK_LED2_FREQUENCY
+#define PIO_BLINK_LED2_FREQUENCY 3
```
peterharperuk, line 50: and make this 4?

peterharperuk (dismissed):
I tested it and it works.

pio/pio_blink/blink.c
lurch: I didn't choose the original frequencies, I just moved them to being `#define`s :slightly_smiling_face: 
I guess one advantage to having frequencies of 4, 3, 2 and 1Hz, is that it's more visually obvious that they're "independent" frequencies (i.e. 3 and 4 are co-prime), rather than just all being driven from the same core-loop but being switched on and off at different loop-counter values?
E.g. frequencies of 8, 4, 2 and 1Hz could just be created with something like:
```c
while (true) {
    for (int i=0; i < 8; i++) {
        toggle_led(0); // 8Hz
        if (i % 2 == 0) toggle_led(1); // 4Hz
        if (i % 4 == 0) toggle_led(2); // 2Hz
        if (i == 0) toggle_led(3); // 1Hz
        sleep_ms(63); // one sixteenth of a second
    }
}
```

peterharperuk: I mean it's not obvious that all the frequencies are different from looking at the leds. But, submit as it is.

lurch: I'll have a look at your working-hardware-demo on Monday, and then merge this PR after that :slightly_smiling_face: (perhaps if we insert a small delay between starting each LED, rather than starting them all at the same time, that'll make it more obvious that they're operating at different cadences?)

peterharperuk (approved):

lurch:
I can't merge this yet, because GitHub thinks that Graham is still requesting changes?

kilograham (approved):

PR #743, opened 2026-03-11
Fix bt examples

peterharperuk:
Rework BT examples. Move them to a separate "bluetooth" folder.
Put the btstack examples in "bluetooth/btstack_examples"
Add new standalone BT examples


README.md
```diff
@@ -296,8 +297,11 @@ Some standalone Bluetooth examples (without all the common example build infrast
 App|Description
 ---|---
 [picow_ble_temp_sensor](pico_w/bt/standalone) | Reads from the on board temperature sensor and sends notifications via BLE.
-[picow_ble_temp_sensor_with_wifi](pico_w/bt/standalone) | Same as above but also connects to Wi-Fi and starts an "iperf" server.
-[picow_ble_temp_reader](pico_w/bt/standalone) | Connects to one of the above "sensors" and reads the temperature.
+[picow_ble_temp_reader](pico_w/bt/standalone) | Connects to the above sensor and reads the temperature.
+[picow_ble_pointer](pico_w/bt/standalone/ble_pointer) | Bluetooth HID mouse using mpu6050 to detect angle and move cursor.
+[picow_ble_doorbell](pico_w/bt/standalone/doorbell) | Detects button press on transmitter Pico and illuminates LED on reciever Pico.
```
lurch: I guess these links in the README will need to be updated to e.g. `bluetooth/doorbell` ?

peterharperuk: I've modified the readme

peterharperuk:
@kilograham Does this look ok to you. I'd like to get it submitted to develop.

1. Removes pico_w/bt
2. Adds bluetooth
3. All btstack examples are in bluetooth/btstack_examples and uses less mad cmake

More bt examples added to bluetooth folder...

1. ble_pointer
2. doorbell
3. temp_sensor (moved from pico_w/bt/standalone)
4. secure_temp_sensor
5. wifi_provisioner

bluetooth/README.md
```diff
@@ -6,11 +6,7 @@ There's a standalone example in `pico-examples/pico_w/bt/standalone`.
 ## Debugging
 
 To debug Bluetooth issues you can enable [btstack](https://github.com/bluekitchen/btstack) debug output which also enables packet logging.
-Define `WANT_HCI_DUMP=1` in your CMakeLists.txt file. Uncomment this line to enable debug in the btstack examples.
-
-	target_compile_definitions(picow_bt_example_common INTERFACE
-	    #WANT_HCI_DUMP=1 # This enables btstack debug
-	    )
+Define `WANT_HCI_DUMP=1` in your CMakeLists.txt file.
```
kilograham, line 9: is this true?>

kilograham (changes requested):
one comment - but also why is the addition of a wifi example in this PR, and does it belong under pico_w?

bluetooth/README.md
peterharperuk: Not any more. The readme should be fixed here.

peterharperuk:
> one comment - but also why is the addition of a wifi example in this PR, and does it belong under pico_w?

Brain fade (it was in the same branch that the BT examples came from). It should be in a separate PR.

bluetooth/ble_pointer/btstack_config.h
```diff
@@ -1,6 +1,6 @@
 #ifndef _PICO_BTSTACK_CONFIG_H
 #define _PICO_BTSTACK_CONFIG_H
 
-#include "btstack_config_common.h"
+#include "../config/btstack_config_common.h"
```
will-v-pi: This change will cause issues with VS Code - could the CMake includes be changed instead?

The CMake would be:
```cmake
target_include_directories(ble_pointer PRIVATE
    ${CMAKE_CURRENT_LIST_DIR}
    ${CMAKE_CURRENT_LIST_DIR}/../config
)
```

bluetooth/wifi_provisioner/lwipopts.h
```diff
@@ -0,0 +1 @@
+#include "../../pico_w/wifi/lwipopts_examples_common.h"
```
will-v-pi: Similarly this will cause problems with VS Code, so should be changed in the CMake instead

bluetooth/btstack_examples/a2dp_sink_demo/CMakeLists.txt
will-v-pi, line 1: This new way of doing things should make VS Code inclusion much easier for the bluetooth examples - thank you!

bluetooth/btstack_examples/a2dp_sink_demo/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+    pico_stdlib
+    pico_cyw43_arch_none # we don't need lwip support
+    pico_btstack_cyw43 # enable bt
+    pico_btstack_ble # ble support
+    pico_btstack_classic # classic support
+    pico_btstack_sbc_decoder
+    pico_audio_i2s
+)
+target_include_directories(a2dp_sink_demo PRIVATE
+    . # for btstack config
```
will-v-pi: Should use `${CMAKE_CURRENT_LIST_DIR}` instead of `.` (and as mentioned before, should also include `${CMAKE_CURRENT_LIST_DIR}/../config` so the `btstack_config.h` can just be `#include "btstack_config_common.h"`)

peterharperuk:
> > one comment - but also why is the addition of a wifi example in this PR, and does it belong under pico_w?
> 
> Brain fade (it was in the same branch that the BT examples came from). It should be in a separate PR.

Actually - this is a Bluetooth example. The wifi ssid and password are sent from another device via BLE. So I think it belongs here.

bluetooth/README.md
peterharperuk: Actually this is correct. All the cmake files have #WANT_HCI_DUMP=1 in them. You can remove the comment to get a packet log.

will-v-pi:
> > > one comment - but also why is the addition of a wifi example in this PR, and does it belong under pico_w?
> > 
> > 
> > Brain fade (it was in the same branch that the BT examples came from). It should be in a separate PR.
> 
> Actually - this is a Bluetooth example. The wifi ssid and password are sent from another device via BLE. So I think it belongs here.

That's the `wifi_provisioner` example - there's also the `AP_provisioning` example which is wifi only

peterharperuk:
> > > > one comment - but also why is the addition of a wifi example in this PR, and does it belong under pico_w?
> > > 
> > > 
> > > Brain fade (it was in the same branch that the BT examples came from). It should be in a separate PR.
> > 
> > 
> > Actually - this is a Bluetooth example. The wifi ssid and password are sent from another device via BLE. So I think it belongs here.
> 
> That's the `wifi_provisioner` example - there's also the `AP_provisioning` example which is wifi only

ok - I'll get rid of access_point_wifi_provisioning

peterharperuk:
I've squashed all the junk commits

bluetooth/wifi_provisioner/lwipopts.h
will-v-pi: This fix got missed - the CMake was changed correctly though

```suggestion
#include "lwipopts_examples_common.h"
```

bluetooth/ble_pointer/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+add_executable(ble_pointer 
+    ble_pointer.c
+)
+target_link_libraries(ble_pointer
+    pico_stdlib
+    pico_btstack_ble
+    pico_btstack_cyw43
+    pico_cyw43_arch_none
+    mpu6050_i2c_lib
+)
```
will-v-pi: Can this be changed to not use a separate library for `mpu6050_i2c_lib` - the library isn't used anywhere else, and triggers bugs in my VS Code parsing of the `mpu6050_i2c` example
```suggestion
add_executable(ble_pointer 
    ble_pointer.c
    mpu6050_i2c_lib.c
)
target_link_libraries(ble_pointer
    pico_stdlib
    pico_btstack_ble
    pico_btstack_cyw43
    pico_cyw43_arch_none
    hardware_i2c
)
```

bluetooth/doorbell/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+# Standalone example that connects to the server Pico and flashes an external LED when the button is pressed 
+# Flashes the on-board LED once quickly each second when it's running but not connected to another device
+# Flashes the on-board LED twice quickly each second when connected to another device 
+add_executable(doorbell_client
```
will-v-pi: Would it be possible to prefix these standalone examples with `ble_` or `bt_` to make it clear what they are? VS Code just searches by the target name, it currently doesn't categorise the examples based on where they came from, so prefixing them with `ble_` (like the `pio_` examples already have) will make it easier to search for them.

E.g for this one, with similar changes to the temp_sensors 
```suggestion
add_executable(ble_doorbell_client
```

Not sure if the same change would be useful for the btstack examples, their names might already be clear enough what they do

peterharperuk: yes, that should be ok

will-v-pi (approved):
LGTM from a VS Code perspective - along with https://github.com/raspberrypi/pico-vscode/pull/261 this will make the bluetooth examples work with VS Code

peterharperuk:
> LGTM from a VS Code perspective - along with [raspberrypi/pico-vscode#261](https://github.com/raspberrypi/pico-vscode/pull/261) this will make the bluetooth examples work with VS Code

nice

kilograham (approved):

will-v-pi:
The final commit to this PR to add `ENABLE_LE_ENHANCED_CONNECTION_COMPLETE_EVENT` appears to have broken the `ble_temp_sensor` example for me? The devices print that they are connected, but the `ble_temp_reader` never prints `Read temp ...` despite the `ble_temp_server` printing `Write temp ...`. Commenting out `ENABLE_LE_ENHANCED_CONNECTION_COMPLETE_EVENT` in the config fixes the issue for me, and they both print again.

@peterharperuk can you reproduce what I'm seeing? We should maybe revert that change, if there's a risk it may have broken other examples too.

peterharperuk:
@will-v-pi Yes, I can reproduce that. Let me see if I can work out what's going on before we take it out.

peterharperuk:
I've raised this https://github.com/raspberrypi/pico-examples/issues/772
I'd rather not remove ENABLE_LE_ENHANCED_CONNECTION_COMPLETE_EVENT as there was a good reason it was added.

PR #745, opened 2026-03-18
Deinit everything when restarting bt standalone temp sensor

will-v-pi:
Without this, I get an error when attempting to reconnect after a restart by pressing the S key:
```
Connecting to device with addr DE:AB:01:06:C1:6F.
SERVICE_QUERY_RESULT, ATT Error 0x7f.
Disconnected DE:AB:01:06:C1:6F
Connecting to device with addr DE:AB:01:06:C1:6F.
SERVICE_QUERY_RESULT, ATT Error 0x7f.
Disconnected DE:AB:01:06:C1:6F
...
```

With this fix everything works correctly when starting & stopping using the S key

I'm not sure if everything needs to be deinit'ed, but it is probably safest to deinit all three of these as they are init'ed above

### DIFF
--- a/pico_w/bt/standalone/server/server.c
+++ b/pico_w/bt/standalone/server/server.c
@@ -190,6 +190,11 @@ restart:
         async_context_wait_for_work_until(cyw43_arch_async_context(), at_the_end_of_time);
     }
 
+    att_server_deinit();
+
+    sm_deinit();
+    l2cap_deinit();
+
     cyw43_arch_deinit();
 
     printf("Press the \"S\" key to Start bluetooth\n");


peterharperuk (approved):

will-v-pi:
@peterharperuk Do you want to absorb this change into your larger PR https://github.com/raspberrypi/pico-examples/pull/743 - or should I merge this separately?

peterharperuk:
It might be quicker if you merge this first. I can make sure this change is picked up in there

will-v-pi:
Sounds good - I'll merge this now then (once the CI is done)

mringwal:
Hi there. The _deinit functions have been added for reproducible testing. They should not be required/used in regular applications. (we're not going to make any changes to them, just suggesting that this is probably not the correct fix for the underlying issue). 

The idea is that main sets up the application which then initializes all needed service. In case of a power down, it's sufficient to call hci_power_control(off) and later turn it on in the same way. 

What's / where can I find the original problem?

will-v-pi:
This was reproducible with the [picow_ble_temp_sensor](https://github.com/raspberrypi/pico-examples/tree/develop/pico_w/bt/standalone/server) example on Pico 2 W - I can reproduce it using the latest `develop` branches of pico-sdk and pico-examples, if I remove the `deinit` lines added by this PR.

You need to load `picow_ble_temp_sensor` onto one device and `picow_ble_temp_reader`, then send `S` over UART to `picow_ble_temp_sensor` to turn the bluetooth off and back on - after turning it off-on-off you get this error on the `picow_ble_temp_reader`:
```
Connecting to device with addr 2C:CF:67:CB:B2:B9.
SERVICE_QUERY_RESULT, ATT Error 0x1f.
Disconnected 2C:CF:67:CB:B2:B9
```

I've just tried replacing these `deinit` lines with `hci_power_control(HCI_POWER_OFF)` but that is still showing the same error on the reader.

mringwal:
I'm sorry I haven't looked at the standalone examples yet. The while (1) { poll asysnc } looks scary to me.
A regular Bluetooth app should call _init functions once and not call _deinit, as it's not needed.

Ideally, main sets up the environment and then calls run_loop_execute. The keypress could be tested from a BTstack timer,. On keypress we could call hci_power_control (off), wait for the off state, then wait for the next key, and turnon again. 

Should I rework the standalone client & server examples? I guess I should be able to do that next wee.

--- For the curious ---
More about _init & _deiniti. The next to the visible _init functions there's are other _init function that initialize the run loop and setup the memory pools. Using _deinit correctly requires all _deinit functions to be called (as each layer just drop's it's own stuff, but does not free anything, as btstack_memory_free will do that).

will-v-pi:
I'll ping @peterharperuk over rewriting as he knows more, but I would want the re-written example to still exit the run loop before restarting, rather than staying inside the run loop and just calling `hci_power_control(on/off)`.

The reason for this is the current example can be modified to stop after X seconds (instead of waiting for a keypress), and then call `low_power_pstate_for_ms` to turn off the RP2350 for a time (instead of waiting for another keypress), for much lower power consumption. If it never exits the run loop, then I don't think this form would work, as it would be turning off the RP2350 while the run loop is still running, so wouldn't be able to call `cyw43_arch_deinit` to turn off the CYW43.

PR #746, opened 2026-03-23
Make universal blink example work with W boards

will-v-pi:
This detects between W and non-W boards using the ADC, as described in "Connecting to the Internet with Raspberry Pi Pico W-series"

This method should work for all boards powered from VSYS, which should cover all standard use cases for the blink example

This does increase the UF2 file size from 96KB to 1,023KB, as the combined UF2 contains 2 copies of the wifi firmware, but in flash it is still only half that size

### DIFF
--- a/universal/CMakeLists.txt
+++ b/universal/CMakeLists.txt
@@ -146,18 +146,31 @@ function (add_universal_target TARGET SOURCE)
     # into a single universal UF2. This is required as there isn't a single family
     # ID which is accepted by both RP2040 and RP2350. Instead, the 2 UF2 files are
     # concatenated together and the device ignores the part not targeting it.
-    add_custom_target(${TARGET}_rp2350_uf2
-            COMMAND picotool uf2 convert ${COMBINED} ${BINDIR}/rp2350.uf2 --family absolute --offset ${PACKADDR}
+    add_custom_target(${TARGET}_absolute_uf2
+            COMMAND picotool uf2 convert ${COMBINED} ${BINDIR}/absolute.uf2 --family absolute --offset ${PACKADDR}
             DEPENDS ${TARGET}_combined
             )
     add_custom_target(${TARGET}_rp2040_uf2
             COMMAND picotool uf2 convert ${RP2040_COMBINED} ${BINDIR}/rp2040.uf2 --family rp2040 --offset ${PACKADDR}
             DEPENDS ${TARGET}_combined
             )
-    add_custom_target(${TARGET}_uf2
-            COMMAND ${CMAKE_COMMAND} -E cat ${BINDIR}/rp2040.uf2 ${BINDIR}/rp2350.uf2 > ${BINDIR}/${TARGET}.uf2
-            DEPENDS ${TARGET}_rp2350_uf2 ${TARGET}_rp2040_uf2
-            )
+
+    if (CMAKE_HOST_UNIX)
+        # Can use tail and truncate to make a better UF2 file
+        add_custom_target(${TARGET}_uf2
+                COMMAND tail -c 512 ${BINDIR}/rp2040.uf2 > ${BINDIR}/rp2040.end.uf2
+                COMMAND cat ${BINDIR}/rp2040.uf2 > ${BINDIR}/rp2040.start.uf2
+                COMMAND truncate --size=-512 ${BINDIR}/rp2040.start.uf2
+                COMMAND cat ${BINDIR}/rp2040.start.uf2 ${BINDIR}/absolute.uf2 ${BINDIR}/rp2040.end.uf2 > ${BINDIR}/${TARGET}.uf2
+                DEPENDS ${TARGET}_absolute_uf2 ${TARGET}_rp2040_uf2
+                )
+    else()
+        # Not guaranteed to have tail or truncate, so 
+        add_custom_target(${TARGET}_uf2
+                COMMAND ${CMAKE_COMMAND} -E cat ${BINDIR}/rp2040.uf2 ${BINDIR}/absolute.uf2 > ${BINDIR}/${TARGET}.uf2
+                DEPENDS ${TARGET}_absolute_uf2 ${TARGET}_rp2040_uf2
+                )
+    endif()
 
     add_dependencies(${TARGET} ${TARGET}_combined ${TARGET}_uf2)
 endfunction()
--- a/universal/CMakeLists.txt
+++ b/universal/CMakeLists.txt
@@ -21,9 +21,10 @@ include(ExternalProject)
 # The build will output a TARGET.bin file which can be written using picotool, and a
 # TARGET.uf2 file which can be dragged and dropped onto the device in BOOTSEL mode
 function (add_universal_target TARGET SOURCE)
+    set(zeroValueArgs SEPARATE_RP2040)
     set(oneValueArgs SOURCE_TARGET PADDING PACKADDR BOARD_RP2040 BOARD_RP2350)
     set(multiValueArgs PLATFORMS)
-    cmake_parse_arguments(PARSE_ARGV 2 PARSED "" "${oneValueArgs}" "${multiValueArgs}")
+    cmake_parse_arguments(PARSE_ARGV 2 PARSED "${zeroValueArgs}" "${oneValueArgs}" "${multiValueArgs}")
 
     set(SOURCE_TARGET ${TARGET})
     if (PARSED_SOURCE_TARGET)
@@ -107,13 +108,35 @@ function (add_universal_target TARGET SOURCE)
         message(FATAL_ERROR "Cannot link universal binary without picotool")
     endif()
 
-    # Link the binaries for different platforms into a single block loop, with
-    # appropriate rolling window deltas. This creates a universal binary file,
-    # which will run on any of the platforms when loaded using picotool.
-    add_custom_target(${TARGET}_combined
-            COMMAND picotool link ${COMBINED} ${BINS} --pad ${PADDING}
-            DEPENDS ${DEPS}
-            )
+    list(FIND PLATFORMS "rp2040" idx)
+    if (idx EQUAL 0 AND PARSED_SEPARATE_RP2040)
+        # Don't include RP2040 bin in the combined BIN, just include it in the RP2040 UF2
+        # POP_FRONT only added in CMake 3.15, so use GET and REMOVE_AT instead
+        list(GET BINS 0 RP2040_BIN)
+        list(REMOVE_AT BINS 0)
+        set(RP2040_COMBINED ${RP2040_BIN})
+    else()
+        set(RP2040_COMBINED ${COMBINED})
+    endif()
+
+    list(LENGTH BINS BINS_COUNT)
+    if (BINS_COUNT GREATER 1)
+        # Link the binaries for different platforms into a single block loop, with
+        # appropriate rolling window deltas. This creates a universal binary file,
+        # which will run on any of the platforms when loaded using picotool.
+        add_custom_target(${TARGET}_combined
+                COMMAND picotool link ${COMBINED} ${BINS} --pad ${PADDING}
+                DEPENDS ${DEPS}
+                )
+    else()
+        # Only one binary left, so no picotool link is needed - just copy instead
+        # This could be the case if only building for rp2040 and rp2350-arm-s,
+        # with SEPARATE_RP2040 set
+        add_custom_target(${TARGET}_combined
+                COMMAND ${CMAKE_COMMAND} -E copy ${BINS} ${COMBINED}
+                DEPENDS ${DEPS}
+                )
+    endif()
 
     # Create UF2s targeting the absolute and rp2040 family IDs, then combine these
     # into a single universal UF2. This is required as there isn't a single family
@@ -124,7 +147,7 @@ function (add_universal_target TARGET SOURCE)
             DEPENDS ${TARGET}_combined
             )
     add_custom_target(${TARGET}_rp2040_uf2
-            COMMAND picotool uf2 convert ${COMBINED} ${BINDIR}/rp2040.uf2 --family rp2040 --offset ${PACKADDR}
+            COMMAND picotool uf2 convert ${RP2040_COMBINED} ${BINDIR}/rp2040.uf2 --family rp2040 --offset ${PACKADDR}
             DEPENDS ${TARGET}_combined
             )
     add_custom_target(${TARGET}_uf2
@@ -146,7 +169,7 @@ add_universal_target(blink_universal
     SOURCE_TARGET blink_universal
     BOARD_RP2040 universal
     BOARD_RP2350 universal
-    PLATFORMS "rp2040;rp2350-arm-s" # Skip RISC-V, as wifi firmware takes up lots of space
+    SEPARATE_RP2040 PLATFORMS "rp2040;rp2350-arm-s" # Skip RISC-V and keep RP2040 separate, as wifi firmware takes up lots of space
     )
 
 # nuke binary - is no_flash, so needs to be sent to SRAM on RP2040
--- a/universal/CMakeLists.txt
+++ b/universal/CMakeLists.txt
@@ -156,7 +156,10 @@ function (add_universal_target TARGET SOURCE)
             )
 
     if (CMAKE_HOST_UNIX)
-        # Can use tail and truncate to make a better UF2 file
+        # If you just concatenate the files, then on Windows you get an alarming 'write error' dialog when copying
+        # large UF2s, because the device reboots too soon. Moving the final RP2040 UF2 block to the end of the file
+        # fixes this, as it won't reboot until the final block is received. This fix requires `tail` and `truncate`,
+        # so only use it when building on Unix-like systems.
         add_custom_target(${TARGET}_uf2
                 COMMAND tail -c 512 ${BINDIR}/rp2040.uf2 > ${BINDIR}/rp2040.end.uf2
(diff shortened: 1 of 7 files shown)


universal/blink_universal/blink_universal.c
```diff
@@ -0,0 +1,66 @@
...
+void pico_set_led(bool led_on, bool is_w) {
+    if (is_w) {
+        cyw43_arch_gpio_put(CYW43_WL_GPIO_LED_PIN, led_on);
+    } else {
+        gpio_put(PICO_DEFAULT_LED_PIN, led_on);
+    }
+}
+
+int main() {
+    bool is_w = detect_is_w_using_adc();
```
lurch: Should this be modified to check that it's being built for a board where all of `PICO_VSYS_PIN`, `PICO_DEFAULT_LED_PIN` and `CYW43_WL_GPIO_LED_PIN` have been defined? (which probably won't be many, as I don't think many board-headers define both `PICO_DEFAULT_LED_PIN` and `CYW43_WL_GPIO_LED_PIN` :wink: )

universal/blink_universal/blink_universal.c
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Turn the led on or off
+void pico_set_led(bool led_on, bool is_w) {
+    if (is_w) {
+        cyw43_arch_gpio_put(CYW43_WL_GPIO_LED_PIN, led_on);
+    } else {
+        gpio_put(PICO_DEFAULT_LED_PIN, led_on);
+    }
+}
```
lurch, line 73: Does it make sense to use @peterharperuk 's status_led library for this?

universal/blink_universal/blink_universal.c
will-v-pi: The CMakeLists.txt already checks to ensure you're using the custom universal board when building it

universal/blink_universal/blink_universal.c
will-v-pi: No, because that relies on compile-time knowledge of what board you'll be running on - the whole point of this example is so that it doesn't need to know what it's running on at compile-time, and can figure it out at runtime

universal/blink_universal/blink_universal.c
```diff
@@ -0,0 +1,66 @@
...
+
+#include "pico/stdlib.h"
+#include "pico/cyw43_arch.h"
+#include "hardware/adc.h"
+
+#ifndef LED_DELAY_MS
+#define LED_DELAY_MS 250
+#endif
+
+bool detect_is_w_using_adc(void) {
```
lurch: Minor nitpick, but should this be named `detect_is_pico_w_using_adc` to clarify that it only detects Raspberry Pi Pico W and Raspberry Pi Pico 2 W, as I guess there's no guarantee that it would detect "is_w" on any third-party boards that are using RM2?

universal/CMakeLists.txt
```diff
@@ -135,8 +165,11 @@ add_universal_target(hello_universal
 
 # blink binary
 add_universal_target(blink_universal
-    ${CMAKE_CURRENT_LIST_DIR}/../blink
-    SOURCE_TARGET blink
+    ${CMAKE_CURRENT_LIST_DIR}/blink_universal
+    SOURCE_TARGET blink_universal
+    BOARD_RP2040 universal
+    BOARD_RP2350 universal
+    SEPARATE_RP2040 PLATFORMS "rp2040;rp2350-arm-s" # Skip RISC-V and keep RP2040 separate, as wifi firmware takes up lots of space
```
lurch: Perhaps this should be split over two lines, as I initially misread this as `SEPARATE_RP2040_PLATFORMS` (and got confused why it was then referencing `rp2350-arm-s`) :man_facepalming:

universal/blink_universal/blink_universal.c
lurch: Ahh right, thanks for the clarification. I think you probably told me that in-person before too :laughing:

universal/CMakeLists.txt
will-v-pi: Can do - I put them on the same line as the comment applies to both, but this is probably cleaner

```suggestion
    # Skip RISC-V and keep RP2040 separate, as wifi firmware takes up lots of space
    SEPARATE_RP2040
    PLATFORMS "rp2040;rp2350-arm-s" 
```

lurch:
I guess the `universal.h` is only designed to run on the official Raspberry Pi microcontroller boards (Pico, Pico W, Pico WH, Pico 2, Pico 2 W) rather than on any 3rd-party RP2040 or RP2350 boards, so perhaps it ought to be named something like `pico_universal.h` ?

will-v-pi:
> I guess the `universal.h` is only designed to run on the official Raspberry Pi microcontroller boards (Pico, Pico W, Pico WH, Pico 2, Pico 2 W) rather than on any 3rd-party RP2040 or RP2350 boards, so perhaps it ought to be named something like `pico_universal.h` ?

It does run on quite a few 3rd-party boards - as long as they use the same LED and VSYS pins it will work (which applies to most of the boards on my desk)

lurch:
> It does run on quite a few 3rd-party boards - as long as they use the same LED and VSYS pins it will work (which applies to most of the boards on my desk)

Oh, neat!

universal/CMakeLists.txt
lurch: Maybe it makes sense to swap them around to 
```suggestion
    # Skip RISC-V and keep RP2040 separate, as wifi firmware takes up lots of space
    PLATFORMS "rp2040;rp2350-arm-s"
    SEPARATE_RP2040
```
so that the order of the parameters matches the order in the comment?

universal/blink_universal/blink_universal.c
peterharperuk: I think a comment about what this function is doing would be useful. And add a reference to the doc with a link?

peterharperuk:
Could we add an example partition json for this? And maybe a readme for how to run the universal examples on a blank Pico 2 board?

peterh@rpi5-peterh-dev2:~ $ picotool load blink_universal.uf2
WARNING: Multiple family IDs in a single UF2 file - only using first one
Family ID 'rp2040' cannot be downloaded anywhere
ERROR: This file cannot be loaded onto a device with no partition table

will-v-pi:
> Could we add an example partition json for this? And maybe a readme for how to run the universal examples on a blank Pico 2 board?
> 
> peterh@rpi5-peterh-dev2:~ $ picotool load blink_universal.uf2
> WARNING: Multiple family IDs in a single UF2 file - only using first one
> Family ID 'rp2040' cannot be downloaded anywhere
> ERROR: This file cannot be loaded onto a device with no partition table

This doesn't use partitions - the error you're getting is because when picotool sees two UF2s concatenated (which is what a universal UF2 is), it only uses the first file, which is the RP2040 one. That could be fixed, but this is an existing issue for all universal UF2s.

These universal UF2s are designed for drag & drop instead

peterharperuk:
> These universal UF2s are designed for drag & drop instead

Ok, understood. It does work if you drag and drop. Maybe we should say this in the readme.

peterharperuk (dismissed):

universal/README.md
```diff
@@ -0,0 +1,55 @@
...
+  - Same as the Arm one
+
+All of the embedded blocks are linked into one big block loop.
+
+These are then booted by the respective bootroms:
+- **RP2040** - sees the boot2 at the start and uses that to execute the RP2040 binary, as
+RP2040 has no support for embedded blocks.
+- **RP2350** - sees the block loop and parses it to find the correct embedded block to boot
+from (Arm vs RISC-V). It then translates the flash address according to the
+`ROLLING_WINDOW_DELTA` so that the binary containing that embedded block is at the start of the
```
lurch:
```suggestion
`ROLLING_WINDOW_DELTA` so that the binary containing that embedded block appears at the start of the
```
?

universal/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+For no_flash binaries the RP2040 boot2 is omitted as the bootrom just executes from the start
+of SRAM, and instead of `ROLLING_WINDOW_DELTA` items the RP2350 binaries use `LOAD_MAP` items,
+to copy the code in SRAM to the correct location for execution rather than using address
+translation.
+
+## How you should use them
+
+For most use cases, **Universal UF2s** are the best option to use. They will only load into
+flash the code that runs on that device. The [blink_universal](blink_universal) example uses a
```
lurch:
```suggestion
For most use cases, **Universal UF2s** are the best option to use. They will only load the
code that runs on that device into flash. The [blink_universal](blink_universal) example uses a
```
IMHO easier to read, this is :wink:

universal/README.md
```diff
@@ -0,0 +1,55 @@
...
+Universal UF2 for that reason, as the Wi-Fi firmware is quite large. **Universal Binaries**
```
lurch, line 53: Just out of curiosity, how big would blink_universal be if it was built as a universal binary?

universal/blink_universal/blink_universal.c
```diff
@@ -12,7 +12,21 @@
 #define LED_DELAY_MS 250
 #endif
 
-bool detect_is_w_using_adc(void) {
+enum BOARD_TYPE {
+    BOARD_TYPE_PICO,
+    BOARD_TYPE_PICO_W,
+    BOARD_TYPE_UNKNOWN,
+};
```
lurch, line 19: Maybe it's worth adding a small comment that BOARD_TYPE_PICO matches both Pico and Pico 2, and BOARD_TYPE_PICO_W matches both Pico W and Pico 2 W ?

universal/blink_universal/blink_universal.c
```diff
@@ -25,15 +39,20 @@ bool detect_is_w_using_adc(void) {
...
-        return true;
+        // Pico W-series board
+        return BOARD_TYPE_PICO_W;
+    } else if (value == 0) {
+        // Pico-series board
+        return BOARD_TYPE_PICO;
     } else {
-        return false;
+        // Unknown board
+        return BOARD_TYPE_UNKNOWN;
```
lurch, line 49: The code elsewhere in this file always does just:
```c
if (board_type == BOARD_TYPE_PICO_W) {
    // do this
} else {
    // do that
}
```

Should the code be doing anything different for `BOARD_TYPE_PICO` and `BOARD_TYPE_UNKNOWN` ?

universal/README.md
```diff
@@ -0,0 +1,55 @@
...
+## Universal Binary vs Universal UF2
+
+There is a difference between a **Universal Binary** and a **Universal UF2**,
+for the purposes of these examples:
+- A **Universal Binary** is a `.bin` file that can be loaded into flash (or sram) and executed,
+allowing RP2040 and RP2350 (Arm & RISC-V) to run from identical flash contents.
+- A **Universal UF2** is multiple individual `.uf2` files with different family IDs
+concatenated together to create a single `.uf2` file. When dragged & dropped onto a device, 
+only the file with a family ID corresponding to that device will be loaded onto it, and the
+rest of the files will be ignored.
```
lurch:
```suggestion
only the portion of the file with a family ID corresponding to that device will be processed, and the
rest of the file will be ignored.
```
?

universal/README.md
will-v-pi: 1.5MB if it was Arm, RISC-V and RP2040 - so a 3MB UF2 file

universal/README.md
```diff
@@ -0,0 +1,55 @@
...
+- RP2040 boot2
+  - Required by the RP2040 bootrom
+- RP2040 binary containing an embedded block
+  - The embedded block contains an `IGNORED` item due to RP2350-E13, but you can use an RP2040
+  `IMAGE_DEF` item instead if not using RP2350-A2 chips
+- RP2350 Arm binary containing an embedded block
+  - In addition to the RP2350 `IMAGE_DEF` item, this embedded block contains a
+  `ROLLING_WINDOW_DELTA` item to translate this binary to the start of flash for execution
+- RP2350 RISC-V binary containing an embedded block
+  - Same as the Arm one
```
lurch: I guess the "Same as the Arm one" _might_ be misinterpreted to mean that the RISC-V binary is the same as the Arm binary, which could be confusing?

universal/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+A **Universal Binary** can be packaged into a UF2 file for loading onto a device. However,
+as there isn't a common family ID between RP2040 and RP2350, you would have to package it into a **Universal UF2** with two copies (using `rp2040` and `absolute` family IDs), thus creating a **Universal UF2** of a **Universal Binary**.
+
+## How Universal Binaries work
+
+Universal binaries must be recognised by both the RP2040 and RP2350 bootroms. Therefore, they need the following structure for flash binaries:
+- RP2040 boot2
+  - Required by the RP2040 bootrom
+- RP2040 binary containing an embedded block
```
lurch, line 25: So is the "embedded block" in the RP2040 binary **only** for use by the RP2350, and ignored by the RP2040? (if so, maybe that's worth clarifying?)

universal/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+These are then booted by the respective bootroms:
+- **RP2040** - sees the boot2 at the start and uses that to execute the RP2040 binary, as
+RP2040 has no support for embedded blocks.
+- **RP2350** - sees the block loop and parses it to find the correct embedded block to boot
+from (Arm vs RISC-V). It then translates the flash address according to the
+`ROLLING_WINDOW_DELTA` so that the binary containing that embedded block is at the start of the
+flash address space, and executes from there.
+
+For no_flash binaries the RP2040 boot2 is omitted as the bootrom just executes from the start
```
lurch:
```suggestion
For no_flash binaries the RP2040 boot2 is omitted as the RP2040 bootrom just executes from the start
```
?

universal/blink_universal/blink_universal.c
will-v-pi:
```suggestion
enum BOARD_TYPE {
    BOARD_TYPE_PICO,    // Pico-series board
    BOARD_TYPE_PICO_W,  // Pico W-series board
    BOARD_TYPE_UNKNOWN,
};
```

universal/blink_universal/blink_universal.c
will-v-pi: I think treating unknown boards as Picos is the best option, as that will still work for any board with an LED on GPIO25

universal/README.md
will-v-pi: It's clarified in the boot process description below
```markdown
- **RP2040** - sees the boot2 at the start and uses that to execute the RP2040 binary, as
RP2040 has no support for embedded blocks.
```

universal/README.md
will-v-pi: Maybe just this then?
```suggestion
  - Ditto
```

universal/blink_universal/blink_universal.c
lurch: Fair enough. I guess it just felt a bit weird that the code was detecting BOARD_TYPE_PICO vs. BOARD_TYPE_UNKNOWN but then not doing anything with that information. :shrug:

peterharperuk (dismissed):
It works. I seem to be getting an interesting copy error from windows (I don't remember getting this before?) with pico1 devices , presumably because the device has rebooted before the whole uf2 has been written? It still works despite this although it's a bit disconcerting.

will-v-pi:
> It works. I seem to be getting an interesting copy error from windows (I don't remember getting this before?) with pico1 devices , presumably because the device has rebooted before the whole uf2 has been written? It still works despite this although it's a bit disconcerting.

Yes, I get that too - I think it doesn't occur for the smaller universal binaries because the device doesn't disappear until after Windows thinks it's loaded the whole file, but this one is larger so you get this error when the device disappears

will-v-pi:
@peterharperuk I think I've fixed it with that latest commit - it now puts the last RP2040 UF2 block at the end of the file, so the RP2040 doesn't reboot until it has the whole file

The fix is gated to only run when compiling on Unix-like OSes, as it requires `tail` and `truncate` which may not be available on others. At some point this functionality could be added into `picotool`, but that can happen later.

universal/CMakeLists.txt
```diff
@@ -100,30 +112,65 @@ function (add_universal_target TARGET SOURCE)
...
+            COMMAND picotool uf2 convert ${RP2040_COMBINED} ${BINDIR}/rp2040.uf2 --family rp2040 --offset ${PACKADDR}
             DEPENDS ${TARGET}_combined
             )
-    add_custom_target(${TARGET}_uf2
-            COMMAND ${CMAKE_COMMAND} -E cat ${BINDIR}/rp2040.uf2 ${BINDIR}/rp2350.uf2 > ${BINDIR}/${TARGET}.uf2
-            DEPENDS ${TARGET}_rp2350_uf2 ${TARGET}_rp2040_uf2
-            )
+
+    if (CMAKE_HOST_UNIX)
+        # Can use tail and truncate to make a better UF2 file
```
peterharperuk: Sorry to be picky but could you be a bit more descriptive about why we're doing this for any future people to understand the reason? e.g. "On Windows you get an alarming 'write error' dialog because the device reboots too soon. Moving the RP2040 UF2 block to the end fixes this."

Are RP2040 UF2 blocks always 512 bytes? Not 256 bytes?

will-v-pi: UF2 blocks are always 512 bytes (https://github.com/microsoft/uf2#overview) - on RP2040 and RP2350 they only contain 256 bytes of data, but the block in the file is always 512 bytes

I'll improve the comments

will-v-pi: I've pushed the improved comments

peterharperuk (approved):

PR #749, opened 2026-04-21
Examples using low power API

peterharperuk:


low_power/low_power_pstate/low_power_pstate_timer.c
```diff
@@ -0,0 +1,55 @@
...
+
+// The example will repeatedly wait 10 seconds then switch off for 10 seconds
+// The debugger will appear to be unresponsive while the device is off
+int main() {
+    stdio_init_all();
+    // Must start the aon timer if needed
+    printf("Current time: %llu\n", aon_timer_get_absolute_time());
+    if (!aon_timer_is_running()) {
+        low_power_start_aon_timer_at_time_ms(1776858754000);
+    }
```
will-v-pi: It might be better to use persistent data here, so that it always sets the time when run? Otherwise when you load this repeatedly using picotool, the time will keep going up and never be reset to the start time

I think this would work
```suggestion
    static bool __persistent_data(aon_timer_started);
    if (!aon_timer_started) {
        low_power_start_aon_timer_at_time_ms(1776858754000);
        aon_timer_started = true;
    }
```

peterharperuk: I added this. But if I enable stdusb to test loading via picotool I get an assert when the low power api calls stdio_usb_deinit. It's not clear to me why - I'll raise an issue.

kilograham:
i guess this is still not complete

peterharperuk:
Why? I don't think I have anything more to do.

kilograham:
well there is no README.md update, and the title is "Start"

will-v-pi (commented):
LGTM - just needs adding to the readme, and most of the copyrights need updating from 2024 to 2026

will-v-pi (approved):

Issue #605, opened 2025-02-01
Bug - Memory leak in Pico W http example


cirrith:
I'm still relatively new to Pico and was trying to connect my Pico 2 W to my Wi-Fi so that it could act as a simple remote sensor.

I believe the provided example `pico_w/wifi/http_client/example_http_client_util.c` has a memory leak, which prevents it from making more than a few requests before exhausting the allocated protocol control blocks (PCB).

This issue wouldn't have appeared in the provided code as it only makes a few requests before exiting. However, if the example is adapted to a real project, the bug would prevent boards from making more than ~PBUF_POOL_SIZE number of requests.

A `pbuf_free(p);` should be added to the receiver function to free the buffer after use. Here is the fixed version:

```c
err_t http_client_receive_print_fn(__unused void *arg, __unused struct altcp_pcb *conn, struct pbuf *p, err_t err) {
    HTTP_INFO("\ncontent err %d\n", err);
    u16_t offset = 0;
    while (offset < p->tot_len) {
        char c = (char)pbuf_get_at(p, offset++);
        HTTP_INFOC(c);
    }
    pbuf_free(p);
    return ERR_OK;
}
```

Optionally, it can be added to the `internal_recv_fn` function to make it easier to swap in a custom receiver.

peterharperuk:
Yes, your right. I have a branch somewhere with this fixed. But feel free to raise a PR if you want.

peterharperuk:
This should be fixed now.

PR #752, opened 2026-05-26
http_client free pbuf in callbacks

anatol-newton:
Free pbufs after they were used in multiple callbacks:
`http_client_header_print_fn` and `http_client_receive_print_fn` are example callbacks, that print the received data. The buffers should be freed afterwards, as they are not further used,

`internal_header_fn` is the internal function, that later calls a user callback. Only free the pbuf if no user callback is called. When a user callback is called, the freeing has to take place in there.

If the buffers are not freed, the MCU will eventually run out of RAM (in my application where I found the issue this occurred already after about twelve requests)

### DIFF
--- a/pico_w/wifi/http_client/example_http_client_util.c
+++ b/pico_w/wifi/http_client/example_http_client_util.c
@@ -43,6 +43,7 @@ err_t http_client_header_print_fn(__unused httpc_state_t *connection, __unused v
         char c = (char)pbuf_get_at(hdr, offset++);
         HTTP_INFOC(c);
     }
+    pbuf_free(hdr);
     return ERR_OK;
 }
 
@@ -54,6 +55,7 @@ err_t http_client_receive_print_fn(__unused void *arg, __unused struct altcp_pcb
         char c = (char)pbuf_get_at(p, offset++);
         HTTP_INFOC(c);
     }
+    pbuf_free(p);
     return ERR_OK;
 }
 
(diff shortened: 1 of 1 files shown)


peterharperuk:
Do you have an example that demonstrates the problem? I'm not convinced. pbuf_free is called in internal_recv_fn.
I changed the code to call http_client_request_sync 100 times and it worked ok.

JeanMax:
Hello!

I agree we miss a call to `pbuf_free` somewhere.
A fix was proposed here in #605, but nothing as been merged as far as I can tell.

You can check the call to  the recv callback in lwip here: 
https://github.com/lwip-tcpip/lwip/blob/master/src/apps/http/http_client.c#L350

Notice how a call to `pbuf_free` is done if no recv callback is supplied.

But I don't think we should call `pbuf_free` in each of the `example_http_client_util.c` callbacks, as it will be applied to the same `pbuf`: both headers_done and recv are handled sequentially in `httpc_tcp_recv` (the lwip fun I just linked above).

Now I'm wondering if we miss a `altcp_recved(conn, p->tot_len)` too? :|




Edit: I just did a checkout develop, and now I see why @peterharperuk says the `pbuf_free`  is already in `internal_recv_fn` ... there is also the `altcp_recved` hahaha! 
Sorry about that, cheers

peterharperuk:
A leak was fixed in this commit on the develop branch https://github.com/raspberrypi/pico-examples/commit/d3c4bace50cb155d9638320a6e42787e9bac167d

PR #753, opened 2026-06-03
feat(devcontainer): Add DevContainer

JakubK:
This PR adds basic [DevContainer](https://containers.dev) which allows to build `.uf2` files of samples and any other basic project based on Pico SDK.

I had some issues with configuring and installing all dependencies on my distro, working in devcontainer helped to get started.
I thought it might be helpful for others.

If this PR gets accepted, next step would be to update "Quick-start your own project" section in official SDK README to mention this alternative way.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,12 @@
+FROM ubuntu:22.04
+
+RUN apt-get update && apt-get install -y \
+    git \
+    cmake \
+    ninja-build \
+    build-essential \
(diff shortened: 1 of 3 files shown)


lurch:
If you're having problems installing pico-sdk and pico-examples and https://pip.raspberrypi.com/documents/RP-008276-DS doesn't help, I think that we'd prefer to make improvements to our documentation, rather than adding support for additional build-systems that we have no familiarity with.

Please report any problems with the Pico-related PDFs to https://github.com/raspberrypi/pico-feedback/

Issue #382, opened 2023-05-17
Fix needed to DHCP server


peterharperuk:
If you have two netifs there's a chance a DHCP request will be received on one and the response sent on the other. Fix below...

This is the patch against the access_point example

server/dhcpserver.c
116c116
< static int dhcp_socket_sendto(struct udp_pcb **udp, struct netif *nif, const void *buf, size_t len, uint32_t ip, uint16_t port) {
---
> static int dhcp_socket_sendto(struct udp_pcb **udp, const void *buf, size_t len, uint32_t ip, uint16_t port) {
130,138c130
<
<     err_t err;
<     if(nif != NULL) {
<         err = udp_sendto_if(*udp, p, &dest, port, nif);
<     }
<     else {
<         err = udp_sendto(*udp, p, &dest, port);
<     }
<
---
>     err_t err = udp_sendto(*udp, p, &dest, port);
191d182
<     struct netif *nif = ip_current_input_netif();
294c285
<     dhcp_socket_sendto(&d->udp, nif, &dhcp_msg, opt - (uint8_t *)&dhcp_msg, 0xffffffff, PORT_DHCP_CLIENT);
---
>     dhcp_socket_sendto(&d->udp, &dhcp_msg, opt - (uint8_t *)&dhcp_msg, 0xffffffff, PORT_DHCP_CLIENT);

PR #754, opened 2026-06-05
Add pico_w/wifi/access_point_wifi_provisioning/

peterharperuk:


kilograham (approved):

will-v-pi (commented):
This doesn't seem to handle SSIDs with spaces correctly, because the form submission passes them as `+` instead, so it tries to connect to "My+Network" instead of "My Network". This might be an issue for other special characters too.

kilograham (changes requested):
un-approving to allow for will's comment

peterharperuk:
This example relies on this change https://github.com/raspberrypi/pico-sdk/pull/3000

will-v-pi:
> This example relies on this change [raspberrypi/pico-sdk#3000](https://github.com/raspberrypi/pico-sdk/pull/3000)

I'm still seeing the issue with SSIDs with spaces now that change has been merged - do SSIDs with spaces work for you?

peterharperuk:
I haven't looked into that problem yet

peterharperuk:
Oops...
SSID AND PASSWORD: My+SSID+Name blah

peterharperuk:
I fixed the issue with spaces in ssid names. But while testing it I realised the wifi connection logic wasn't reliable. We can't connect in a callback. I think this is fixed by moving the connecting to main.

I've made other improvements.

If we add this, it means we have another dhcpserver and dnsserver in the pico-examples tree which isn't ideal. I'm wondering whether I should just replace the existing access_point example with this one. Or leave the simpler one as it's easier to understand?

I think I should get rid of the duplicate dhcp / dns servers though as I've fixed another bug like https://github.com/raspberrypi/pico-examples/issues/382 where the dhcp server was handing out addresses on the network.

will-v-pi (approved):
Working now for me

I think we should keep the simpler example too, but de-duplication of the dhcp/dns servers would be useful at a later date - preferably not before this next release, as it will probably require some fixups to VS Code to handle copying in the dhcp/dns libraries into the generated examples

kilograham (approved):

PR #760, opened 2026-06-17
Rationalize bluetooth example paths

kilograham:
there is much confusion here, but i think this is supposed to be ble_pointer, right?


lurch:
Looking through the commit history, it looks like this was previously renamed in commit 859c3bcd0db3359b61b9c4ac54a33c5c8c06b05b as part of the #743 PR ?

peterharperuk:
Yes, I renamed it at the last minute to just "pointer" as they're all BLE examples so it seemed a bit odd for it to be called "ble_pointer". Was there an issue?

kilograham:
I was having merge issues with another branch, and couldn't see which was correct.

I think (since all the targets are "ble_foo") we should probably rename all the directories to be ble?

will-v-pi (approved):
I've pushed a README update as there were some examples missing from the list, and one in the list which didn't exist under `bluetooth/btstack_examples` - LGTM now

PR #761, opened 2026-06-18
Turn on LED at start of flash nuke

will-v-pi:
This is useful for boards that take a long time to erase flash (eg Pimoroni Pico Plus 2), to offer reassurance that something is happening

### DIFF
--- a/flash/nuke/nuke.c
+++ b/flash/nuke/nuke.c
@@ -28,6 +28,13 @@
 #define MIN_FLASH_SIZE (1024 * 1024 * 2)
 
 int main() {
+#ifdef PICO_DEFAULT_LED_PIN
+    // Turn on LED to show erase starting
+    gpio_init(PICO_DEFAULT_LED_PIN);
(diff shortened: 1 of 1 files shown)


kilograham (commented):
perhaps we should use pico_status_led now (but without wifi support probably) - would give us WS2812

will-v-pi:
> perhaps we should use pico_status_led now (but without wifi support probably) - would give us WS2812

Can do, but that will only help people compiling their own flash_nuke - the universal one we distribute will still only toggle GPIO 25

kilograham:
oh good point; lets not worry about it here - perhaps we should however update pico_status_led to support pico_config too (at least optionally)

will-v-pi:
> oh good point; lets not worry about it here - perhaps we should however update pico_status_led to support pico_config too (at least optionally)

Maybe even add an option for runtime detection, like the new `blink_universal` example - although that doesn't include WS2812

kilograham (approved):

PR #766, opened 2026-06-23
Use new pio functions

peterharperuk:
We twiddle input_sync_bypass directly in the pio spi example. Use the new pio_set_input_sync_bypass_with_mask64 instead.

Run the loopback example at full speed and improve it so it's more obvious when the example doesn't work.

### DIFF
--- a/pio/spi/spi_loopback.c
+++ b/pio/spi/spi_loopback.c
@@ -86,5 +86,5 @@ int main() {
         }
     }
     printf("Example %s\n", match_count == spi_example_runs ? "PASSED" : "FAILED");
-    assert(match_count == 4);
+    assert(match_count == spi_example_runs);
 }
--- a/pio/spi/spi_loopback.c
+++ b/pio/spi/spi_loopback.c
@@ -21,6 +21,8 @@
 
 #define BUF_SIZE 20
 
+static uint32_t match_count;
+
 void test(const pio_spi_inst_t *spi) {
     static uint8_t txbuf[BUF_SIZE];
     static uint8_t rxbuf[BUF_SIZE];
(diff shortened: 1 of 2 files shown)


peterharperuk:
Depends on https://github.com/raspberrypi/pico-sdk/pull/3024

pio/spi/spi_loopback.c
```diff
@@ -75,4 +83,6 @@ int main() {
             sleep_ms(10);
         }
     }
+    printf("Example %s\n", match_count == 4 ? "passed" : "failed");
```
lurch: Might be better to do e.g.
```c
    int spi_tests_count = 0;
    for (int cpha = 0; cpha <= 1; ++cpha) {
        for (int cpol = 0; cpol <= 1; ++cpol) {
            spi_tests_count++;
            printf("CPHA = %d, CPOL = %d\n", cpha, cpol);
            // etc. etc.
        }
    }
    printf("Example %s\n", match_count == spi_tests_count ? "passed" : "failed");
    assert(match_count == spi_tests_count);
```

Also, I wonder if it might be worth using the same PASSED / FAILED strings as e.g. https://github.com/raspberrypi/pico-sdk/blob/develop/test/pico_float_test/custom_float_funcs_test.c#L828 ?

peterharperuk: I get in trouble when I treat examples like test code.

lurch: Ahhh, I'd missed that this was the pico-examples repo and not the pico-sdk repo :facepalm:

peterharperuk: Done

pio/spi/spi_loopback.c
```diff
@@ -75,4 +85,6 @@ int main() {
             sleep_ms(10);
         }
     }
+    printf("Example %s\n", match_count == spi_example_runs ? "PASSED" : "FAILED");
+    assert(match_count == 4);
```
lurch:
```suggestion
    assert(match_count == spi_example_runs);
```

(Unless there's a good reason for not doing that?)

pio/spi/spi_loopback.c
```diff
@@ -40,26 +42,34 @@ void test(const pio_spi_inst_t *spi) {
...
     float clkdiv = 31.25f;  // 1 MHz @ 125 clk_sys
+#endif
 
     // pio_claim_free_sm_and_add_program_for_gpio_range finds a free pio and state machine for a program and sets the gpio base correctly
     const uint pin_base = MIN(PIN_SCK, MIN(PIN_MOSI, PIN_MISO));
     const uint pin_count = MAX(PIN_SCK, MAX(PIN_MOSI, PIN_MISO)) - pin_base + 1;
     hard_assert(pio_claim_free_sm_and_add_program_for_gpio_range(&spi_cpha0_program, &spi[0].pio, &spi[0].sm, &spi[0].offset, pin_base, pin_count, true));
     hard_assert(pio_claim_free_sm_and_add_program_for_gpio_range(&spi_cpha1_program, &spi[1].pio, &spi[1].sm, &spi[1].offset, pin_base, pin_count, true));
 
+    int spi_example_runs = 0;
```
lurch: Only a very minor nitpick, but perhaps it makes sense for `match_count` and `spi_example_runs` to be of the same type?

peterharperuk: yes - and maybe it should be called spi_example_passes?

lurch (approved):

PR #769, opened 2026-06-26
Access point fix to DNS and DHCP server

peterharperuk:
Make sure the DHCP and DNS server only use the AP netif. Should stop them answer queries from other networks.
Also log the host in DHCP


liamfraser (approved):

andygpz11 (approved):
Looks good to me.

PR #770, opened 2026-06-29
Add ina237 and ina219 examples

will-v-pi:
Add ina219 and ina237 examples, following on from the ina260 example

ina237 is newer and generally better all-round (higher precision, and higher voltage & current ranges), but ina219 has been around for longer so is probably more common, hence adding examples for both of them

The ina237 example also adds UART passthrough for the Pico being measured


peterharperuk (approved):
the uart pass through sounds like a good idea.

PR #767, opened 2026-06-25
Improve ble secure temp sensor

peterharperuk:
This tidies up the example a bit and adds some scripts to make the different combinations easier to test

PR #773, opened 2026-06-29
Bt con handle fixes

peterharperuk:
A change was made to add ENABLE_LE_ENHANCED_CONNECTION_COMPLETE_EVENT to the btstack config as it makes pairing more reliable. But this has shown up some faulty example code.

We no longer get the HCI_EVENT_LE_META / HCI_SUBEVENT_LE_CONNECTION_COMPLETE event.
Instead we should get the handle from HCI_EVENT_META_GAP which handles the difference for us.
This is what the rest of the btstack examples do.


peterharperuk:
The secure example has some redundant code for this event but actually behaves properly - I've pushed a change to remove the bad code in here https://github.com/raspberrypi/pico-examples/pull/767

will-v-pi (dismissed):
Thanks - works for me

will-v-pi (approved):
Sorry, accidentally pushed a different commit to this branch - have removed it

PR #774, opened 2026-06-30
Fix host build

peterharperuk:


low_power/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-add_subdirectory(low_power_sleep)
-add_subdirectory(low_power_dormant)
+add_subdirectory_exclude_platforms(low_power_sleep host)
+add_subdirectory_exclude_platforms(low_power_dormant host)
 if (NOT PICO_RP2040)
-    add_subdirectory(low_power_pstate)
+    add_subdirectory_exclude_platforms(low_power_pstate host)
```
kilograham: seems like we should just add rp2040 to excluded platforms ;-)

peterharperuk: it' s wrapped in NOT PICO_RP2040 but that's probably a better solution.

kilograham: i know; fixing to use the better solution was my point

peterharperuk: Done

kilograham (approved):

PR #777, opened 2026-07-06
Improve BT example run loop use

will-v-pi:
This standardises the standalone BT examples to use `btstack_run_loop_execute` by default, with a brief explanation that you can also run user code instead, but must not perform any BT work in that code. These examples all use `threadsafe_background`, so I removed the references to polling. If we do want to support polling, that can be added with `PICO_CYW43_ARCH_POLL`, similar to the wifi examples.

`ble_temp_server` is modified to support staying in the run loop when starting/stopping, to avoid the need to call the `deinit` functions as those should not be used (see https://github.com/raspberrypi/pico-examples/pull/745#issuecomment-4679259208)

A new `ble_temp_server_low_power` example is also added, demonstrating a custom lower power run loop, and running for 15s on then powering down for 60s (pstate if available, otherwise dormant followed by watchdog reboot).

### DIFF
--- a/bluetooth/ble_doorbell/client.c
+++ b/bluetooth/ble_doorbell/client.c
@@ -299,8 +299,8 @@ int main() {
 #if 1
     btstack_run_loop_execute();
 #else
-    // This example uses the 'threadsafe background` method, where BT work is handled in a low priority IRQ, so it
-    // is fine to call bt_stack_run_loop_execute() but equally you can continue executing user code. You must not
+    // This example uses the `threadsafe background` method, where BT work is handled in a low priority IRQ, so it
+    // is possible to continue executing user code instead of calling bt_stack_run_loop_execute(). You must not
     // perform any BT work in user code (e.g calling `hci_power_control(HCI_POWER_OFF)`) - those operations must be
     // performed inside the run loop (e.g. using `btstack_run_loop_execute_on_main_thread`).
     while(true) {
--- a/bluetooth/ble_doorbell/client.c
+++ b/bluetooth/ble_doorbell/client.c
@@ -296,18 +296,14 @@ int main() {
     // turn on!
     hci_power_control(HCI_POWER_ON);
 
-    // btstack_run_loop_execute is only required when using the 'polling' method (e.g. using pico_cyw43_arch_poll library).
-    // This example uses the 'threadsafe background` method, where BT work is handled in a low priority IRQ, so it
-    // is fine to call bt_stack_run_loop_execute() but equally you can continue executing user code.
-
-#if 1 // this is only necessary when using polling (which we aren't, but we're showing it is still safe to call in this case)
+#if 1
     btstack_run_loop_execute();
 #else
-    // this core is free to do it's own stuff except when using 'polling' method (in which case you should use 
-    // btstacK_run_loop_ methods to add work to the run loop.
-
-    // this is a forever loop in place of where user code would go.
-    while(true) {      
+    // This example uses the 'threadsafe background` method, where BT work is handled in a low priority IRQ, so it
+    // is fine to call bt_stack_run_loop_execute() but equally you can continue executing user code. You must not
+    // perform any BT work in user code (e.g calling `hci_power_control(HCI_POWER_OFF)`) - those operations must be
+    // performed inside the run loop (e.g. using `btstack_run_loop_execute_on_main_thread`).
+    while(true) {
         sleep_ms(1000);
     }
 #endif
--- a/bluetooth/ble_temp_sensor/server.c
+++ b/bluetooth/ble_temp_sensor/server.c
@@ -242,7 +242,7 @@ int main() {
 #if SERVER_LOW_POWER
     // Custom low power run loop, using low power sleep
     while (!exit_run_loop) {
-        // This will us the same timer as the bluetooth processing, so that will continue to work
+        // This will use the same timer as the bluetooth processing, so that will continue to work
         low_power_sleep_for_ms(1000, NULL, false);
     }
 #else
@@ -255,8 +255,8 @@ int main() {
     printf("Going to low power state for %ds\n", POWER_DOWN_TIME_S);
     low_power_set_pins_low_leakage_exclude_mask(0);
 #if HAS_POWMAN_TIMER
-    low_power_pstate_for_ms(POWER_DOWN_TIME_MS, NULL, NULL);
-    while(true); // wait for reboot
+    int rc = low_power_pstate_for_ms(POWER_DOWN_TIME_MS, NULL, NULL);
+    printf("low_power_pstate_for_ms returned error %d\n", rc);
 #else
     low_power_dormant_for_ms(POWER_DOWN_TIME_MS, DORMANT_CLOCK_SOURCE_DEFAULT, NULL);
     watchdog_reboot(0, 0, 10);
--- a/bluetooth/ble_temp_sensor/server.c
+++ b/bluetooth/ble_temp_sensor/server.c
@@ -119,17 +119,24 @@ static void heartbeat_handler(struct btstack_timer_source *ts) {
     static uint32_t counter = 0;
     counter++;
 
-    // Update the temp every 10s
-    if (counter % 10 == 0) {
-        poll_temp();
-        if (le_notification_enabled) {
-            att_server_request_can_send_now_event(con_handle);
+    static int led_on = true;
+
+    if (hci_get_state() == HCI_STATE_OFF) {
+        // Blink the led every 10s
+        led_on = (counter % 10 == 0);
+    } else {
+        // Update the temp every 10s
+        if (counter % 10 == 0) {
+            poll_temp();
+            if (le_notification_enabled) {
+                att_server_request_can_send_now_event(con_handle);
+            }
         }
+
+        // Invert the led every 1s
+        led_on = !led_on;
     }
 
-    // Invert the led
-    static int led_on = true;
-    led_on = !led_on;
     cyw43_arch_gpio_put(CYW43_WL_GPIO_LED_PIN, led_on);
 
     // Restart timer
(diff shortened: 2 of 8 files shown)


will-v-pi:
cc @mringwal - I think this PR provides a reasonable balance between recommending users use `btstack_run_loop_execute` by default, while providing some additional explanation if they wish to run user code instead

bluetooth/ble_doorbell/client.c
```diff
@@ -296,18 +296,14 @@ int main() {
...
-#if 1 // this is only necessary when using polling (which we aren't, but we're showing it is still safe to call in this case)
+#if 1
     btstack_run_loop_execute();
 #else
-    // this core is free to do it's own stuff except when using 'polling' method (in which case you should use 
-    // btstacK_run_loop_ methods to add work to the run loop.
-
-    // this is a forever loop in place of where user code would go.
-    while(true) {      
+    // This example uses the 'threadsafe background` method, where BT work is handled in a low priority IRQ, so it
```
lurch: Does this `// This example uses the 'threadsafe background' method...` text apply to the whole example, or only to this (unused-by-default) `#else` section? Either way, perhaps it could be clarified?

(Also, you've got mismatched quotes around `threadsafe background` - which was also true in the original version!)

bluetooth/ble_temp_sensor/server.c
```diff
@@ -184,24 +239,32 @@ int main() {
...
     hci_power_control(HCI_POWER_ON);
 
-    key_pressed = false;
-    while(!key_pressed) {
-        async_context_poll(cyw43_arch_async_context());
-        async_context_wait_for_work_until(cyw43_arch_async_context(), at_the_end_of_time);
+#if SERVER_LOW_POWER
+    // Custom low power run loop, using low power sleep
+    while (!exit_run_loop) {
+        // This will us the same timer as the bluetooth processing, so that will continue to work
```
lurch:
```suggestion
        // This will use the same timer as the bluetooth processing, so that will continue to work
```

bluetooth/ble_temp_sensor/server.c
```diff
@@ -184,24 +239,32 @@ int main() {
...
-    while(!key_pressed) {
-        sleep_ms(1000);
-    }
-    goto restart;
+#if SERVER_LOW_POWER
+    printf("Going to low power state for %ds\n", POWER_DOWN_TIME_S);
+    low_power_set_pins_low_leakage_exclude_mask(0);
+#if HAS_POWMAN_TIMER
+    low_power_pstate_for_ms(POWER_DOWN_TIME_MS, NULL, NULL);
+    while(true); // wait for reboot
```
lurch: Should the comment here be `// wait for pstate` ? :shrug:

bluetooth/ble_temp_sensor/server.c
```diff
@@ -113,51 +126,93 @@ static void poll_temp(void) {
...
+        led_on = !led_on;
     }
 
-    // Invert the led
-    static int led_on = true;
-    led_on = !led_on;
     cyw43_arch_gpio_put(CYW43_WL_GPIO_LED_PIN, led_on);
 
+#if SERVER_LOW_POWER
+    if (counter >= POWER_UP_TIME_S) {
```
lurch: Does there need to be something that resets `counter` back to 0?

will-v-pi: Yes, it gets reset by the reboot

bluetooth/ble_temp_sensor/server.c
will-v-pi: Actually, the above function shouldn’t return, this is just to catch it if it does

bluetooth/ble_doorbell/client.c
will-v-pi: It’s the whole example - the comment is only in the else because it’s advanced information. The point of this PR is to say that you **should** use `btstack_run_loop_execute` by default, but you can do other stuff instead if you need to.

bluetooth/ble_temp_sensor/server.c
lurch: Oh, I didn't realise that a pstate also did a reboot. The docs for [`low_power_pstate_for_ms`](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_pico_low_power_1ga4fb0b9e83359752c9588584548a989da) refer to the docs for [`low_power_pstate_until_aon_timer`](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_pico_low_power_1ga1aa14b25f7f3d36ccc9fdf2c9a94376c) which says "If pstate is NULL, it will go to the minimum Pstate that will keep persistent data powered on." - so this `counter` isn't part of the "persistent data" ?

...ah, I've just seen that https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#detailed-description-8 says "Waking up from Pstate will run your program from the start" (which I guess is equivalent to a reboot?) and "Variables can be marked as persistent using the `__persistent_data` macro" - which answers my questions :slightly_smiling_face:

bluetooth/ble_doorbell/client.c
lurch: Perhaps I'm misunderstanding things, but I don't think it's clear which part of the comment is referring to the `#if` part and which is referring to the `#else` part. E.g. is the "but equally you can continue executing user code" part of the comment referring to the `#else` part? :confused:

bluetooth/ble_temp_sensor/server.c
```diff
@@ -17,6 +17,19 @@
 #define ADC_CHANNEL_TEMPSENSOR 4
 #define APP_AD_FLAGS 0x06
 
+#if SERVER_LOW_POWER
+#include "pico/low_power.h"
+#if !HAS_POWMAN_TIMER
+// Uses the watchdog to reboot, as there is no powman
+#include "hardware/watchdog.h"
+#endif
+#define POWER_UP_TIME_S 15
```
lurch: Seems like this is only in "seconds" because `HEARTBEAT_PERIOD_MS` is set to 1000. I guess if `HEARTBEAT_PERIOD_MS` was set to e.g. 2000 then a `POWER_UP_TIME_S` of 15 would actually be awake for 30 seconds before rebooting?
Perhaps it's worth doing
```c
#define POWER_UP_HEARTBEATS (uint)((POWER_UP_TIME_S * 1000) / HEARTBEAT_PERIOD_MS)
```
and then changing the later check to:
```c
if (counter >= POWER_UP_HEARTBEATS) {
```
?

bluetooth/ble_doorbell/client.c
will-v-pi: Would this make it clearer that the comment is only referring to the `#else` part?
```
    // This example uses the `threadsafe background` method, where BT work is handled in a low priority IRQ, so it
    // is possible to continue executing user code instead of calling bt_stack_run_loop_execute(). You must not
    // perform any BT work in user code (e.g calling `hci_power_control(HCI_POWER_OFF)`) - those operations must be
    // performed inside the run loop (e.g. using `btstack_run_loop_execute_on_main_thread`).
```

bluetooth/ble_temp_sensor/server.c
will-v-pi: I'll just change the variable to `POWER_UP_HEARTBEATS` - if the user adjusts `HEARTBEAT_PERIOD_MS` they can adjust `POWER_UP_HEARTBEATS` to suit

bluetooth/ble_temp_sensor/server.c
will-v-pi: Yeah, rather than duplicating the details of sleep vs dormant vs pstate for every function, I put the descriptions at the top level.

bluetooth/ble_temp_sensor/server.c
will-v-pi: Updated to print the error if it returns instead

peterharperuk (dismissed):

bluetooth/ble_doorbell/client.c
lurch: If this is now specific to the `#else` part, perhaps "This example..." would make more sense as "This version of the example..."? :thinking:

bluetooth/ble_temp_sensor/server.c
lurch: "Power Up Heartbeats" - like in the film "Crank"? :joy:

bluetooth/ble_doorbell/client.c
will-v-pi: It is true for the whole example (changing `#if 0` to `#if 1` doesn't change anything else), but only needs explanation in the context of not using `bt_stack_run_loop_execute()`.

The idea is that this is an advanced explanation of how the underlying btstack run loop works, hence putting it in the `#else` to make it clear that the default is to use `bt_stack_run_loop_execute()`.

lurch: Sorry to be pedantic, but I still think that including a comment within an `#else` block, and starting it with "This example..." makes it confusing exactly which part you're describing :confused: 
I'm still not sure what this is trying to explain (I've never looked too closely at the BT code), but would "If not using `bt_stack_run_loop_execute()` (i.e. by changing the `#if 1` to `#if 0`), then..." be a viable alternative?

will-v-pi: They way I've written it, I basically want the entire `#else` section to be considered a comment (in VS Code it renders as such), but an advanced one that you don't need to read unless you don't want to use `bt_stack_run_loop_execute()`.

image

The "This example uses the `threadsafe background` method" is true in all circumstances, so putting that behind "If not using `bt_stack_run_loop_execute()` ..." would not be correct, as it still uses the `threadsafe background` method when you are using `bt_stack_run_loop_execute()`. The rest of the comment also still applies in all circumstances, you just won't be executing user code outside the run loop if you use `bt_stack_run_loop_execute()`.

The point of this change is to make this comment much less prominent, so that most users won't have to consider it, but still keep the information available for more advanced users who do want to run a custom run loop (eg, as demonstrated in the `ble_temp_server_low_power` example).

lurch: Okay

mringwal:
Thanks @will-v-pi for the updates. The _deinit functions are gone and the default uses run loop execute as requested.
The "don't call BTstack APIs" from a different context is important and there.

--

The current code allows for 'sS' to toggle on/off. On key press, bt_start_stop_fn is executed on the main thread, which checks the current HCI state. Next to on & off, there's also "halting" and "starting" up. If the stack is in an intermediate state, this might not give the expected result. Ideally, we should read stdin only in states on and off. As hci_get_state is a const function, it could be called kind of safely from a different thread, but I'd rather avoid that. I guess the best we can do here is to check the current state in bt_start_stop_fn. If it's OFF or WORKING, use the current code. if not, use a btstack timer to retry e.g. 100 ms later (implementing a very slow spin lock).

An alternative to that would be to have key_pressed_func managed a shared global that indicates the 'target' state of the stack. Upon key press, the function on the main thread can call hci_power_control as needed. The stack is robust to correctly handle request during intermediate states (at least that's the plan). This might result in simpler code and avoids the spin lock.

will-v-pi:
Thanks @mringwal for that point - I've modified `bt_start_stop_fn` to keep track of the 'target' state of the stack in a local static variable, as it's just a toggling function so I don't think it needs a global variable. I think that solves the issue, as `hci_get_state() == HCI_STATE_OFF` is now only used in the heartbeat handler to decide how to toggle the LED.

Let me know if there's still any issues

mringwal:
> Thanks @mringwal for that point - I've modified `bt_start_stop_fn` to keep track of the 'target' state of the stack in a local static variable, as it's just a toggling function so I don't think it needs a global variable. I think that solves the issue, as `hci_get_state() == HCI_STATE_OFF` is now only used in the heartbeat handler to decide how to toggle the LED.
> 
> Let me know if there's still any issues

As you toggle in bt_start_stop, the local variable is fine / that's even better than my suggestion. Looks good to me.

PR #778, opened 2026-07-23
tinyusb examples fail to build with latest tinyusb

peterharperuk:
The tinyusb examples in the latest version of tinyusb 0.21.0 try and run tinyusb/tools/linkermap/linkermap.py which only exists if you have run tools/get_deps.py

Override family_add_linkermap to do nothing if this file does not exist - which allows the examples to build cleanly.

### DIFF
--- a/usb/CMakeLists.txt
+++ b/usb/CMakeLists.txt
@@ -11,6 +11,14 @@ set(TINYUSB_OPT_OS OPT_OS_PICO)
 # Same issue as above - this variable goes out of scope
 set(LINKERMAP_PY ${PICO_TINYUSB_PATH}/tools/linkermap/linkermap.py)
 
+if (NOT EXISTS ${LINKERMAP_PY})
+    # family_add_linkermap tries to run tinyusb/tools/linkermap/linkermap.py
+    # But this file only exists if you run python tools/get_deps.py
+    # So override the function to do nothing if the file does not exist
+    function(family_add_linkermap TARGET)
+    endfunction()
+endif()
+
 # List of freertos examples
 set(tinyusb_freertos_examples
     audio_4_channel_mic_freertos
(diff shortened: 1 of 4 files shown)


lurch:
I'm assuming that this is due to https://github.com/hathach/tinyusb/blob/master/hw/bsp/family_support.cmake#L521-L524 ?
The [`family_add_bloaty` function](https://github.com/hathach/tinyusb/blob/master/hw/bsp/family_support.cmake#L230) already has an early-return if `bloaty` isn't installed, and this PR now overrides `family_add_linkermap`, so should this PR also override `family_add_membrowse` ? :shrug: 

@hathach Maybe it'd be nice if tinyusb had some simple way of "turning off" these additional checks / analysis?

peterharperuk:
> I'm assuming that this is due to https://github.com/hathach/tinyusb/blob/master/hw/bsp/family_support.cmake#L521-L524 ? The [`family_add_bloaty` function](https://github.com/hathach/tinyusb/blob/master/hw/bsp/family_support.cmake#L230) already has an early-return if `bloaty` isn't installed, and this PR now overrides `family_add_linkermap`, so should this PR also override `family_add_membrowse` ? 🤷
> 
> @hathach Maybe it'd be nice if tinyusb had some simple way of "turning off" these additional checks / analysis?

It's actually this line that causes the problem. https://github.com/hathach/tinyusb/blob/master/hw/bsp/rp2040/family.cmake#L245
I guess it was always a mistake to define this as part of the interface

lurch:
> It's actually this line that causes the problem.

I was actually talking about the first commit in this PR, rather than the second :slightly_smiling_face:

hathach:
will look into this later

kilograham:
subsumed by #787 hopefully.

Once again though we should add a CI to TinyUSB to check for breakages here (we can help likely)... there are some other breakages I want to fix it TinyUSB itself too, which i will PR when i have more time

PR #786, opened 2026-08-30
Enable USB and UART for ws2812 targets

brenerpittner:
Enable USB and UART for ws2812 targets, now we can read what the example sends via the serial port


lurch:
Sounds like a duplicate of #321 ?
Also, you appear to be turning **off** UART, which doesn't match what your PR description says.

brenerpittner:
Yes, it's the same. Thanks.

PR #789, opened 2026-09-19
Add rpi-connect/rpi-connect-ota-demo example

kilograham:
This requires the rpi-connect-preview branch of pico-sdk


rpi-connect/rpi-connect-ota-demo/README.md
```diff
@@ -0,0 +1,464 @@
...
+    --device-pubkey device-pub-key.pem \
+    --description   "Pico 2 W OTA demo" \
+    --device-name   "pico-ota-01"
+```
+
+The device now exists in your organisation. (You can sanity-check the whole
+auth chain before touching OTP by running
+`rpi-connect-test --device-identity-exchange --serial <serial>
+--device-privkey device-priv-key.pem --device-pubkey device-pub-key.pem`, which
+prints the `RPI_CONNECT_TOKEN` the device would obtain.)
```
will-v-pi: I don't think this is going to work in VS Code, because we don't have host compilers (especially on Windows) - would it be possible to do this setup using a Python script instead?

rpi-connect/rpi-connect-ota-demo/tools/pico_rpi_connect_program_identity_key_otp.sh
will-v-pi: This key should be programmed as ECC, not RAW

Also, this functionality should probably be added into `picotool otp load`, instead of requiring this script - maybe `picotool otp load -s <ROW> <FILE>.pem` (it can detect the file type). `picotool` already has the functionality to convert PEM files to 32-byte scalars in `read_keys`, so that also removes the dependency on OpenSSL here.

rpi-connect/rpi-connect-ota-demo/README.md
will-v-pi: Or alternatively a separate binary that runs on the device to perform first registration with the `RPI_CONNECT_ORG_TOKEN` compiled in, maybe reading the private key from OTP directly and calculating the public key?

rpi-connect/rpi-connect-ota-demo/tools/partition_pico2_for_ffs.sh
```diff
@@ -0,0 +1,250 @@
...
+check_firmware_sealed
+build_ffs_blob
+
+picotool reboot -uf
+sleep $sleep_time
+
+# Workaround lack of E10 errata handling in picotool:
+# Need to request an `info` before requesting the erase!
+picotool info
+picotool erase --range 0x10000000 0x10400000
```
will-v-pi: This workaround is in picotool now

```suggestion

picotool erase --range 0x10000000 0x10400000
```

rpi-connect/rpi-connect-ota-demo/main_pico.c
```diff
@@ -0,0 +1,135 @@
...
+
+// Returns a malloc'd copy of a NUL-terminated string stored in FFS, or NULL.
+static char *ffs_get_string(uint8_t file_id) {
+    const char *data;
+    int rc = ffs_read(file_id, &data);
+    if (rc >= 0 && data) {
+        return strdup(data);
+    }
+    return NULL;
+}
```
will-v-pi: Given `ffs_read` returns the file size, could this be modified to read non-NUL-terminated strings instead? Those would make it possible to create these files using CMake, as CMake cannot write NUL bytes to a file.

Maybe just use `strndup`, which should provide support for both NUL-terminated and non-NUL-terminated strings?
```suggestion
static char *ffs_get_string(uint8_t file_id) {
    const char *data;
    int rc = ffs_read(file_id, &data);
    if (rc >= 0 && data) {
        return strndup(data, rc);
    }
    return NULL;
}
```

rpi-connect/rpi-connect-ota-demo/tools/pico_rpi_connect_program_identity_key_otp.sh
will-v-pi: Implemented in https://github.com/raspberrypi/picotool/pull/362

rpi-connect/rpi-connect-ota-demo/README.md
```diff
@@ -0,0 +1,464 @@
...
+At this point the device has signed in via the OTP key, cached its token in FFS,
+registered the OTA capability, checked for a deployment assigned while it was
+offline, and is waiting for deployments.
+
+### Step 7 — Deploy an update
+
+From the Raspberry Pi Connect dashboard (or API) for your organisation, create a
+deployment targeting this device with a new `rpi_connect_ota_demo.uf2`. The
+image must carry a strictly higher picobin version than the one running or the
+bootrom reverts to the old image at the next power cycle: the minor version
```
will-v-pi: Is this statement true? Presumably the update reboots with a `flash_update_boot`, in which case it will work fine if you update with an older version, it'll just erase the newer version when you buy it

rpi-connect/rpi-connect-ota-demo/README.md
```diff
@@ -0,0 +1,464 @@
...
+and registers its OTA capability, and only then is success reported and the
+bootrom *buy* issued to commit it (both from `rpi_connect_ota_boot_sync()`). So the deployment is
+reported successful only once the new firmware has actually booted and reached
+the API. If the new image can't sign in or reach the server (or crashes
+earlier), neither the success report nor the buy happens, and the bootrom rolls
+back to the previous image on the next reset. The old image then finds the
+leftover `APPLYING` status on a normal (non-flash-update) boot and reports the
+deployment **failed**, so the server knows to re-deploy. If a download is
+interrupted by a reboot mid-way, the resume path picks it up and re-downloads
+automatically.
```
will-v-pi: This describes the TBYB behaviour of the bootrom, but I don't see `PICO_CRT0_IMAGE_TYPE_TBYB` being set anywhere - if that isn't set, then the bootrom buys the image before booting it, and therefore will not roll back to the old image on failure, instead it'll just boot the new image again.

Is `PICO_CRT0_IMAGE_TYPE_TBYB` actually being set on the update UF2?

will-v-pi: Can confirm that it does not behave well without `PICO_CRT0_IMAGE_TYPE_TBYB` set - if I update from major version 1 to 2, and yank power during the first boot of version 2, when I plug back in it boots version 2 but reports that the deployment failed:
```
Update was not applied for deployment ID=xxx; reporting failure
Failing deployment ID=xxx reason=update not applied
```

rpi-connect/rpi-connect-ota-demo/README.md
will-v-pi: Can confirm that it will update fine to an older version, using the normal version downgrade functionality (bootrom erases the newer version, so it always picks the older version)

will-v-pi:
Running this overnight seems to eventually lead to this error - could something be added to recover from this?

```
Received PING event: Timestamp=1790239118
Received PING event: Timestamp=1790239178
Received PING event: Timestamp=1790239238
Request completed with result 5 len 60027 server_response 0 err 0 in 58026041ms
HTTP client result: result=5, err=0, srv_res=0
HTTP client request failed: result=5, err=0
internal_result_fn - complete
Async request completed with code -1 result 5
```

This was with the following debug options:
```
RPI_CONNECT_DEBUG_ENABLE=1
RPI_CONNECT_OTA_DEBUG_ENABLE=1
HTTPC_DEBUG=LWIP_DBG_ON
RPI_CONNECT_OTA_DEMO_DEBUG_ENABLE=1
```

